PR #1, opened 2021-04-26
Add github workflow

klemmari1:
- Update README.
- Add editorconfig.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,67 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'python', 'javascript' ]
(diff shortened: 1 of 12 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'python' ]
```
nenonja, line 31: javascript could be added?

.github/workflows/staging.yml
```diff
@@ -0,0 +1,89 @@
+name: Build & Staging
+on:
+  push:
+#    branches:
```
nenonja, line 4: why commented?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '33 22 * * 1'
```
nenonja, line 21: seems a bit odd scheduled time? https://crontab.guru/#33_22_*_*_1

nenonja (commented):
Quickly browsed through the code and added some comments for the parts I understood :)

.github/workflows/codeql-analysis.yml
klemmari1: Yep, I'll add :+1:

.github/workflows/staging.yml
klemmari1: This was commented to test out that deployment to staging works before we push to the `main` branch. I'll uncomment it now that it is finally working :D

.github/workflows/codeql-analysis.yml
klemmari1: This is an example setup I'm using from https://github.com/City-of-Helsinki/kuva-demo-project. I think it is just to run it once a week, I don't know if the time matters.

nenonja: This could be more convenient: https://crontab.guru/every-week

nenonja (approved):
Looks good now for me

PR #2, opened 2021-04-27
YJDH-37 | Add backend automated CI tests and code style checks

klemmari1:
- Fix static files access.
- Add pre-commit config.
- Reformat files according to the style checks.


backend/docker-entrypoint.sh
```diff
@@ -14,10 +14,13 @@ if [[ "$APPLY_MIGRATIONS" = "1" ]]; then
     ./manage.py migrate --noinput
 fi
 
-# Create superuser
+# Create admin user. Generate password if there isn't one in the environment variables. Password is not printed to log, but needs to be changed at the pod command line
```
suutari-ai, line 17: Wrap the comment for readability
```
# Create admin user. Generate password if there isn't one in the
# environment variables. Password is not printed to log, but needs to be
# changed at the pod command line
```

backend/docker-entrypoint.sh
```diff
@@ -14,10 +14,13 @@ if [[ "$APPLY_MIGRATIONS" = "1" ]]; then
...
 if [[ "$CREATE_SUPERUSER" = "1" ]]; then
-    ./manage.py add_admin_user -u admin -p admin -e admin@example.com
-    echo "Admin user created with credentials admin:admin (email: admin@example.com)"
+    if [[ "$ADMIN_USER_PASSWORD" ]]; then
+      DJANGO_SUPERUSER_PASSWORD=$ADMIN_USER_PASSWORD DJANGO_SUPERUSER_USERNAME=admin DJANGO_SUPERUSER_EMAIL=admin@hel.ninja python /app/manage.py createsuperuser --noinput || true
+    else
+      DJANGO_SUPERUSER_PASSWORD=admin DJANGO_SUPERUSER_USERNAME=admin DJANGO_SUPERUSER_EMAIL=admin@hel.ninja python /app/manage.py createsuperuser --noinput || true
+    fi
```
suutari-ai, line 23: Indentation seems to be 4 spaces here, respect that.

Wrap the long code lines, e.g. :
```sh
    DJANGO_SUPERUSER_PASSWORD=$ADMIN_USER_PASSWORD \
        DJANGO_SUPERUSER_USERNAME=admin \
        DJANGO_SUPERUSER_EMAIL=admin@hel.ninja \
        python /app/manage.py createsuperuser --noinput || true
```

backend/kesaseteli/settings.py
```diff
@@ -26,8 +27,7 @@
     USE_X_FORWARDED_HOST=(bool, False),
     DATABASE_URL=(
         str,
-        "postgres://kesaseteli:kesaseteli"
-        "@localhost/kesaseteli",
+        "postgres://kesaseteli:kesaseteli" "@localhost/kesaseteli",
```
suutari-ai, line 29: There's no point in string splitting if its on a single line, so combine it to single string instead.
Also, maybe the default value could be something simpler such as `postgres:///kesaseteli` which works nicely with a locally running PostgreSQL.

setup.cfg
```diff
@@ -15,7 +15,7 @@ doctest_optionflags = NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL ALLOW_UNICODE
```
suutari-ai, line 15: Moving the setup.cfg from backend to root doesn't seem correct, since setup.cfg is a Python specific configuration file and backend seems to be the only directory containing a Python project here. Why do you move it?

.github/workflows/py-coding-style.yml
```diff
@@ -0,0 +1,48 @@
...
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: pip install -r backend/requirements.txt -r backend/requirements-dev.txt
+
+      - name: Formatting
+        run: black --check backend/
+
+      - name: Linting
+        run: flake8 backend/
```
suutari-ai, line 45: Should probably be something like `cd backend && flake8`, since that's the root of the Python project here.

.github/workflows/py-coding-style.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Import sorting
+        run: isort backend/ -c
```
suutari-ai, line 48: same here `cd backed && isort -c`  (or does it need a period for current dir i.e. `isort -c .`?)

backend/kesaseteli/settings.py
```diff
@@ -1,6 +1,5 @@
-import os
-
 import environ
+import os
```
suutari-ai: Your isort config is incorrect, since `os` is a standard library, but `environ` is a third party library so they should be in different sections.

backend/kesaseteli/wsgi.py
```diff
@@ -8,7 +8,6 @@
 """
 
 import os
-
 from django.core.wsgi import get_wsgi_application
```
suutari-ai, line 11: Same here, incorrect isort config, since the blank between is correct.

suutari-ai (commented):
Great work in setting up all of that important CI stuff!

Commented on some minor details which you could look at.

setup.cfg
klemmari1: I think I moved it because the CI style checks were not finding the config file in the backend folder. But with for example `cd backend && flake8` it should work :+1:

klemmari1: Moved back.

backend/docker-entrypoint.sh
```diff
@@ -14,10 +14,19 @@ if [[ "$APPLY_MIGRATIONS" = "1" ]]; then
     ./manage.py migrate --noinput
 fi
 
-# Create superuser
-if [[ "$CREATE_SUPERUSER" = "1" ]]; then
-    ./manage.py add_admin_user -u admin -p admin -e admin@example.com
-    echo "Admin user created with credentials admin:admin (email: admin@example.com)"
+# Create admin user. Generate password if there isn't one in the
+# environment variables. Password is not printed to log, but needs to be
+# changed at the pod command lineif [[ "$CREATE_SUPERUSER" = "1" ]]; then
```
suutari-ai, line 19: There's something extra in this comment.

suutari-ai (commented):
Looks better now. Just fix the extra code which was somehow copy-pasted to the shell script comment and then it's good to go.

backend/docker-entrypoint.sh
klemmari1: Whoops :D

suutari-ai (approved):
Now everything I found seems to be addressed. Good job! :+1:

PR #4, opened 2021-05-03
YJDH-40 | Preliminary REST API

klemmari1:
Add preliminary models and create a REST API for the applications using Django Rest Framework.

- Add factories for the models
- Add serializers for all the models.
- Add viewset for the applications.
- Add tests and remove test_dummy.


backend/applications/models.py
```diff
@@ -0,0 +1,105 @@
...
+        related_name="applications",
+        verbose_name=_("company"),
+    )
+    status = models.CharField(
+        max_length=64, verbose_name=_("status"), choices=ApplicationStatus.choices
+    )
+
+    invoicer_name = models.CharField(
+        blank=True,
+        null=True,
```
jopesy, line 23: Avoid `null=True` on `CharField`s and `TextField`s. Otherwise you'll end up having two possible values for "no data"; `NULL` or an empty string. (refs. https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.null)

This applies to the following fields on this model:
- invoicer_name
- invoicer_email
- invoicer_phone_number

backend/applications/models.py
```diff
@@ -0,0 +1,105 @@
...
+
+class SummerBill(HistoricalModel, UUIDModel):
+    application = models.ForeignKey(
+        Application,
+        on_delete=models.CASCADE,
+        related_name="summer_bills",
+        verbose_name=_("application"),
+    )
+    summer_bill_id = models.CharField(
+        max_length=256, blank=True, null=True, verbose_name=_("summer bill id")
```
jopesy, line 46: Same here, avoid using `null=True` on text-based fields. This applies to these fields:

- summer_bill_id
- contact_name
- work_postcode
- employee_name
- employee_school
- employee_ssn
- employee_phone_number

backend/applications/models.py
klemmari1: Good point, I'll change!

backend/applications/api/v1/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+            "company",
+            "summer_bills",
+        ]
+
+    def update(self, instance, validated_data):
+        validated_data.pop("company", None)
+        summer_bills_data = validated_data.pop("summer_bills", [])
+        update_or_create_summer_bills_from_api_data(summer_bills_data, instance)
+
+        return super(ApplicationSerializer, self).update(instance, validated_data)
```
frwickst, line 46: This can be simplified to `return super().update(instance, validated_data)`

backend/applications/api/v1/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+            "status",
+            "invoicer_name",
+            "invoicer_email",
+            "invoicer_phone_number",
+            "company",
+            "summer_bills",
+        ]
+
+    def update(self, instance, validated_data):
+        validated_data.pop("company", None)
```
frwickst: This could use a comment to better understand why this is being done. At least for a person that does not know the code base well, this is something that would need to be asked or looked up elsewhere to get answered.

backend/applications/tests/test_applications_api.py
```diff
@@ -0,0 +1,80 @@
...
+from rest_framework.reverse import reverse
+
+from applications.api.v1.serializers import ApplicationSerializer
+from applications.enums import ApplicationStatus
+from applications.tests.factories import ApplicationFactory
+
+
+@pytest.mark.django_db
+def test_applications_list(api_client):
+    ApplicationFactory()
```
frwickst, line 11: Just as a suggestion for the future. If you wanted to "clean up" this a tiny bit, then something like `pytest-factoryboy` could remove this one line and turn the function to `def test_applications_list(api_client, application):` that way you could use fixtures only in the function parameters instead of calling the factory in the test.

backend/companies/migrations/0001_initial.py
```diff
@@ -1,8 +1,9 @@
 # Generated by Django 3.2 on 2021-05-04 14:36
 
-from django.db import migrations, models
 import uuid
 
+from django.db import migrations, models
+
```
frwickst: This looks like a style change that could live in its own commit.

backend/applications/tests/test_applications_api.py
```diff
@@ -0,0 +1,80 @@
...
+
+    assert response.status_code == 200
+    assert response.data["invoicer_name"] == "test"
+
+
+@pytest.mark.django_db
+def test_application_patch(api_client):
+    application = ApplicationFactory()
+    application.status = ApplicationStatus.DRAFT
+    application.save()
```
frwickst, line 47: Can't the status be passed to the factory? Like `ApplicationFactory(status=ApplicationStatus.DRAFT)` ? I might be remembering wrong here however, so please correct me in that case 😄

frwickst (changes requested):
Overall this looks like a solid stats, good work @klemmari1 !

backend/applications/tests/test_applications_api.py
klemmari1: Yes it can :D I'll fix.

backend/applications/api/v1/serializers.py
klemmari1: Yea, I'll try to add a comment about this and refactor the logic to be a bit better.

backend/applications/tests/test_applications_api.py
klemmari1: Good point!

backend/companies/migrations/0001_initial.py
klemmari1: Yea, that slipped in there :D I'll merge this to the original commit.

backend/applications/api/v1/serializers.py
```diff
@@ -0,0 +1,55 @@
...
+            "summer_bills",
+        ]
+
+    def update(self, instance, validated_data):
+        request = self.context.get("request")
+        summer_bills_data = validated_data.pop("summer_bills", [])
+        if request and request.method == "PUT":
+            update_summer_bills_using_api_data(summer_bills_data, instance)
+
+        return super().update(instance, validated_data)
```
tuomas777, line 48: Seems like this whole update should be inside a transaction.

backend/applications/enums.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class ApplicationStatus(models.TextChoices):
+    DRAFT = "draft"
+    ADDITIONAL_INFORMATION_REQUESTED = "additional_information_requested"
+    ADDITIONAL_INFORMATION_PROVIDED = "additional_information_provided"
+    SUBMITTED = "submitted"
+    ACCEPTED = "accepted"
+    REJECTED = "rejected"
+    DELETED_BY_CUSTOMER = "deleted_by_customer"
```
tuomas777: As other strings in the project are translatable these could have translatable labels as well.

backend/applications/api/v1/serializers.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def to_representation(self, obj):
+        data = super().to_representation(obj)
+
+        request = self.context.get("request")
+        if request and request.method == "GET":
+            # Only show the related company on GET request since we do not want to update it manually.
```
tuomas777, line 55: It might be just me but I don't quite understand what this comment is saying 😄

backend/applications/api/v1/views.py
```diff
@@ -0,0 +1,14 @@
...
+from applications.models import Application
+
+
+class ApplicationViewSet(
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    viewsets.GenericViewSet,
+):
+    queryset = Application.objects.all()
```
tuomas777, line 13: This will probably need prefetching sooner than later, something like `Application.objects.select_related("company").prefetch_related("summer_bills")`

backend/applications/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import include, path
+from rest_framework import routers
+
+from applications.api.v1 import views
+
+app_name = "applications"
+
+router = routers.DefaultRouter()
```
tuomas777, line 8: Just a note that there should be only one `DefaultRouter` per API, so you will probably run into problems with this if you add other applications than this `applications` (HEHE) to the API.

backend/applications/api/v1/serializers.py
klemmari1: I'll try to restructure it a bit :D

Basically what I mean is that `company` is read-only.

klemmari1: Also I'm not sure if this is the correct way to implement this in the serializer, so that when reading applications, it would return the `company` but would not allow updating it.

tuomas777: You could also add it as a read-only field. One benefit would be that it would be discovered by automatic documentation generators.

klemmari1: Well that would be only reasonable :D Forgot that I can do that.

tuomas777 (approved):
LGTM, good job! 👍 🚀

PR #5, opened 2021-05-06
YJDH-57 | Add Company API

klemmari1:
YTJ integration will be used to automatically fetch data from YTJ API about the customer company and to create the company instance.

Add Company API for fetching the company info. For now it fetches the company data straight from the YTJ API but in the final implementation this will be done separately.

Refs YJDH-57


backend/companies/api/v1/views.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @transaction.atomic
+    def get(self, request: Request, format: str = None) -> Response:
+        if settings.MOCK_FLAG:
+            # This mocked section will be used for testing purposes for the frontend.
+            session_id = request.META.get("HTTP_SESSION_ID")
+            if session_id == "-1":
+                return self.ytj_api_error
+            else:
+                company, _ = Company.objects.get_or_create(**DUMMY_COMPANY_DATA)
```
frwickst, line 35: This is a personal preference here, but I would maybe split this out into its own function. Yes, it would eat up more lines of code, but it would add readability. The extra indentation and logic here seems to be something that could be avoided in the main get function. Could this be done like:

```
    @transaction.atomic
    def get_mock(self, request: Request, format: str = None) -> Response:
        # This mocked section will be used for testing purposes for the frontend.
        session_id = request.META.get("HTTP_SESSION_ID")
        if session_id == "-1":
          return self.ytj_api_error
        else:
          company, _ = Company.objects.get_or_create(**DUMMY_COMPANY_DATA)

        company_data = CompanySerializer(company).data

        return Response(company_data)


    @transaction.atomic
    def get(self, request: Request, format: str = None) -> Response:
        if settings.MOCK_FLAG:
            return get_mock(request, format)
        ...
```

backend/companies/tests/test_company_api.py
```diff
@@ -0,0 +1,95 @@
...
+    assert response.data == company_data
+    assert (
+        response.data["business_id"] == DUMMY_YTJ_RESPONSE["results"][0]["businessId"]
+    )
+
+
+@pytest.mark.django_db
+@override_settings(MOCK_FLAG=False)
+def test_get_company_from_ytj_results_in_error(api_client):
+    with mock.patch(
```
charn, line 70: You could optionally also use something like (requests-mock)[https://pypi.org/project/requests-mock/] for mocking the responses.

backend/companies/services.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def get_or_create_company_using_company_data(
+    company_data: dict, ytj_data: dict
+) -> Company:
+    """
+    Get or create a company instance using a dict of the company data and attach the ytj_data json
+    for the instance.
+    """
+    company, created = Company.objects.get_or_create(**company_data)
```
frwickst, line 12: Could this be rewritten as:
```
company, created = Company.objects.get_or_create(**company_data, defaults={"ytj_json": ytj_data})
return company
```

backend/companies/ytj/ytj_client.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class YTJClient:
+    """
+    https://avoindata.prh.fi/
+    """
+
+    timeout: int
+
+    def __init__(self, timeout: int = 30):
```
frwickst, line 12: Could the timeout be configurable as well? As in `settings.YTJ_TIMEOUT`?

frwickst (commented):
No real errors that I could find with a quick review, mostly things that I might consider as refactorable code. Your choice with what you do with it here.

backend/companies/api/v1/views.py
```diff
@@ -0,0 +1,59 @@
...
+
+        company_data = CompanySerializer(company).data
+
+        return Response(company_data)
+
+    @transaction.atomic
+    def get(self, request: Request, format: str = None) -> Response:
+        if settings.MOCK_FLAG:
+            return self.get_mock(request, format)
+        else:
```
frwickst, line 44: No need for the if-else statement here since you are returning if the `MOCK_FLAG` is set. That was one of the benefits of splitting it out was, that the indentation could disappear :)

backend/companies/api/v1/views.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+    @transaction.atomic
+    def get_mock(self, request: Request, format: str = None) -> Response:
+        # This mocked get method will be used for testing purposes for the frontend.
+        session_id = request.META.get("HTTP_SESSION_ID")
+        if session_id == "-1":
+            return self.ytj_api_error
+        else:
+            company, _ = Company.objects.get_or_create(**DUMMY_COMPANY_DATA)
```
frwickst, line 34: Else statement should not be needed here since the if-statement is returning

backend/companies/api/v1/views.py
klemmari1: Yep, good point! Fixed.

backend/companies/api/v1/views.py
klemmari1: Fixed

backend/kesaseteli/settings.py
```diff
@@ -136,6 +139,12 @@
     "TEST_REQUEST_DEFAULT_FORMAT": "json",
 }  # TODO: Replace with actual authentication & permissions.
 
+YTJ_BASE_URL = env.str("YTJ_BASE_URL")
+YTJ_TIMEOUT = env.str("YTJ_TIMEOUT")
+
+# Mock flag for testing purposes
+MOCK_FLAG = env.str("MOCK_FLAG")
```
quyenlq, line 146: `env.bool("YTJ_TIMEOUT")` ? 😄

backend/companies/api/v1/views.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+
+    @transaction.atomic
+    def get_mock(self, request: Request, format: str = None) -> Response:
+        # This mocked get method will be used for testing purposes for the frontend.
+        session_id = request.META.get("HTTP_SESSION_ID")
+        if session_id == "-1":
+            return self.ytj_api_error
+
+        company, _ = Company.objects.get_or_create(**DUMMY_COMPANY_DATA)
```
quyenlq, line 34: Would it work if we only build the model object using its constructor `Company(**DUMMY_COMPANY_DATA)`? Do we need to actually save the mock object to the database?

quyenlq (approved):
LGTM! 🤘

PR #6, opened 2021-05-07
YJDH-55: Show company data on UI

nenonja:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-55
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-56
Add very simple page that shows company data from the backend. If loading data fails, it retries three times. React-query is used to call backend rest-api: https://react-query.tanstack.com/overview
Test it here: https://kesaseteli-employer-6.test.kuva.hel.ninja/company
Also:
- component tests with react-query
- company page browser test
- fixed pre-commit hook problems


frontend/employer/src/backend-api/BackendAPIProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+      baseURL: getBackendUrl(),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+
+    axiosClient.interceptors.request.use((config) => {
+      // Read token for anywhere, in this case directly from localStorage
+      // eslint-disable-next-line scanjs-rules/identifier_localStorage
+      const token = localStorage.getItem('token');
```
melniiv, line 21: are we planning at some point to implement and use HttpOnly cookie instead of local storage implementation? I noticed this security leak in many Hki projects.

frontend/employer/src/hooks/useCompanyQuery.ts
```diff
@@ -0,0 +1,14 @@
+import useBackendAPI from 'employer/hooks/useBackendAPI';
+import Company from 'employer/types/company';
+import { useQuery, UseQueryResult } from 'react-query';
+
+export const COMPANY_URL = '/v1/company/';
```
melniiv, line 5: maybe we can have a separate constants file for urls, as i guess we will have many endpoints, so easy to manage

melniiv (approved):
looks good to me, and really well done with project architecture and structure!

frontend/employer/src/backend-api/BackendAPIProvider.tsx
nenonja: This is still completely open thing, I actually used this code as base: https://cretezy.com/2020/react-query-axios-authentication

frontend/employer/src/hooks/useCompanyQuery.ts
nenonja: Usage will be always with hooks, `useXXXQuery` so the endpoint can be hide inside then. On the other hand I still needed to export it to the test, so perhaps It could be a good idea to separate them to a contant/enum

frontend/employer/src/backend-api/BackendAPIProvider.tsx
nenonja: Final solution to authentication will be based to this example project (except using react-query instead of redux): https://github.com/City-of-Helsinki/example-ui-profile/tree/HP-702-two-login-types
I think thsi example has better solution for persisting token.

PR #9, opened 2021-05-17
Fix backend not running migrations in review & staging

klemmari1:


nenonja (approved):
Looks good and all tests are green!

PR #11, opened 2021-05-18
Yamls for triggering pipelines from github

SuviVappula:
These yamls extend the pipelines in azure and are used to trigger builds when changes are made to github


nenonja (approved):
Looks good to me but I dont know anythign about azure pipelines. It's good to realize though that there will be some changes to infrastructure and repo is renamed

SuviVappula:
Making some changes to this as there are now also the benefit project in the same repo. Also separating the pipelines to have a dev/test and staging/prod pipelines separately.

PR #25, opened 2021-05-24
YJDH-48 | Backend: User session handling of suomi.fi eauthorization

klemmari1:
## Description :sparkles:

Create a model (`EAuthorizationProfile`) for storing the tokens for the [suomi.fi e-authorization](https://www.suomi.fi/valtuudet) OAuth connections. Create a hourly job that removes all stale user sessions for Helsinki Profiili and suomi.fi valtuudet API.

## Issues :bug:

[YJDH-48](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-48)
[YJDH-66](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-66)
[YJDH-67](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-67)


backend/oidc/jobs/hourly/clear_user_sessions.py
```diff
@@ -0,0 +1,21 @@
+from django.utils import timezone
+from django_extensions.management.jobs import HourlyJob
+
+from oidc.models import EAuthorizationProfile, OIDCProfile
+from oidc.services import clear_eauthorization_profiles, clear_oidc_profiles
+
+
+class Job(HourlyJob):
```
haavikko, line 8: Need to set up cron so that the hourly jobs are executed.
How should it be set up? Can the similar setup be used as was used in Venepaikka to execute the jobs in a separate pod?

klemmari1: Yea, this is true. Have to investigate how it should be done. Venepaikka could be a good example, I'll check it out.

backend/oidc/models.py
```diff
@@ -43,3 +41,27 @@ def is_active_refresh_token(self):
...
+
+
+class EAuthorizationProfile(AbstractOIDCProfile):
+    oidc_profile = models.OneToOneField(
+        OIDCProfile,
+        on_delete=models.CASCADE,
+        verbose_name=_("oidc_profile"),
+        related_name="eauthorization_profile",
+    )
+    company = models.OneToOneField(
```
haavikko, line 62: Would EAuthorizationProfile be reusable for helsinki-benefit too? For example, I remember discussing with @quyenlq that he had plans to reuse at least some of the code for Company model, but e.g. if there's an abstract Company base then we can't have foreign key to that?

quyenlq: True, it'd be great if Helsinki-benefit can reuse part of the Company and this model too.
@klemmari1  Would it be possible for you to make the Company model abstract and inherit it later in Kesaseteli app with its own model. We can discuss later which fields should be included in the abstract Company class later

.github/workflows/staging.yml
```diff
@@ -115,6 +115,19 @@ jobs:
           SLACK_MESSAGE: Deploying to ${{ matrix.url }} has *FAILED!*
           SLACK_CHANNEL: yjdh-alerts
         if: failure()
+
+      - name: Deploy Hourly Cronjobs
+        uses: City-of-Helsinki/setup-cronjob-action@main
+        with:
+          image_repository: ghcr.io/city-of-helsinki/${{ github.event.repository.name }}
```
quyenlq, line 122: It's recommended to give a name to a cronjob, as it's become mandatory later when we have multiple jobs in one namespace. https://github.com/City-of-Helsinki/setup-cronjob-action

backend/oidc/models.py
klemmari1: True that. I did not think it quite through but yes it should be abstracted.

backend/oidc/services.py
```diff
@@ -17,6 +18,16 @@ def clear_oidc_profile(oidc_profile: OIDCProfile) -> None:
     oidc_profile.delete()
 
 
+def clear_oidc_profiles(oidc_profiles: "QuerySet[OIDCProfile]") -> None:
```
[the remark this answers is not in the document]
klemmari1: This one takes a queryset as the other takes a single OIDCProfile. But it has filtering so I'll fix that.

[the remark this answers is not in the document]
klemmari1: On another thought, those can be merged into one util function.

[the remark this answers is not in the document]
klemmari1: OK, fixed now.

quyenlq (approved):
👌  looking good!

PR #31, opened 2021-05-24
Yjdh-92: Frontend Monorepo Structure

nenonja:
## Description :sparkles:
Frontend monorepo structure: `Kesaseteli frontedservices are now under `frontend/kesaseteli` folder. Also:
- `.env` file is renamed to `.env.kesaseteli`
- `docker-compose.yml` renamed to `docker-compose.kesaseteli.yml`
- npm audit added and audit problems fixed (not part of the original pr but did it while waiting long builds :)
 
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-92


.github/workflows/review.yml
```diff
@@ -101,15 +99,9 @@ jobs:
           DOCKER_IMAGE_NAME: ${{ matrix.service }}
           PROJECT_NAME: ${{ github.event.repository.name }}-${{ matrix.service }}
           K8S_SECRET_VERSION: ${{ github.sha }}
-          VAULT_JWT_PRIVATE_KEY: ${{ secrets.VAULT_ACCESS_PRIVATE_KEY_REVIEW }}
-          VAULT_ADDR: ${{ secrets.VAULT_ADDR }}
-          VAULT_KV_VERSION: "2"
-          VAULT_JWT_AUTH_PATH: ${{ github.event.repository.name }}-${{ matrix.service }}-review
-          VAULT_KV_SECRET_MOUNT_POINT: review
-          SERVICE_PORT: ${{ matrix.port }}
```
tuovinensanttu, line 109: `SERVICE_PORT` should not be removed. Without this, the value will default to `8000` (for example nextjs serves on port 3000 => readiness and health checks fail).

nenonja: Thanks! I forgot about that. Whar about vault values? I think they should be removed only temporary until auth secret is changed? Or is it just auth-path that should point to a old repo's value?

tuovinensanttu: You can try to add those back. Previously the `{{ matrix.service }}` was either  `backend` or `employer` now that the names are changed it should be able to find correct path.

nenonja: seem to work now for vault values

melniiv (approved):
looks good

PR #32, opened 2021-05-24
YJDH-80 | Backend: Add mock setup for Helsinki Profile auth flow

klemmari1:
## Description :sparkles:

For testing purposes mock the oidc auth flow views so that it is not required to call Helsinki Profile.

## Issues :bug:

[YJDH-80](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-80)

### DIFF
--- a/backend/oidc/tests/test_mock_views.py
+++ b/backend/oidc/tests/test_mock_views.py
@@ -1,0 +1,70 @@
+import pytest
+from django.conf import settings
+from django.test import override_settings
+from django.urls import path, reverse
+
+from oidc.mock_views import (
+    MockAuthenticationRequestView,
+    MockLogoutView,
+    MockUserInfoView,
+)
+
+urlpatterns = [
+    path(
+        "oidc/authenticate/",
+        MockAuthenticationRequestView.as_view(),
+        name="oidc_authentication_init",
+    ),
+    path(
+        "oidc/logout/",
+        MockLogoutView.as_view(),
+        name="oidc_logout",
+    ),
+    path(
+        "oidc/userinfo/",
+        MockUserInfoView.as_view(),
+        name="oidc_userinfo",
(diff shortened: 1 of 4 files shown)


backend/oidc/urls.py
```diff
@@ -1,17 +1,44 @@
...
-    path(
-        "userinfo/",
-        HelsinkiOIDCUserInfoView.as_view(),
-        name="oidc_userinfo",
-    ),
-    path("", include("mozilla_django_oidc.urls")),
-]
+urlpatterns = []
+
+if settings.MOCK_FLAG:
```
haavikko, line 13: could the flag name be more verbose, like AUTH_MOCK_FLAG?

Do we need to set default value (False) for the flag in settings.py?

haavikko (approved):
Left a comment about the flag name and value in settings, address that as you see appropriate, other than that, looks good to me :-)

backend/oidc/urls.py
haavikko: hmm on the second thought, it is fine to use the same MOCK_FLAG that the company api already uses

backend/oidc/tests/test_mock_views.py
```diff
@@ -0,0 +1,72 @@
...
+    response = client.post(logout_url)
+
+    assert response.url == settings.LOGOUT_REDIRECT_URL
+    assert "_auth_user_id" not in client.session  # User not authenticated
+
+
+@pytest.mark.django_db
+@override_settings(MOCK_FLAG=True, ROOT_URLCONF=__name__)
+def test_userinfo_view(client, user):
+    client.force_login(user)
```
quyenlq, line 63: I believe we already have fixture for this :)

quyenlq (approved):
LGTM!

PR #37, opened 2021-05-25
YJDH-94: Support Arbitrary User IDs for OpenShift

nenonja:
Support Arbitrary User IDs for OpenShift
https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-94

### DIFF
--- a/frontend/kesaseteli/employer/Dockerfile
+++ b/frontend/kesaseteli/employer/Dockerfile
@@ -78,8 +78,8 @@ COPY --from=staticbuilder --chown=appuser:appuser app/shared/package.json shared
 COPY --from=staticbuilder --chown=appuser:appuser app/shared/src/server/next-server.js shared/src/server/
 
 # Sets the directory and file permissions to allow users in the root group to access them in the built image (for OpenShift)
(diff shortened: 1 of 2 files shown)


mikkovihonen:
@nenonja I think appuser should be added to the group as well. Otherwise this should work nicely.

frontend/kesaseteli/employer/Dockerfile
```diff
@@ -72,13 +73,21 @@ COPY --from=staticbuilder --chown=appuser:appuser app/kesaseteli/employer/next.c
...
+
 # Install production dependencies
 RUN yarn install --production --frozen-lockfile --check-files
 
+# Sets the directory and file permissions to allow users in the root group to access them in the built image (for OpenShift)
+# https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html
+USER root
+RUN chgrp -R 0 . \
+  && chmod -R g+rwX .
+USER appuser
```
mikkovihonen, line 90: appuser should be added to the group as well.

PR #38, opened 2021-05-25
YJDH-95 | Backend: Kesaseteli monorepo structure

klemmari1:
## Description :sparkles:

- Restructure kesaseteli to the monorepo setup.
- Modify github flow & other files to match the monorepo structure.


.github/dependabot.yml
```diff
@@ -11,6 +11,6 @@ updates:
     schedule:
       interval: "daily"
   - package-ecosystem: "pip"
-    directory: "backend/"
+    directory: "backend/kesaseteli"
```
nenonja, line 14: why to limit on `kesaseteli` folder here? if there will be `backend/benefit` it does not get scanned.

nenonja: I fixed it @klemmari1

.pre-commit-config.yaml
```diff
@@ -5,17 +5,17 @@ repos:
     rev: 21.5b1
     hooks:
       - id: black
-        files: ^backend/
+        files: ^backend/kesaseteli
```
quyenlq, line 8: I think we can let pre-commit runs on the backend folder and move the `setup.cfg` to a shared location (e.g backend/shared/precommit-setup.cfg)

.github/workflows/staging.yml
```diff
@@ -39,8 +39,8 @@ jobs:
         service: [ 'backend', 'employer' ]
         include:
           - service: backend
-            context: ./backend
-            dockerfile: ./backend/Dockerfile
+            context: ./backend/kesaseteli
```
quyenlq, line 42: I think we should limit this workflow to only run when there are changes in kesaseteli folder. In helsinki-benefit we use something like this: https://github.com/City-of-Helsinki/yjdh/pull/46/files#diff-280994542061213299f64fa399afda4ec4a787875fe7953d3595900eb09cbc40R6

.github/workflows/review.yml
```diff
@@ -38,8 +38,8 @@ jobs:
         service: [ 'kesa-backend', 'kesa-employer' ]
         include:
           - service: kesa-backend
-            context: ./backend
-            dockerfile: ./backend/Dockerfile
+            context: ./backend/kesaseteli
```
quyenlq, line 41: I think we should limit this workflow to only run when there are changes in kesaseteli folder. In helsinki-benefit we use something like this: https://github.com/City-of-Helsinki/yjdh/pull/46/files#diff-280994542061213299f64fa399afda4ec4a787875fe7953d3595900eb09cbc40R6

.github/workflows/pytest.yml
```diff
@@ -47,10 +47,10 @@ jobs:
             ${{ runner.os }}-
 
       - name: Install dependencies
-        run: pip install -r backend/requirements.txt -r backend/requirements-dev.txt codecov
+        run: pip install -r backend/kesaseteli/requirements.txt -r backend/kesaseteli/requirements-dev.txt codecov
```
quyenlq, line 50: I think we should limit this workflow to only run when there are changes in kesaseteli folder. In helsinki-benefit we use something like this: https://github.com/City-of-Helsinki/yjdh/pull/46/files#diff-280994542061213299f64fa399afda4ec4a787875fe7953d3595900eb09cbc40R6

quyenlq (commented):
One thing that might benefit us is naming the workflow with the project name prefix, then it should be easier to find the correct workflow in this page: https://github.com/City-of-Helsinki/yjdh/actions :D

.github/workflows/review.yml
```diff
@@ -63,17 +63,15 @@ jobs:
...
         include:
-          - service: kesa-backend
-            project: kesaseteli-backend
-            context: backend
-            dockerfile: ./backend/Dockerfile
+          - service: ks-bknd
+            context: ./backend/kesaseteli
+            dockerfile: ./backend/kesaseteli/Dockerfile
             database: true
             vault_auth: ${{ github.event.repository.name }}-kesaseteli-backend-review
```
tuovinensanttu, line 72: `vault_auth` is no longer used so it can be removed.

backend/shared/precommit-setup.cfg
```diff
@@ -0,0 +1,32 @@
...
+exclude = *migrations*
+ignore = E309
+
+[flake8]
+exclude = migrations
+max-line-length = 120
+max-complexity = 10
+
+[tool:pytest]
+DJANGO_SETTINGS_MODULE = kesaseteli.settings
```
haavikko, line 12: Good idea to share the config.
The file is in shared folded but references kesaseteli.settings. How to work around this?

klemmari1: I think the `tool:pytest` section can be removed from the common setup.cfg for now since it is not used in the pre-commit hook. I guess the best option is to still have the individual `setup.cfg` in each folder as well.

tuovinensanttu (approved):
Looks good to me 👍

PR #39, opened 2021-05-25
WIP: testing review pipeline

tuovinensanttu:
Testing review pipeline


nenonja (approved):
Is this ready now? It looks good to me. Acceptance test failure is not related and original problem seems to be solved.

tuovinensanttu:
I guess so, the helm secret naming problem might re-appear, but let's follow the situation 👍

PR #50, opened 2021-05-31
HL-16: Application lists designs

melniiv:
## Description :sparkles:
- UI for application Drafts and Submitted lists in the frontpage 

Tests and logic implemented when the backend is ready


nenonja (commented):
Perhaps additions to docker-compose.benefit.yml could be added. Now I cannot get the environment working locally (neither running yarn dev nor build), so running in docker should show does it really work or not.

PR #51, opened 2021-05-31
Change OIDC logout method from GET -> POST

klemmari1:
## Description :sparkles:

- Modify the OIDC logout view to support POST requests instead of GET.
- Return a HttpResponse with status 200 instead of redirect.


nenonja (approved):
I dont have django experience but I cherry-picked the commit and logout did work without problems, so approvance from me.

nenonja (approved):

PR #52, opened 2021-05-31
YJDH-65: Logout functionality

nenonja:
## Description :sparkles:
Logout functionality. Ability to logout after login. Session will be deleted and react-query data cleared. Company page is only available when logged in. Component tests and browser tests included.
**note**: Most of the authentication logic will be moved under shared folder. This will be done in separate task.
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-65


frontend/kesaseteli/employer/src/pages/index.tsx
```diff
@@ -29,12 +27,11 @@ const EmployerIndex: NextPage = () => {
 
   return (
     <Layout headingText="Työnantajan liittymä">
-      {isLoading && <p>Ladataan...</p>}
       {user && <p>Tervetuloa {user.name}!</p>}
       <Button onClick={onLogout} disabled={isLoading}>
-        Kirjaudu ulos
+        {isLoading ? 'Kirjaudutaan ulos...' : 'Kirjaudu ulos'}
```
melniiv, line 32: shouldnt it be IsLoadingLogout instead of isLoading?

frontend/shared/.eslintrc.js
```diff
@@ -20,5 +20,11 @@ module.exports = {
         'import/prefer-default-export': 'off',
       },
     },
+    {
+      files: ['*.components.ts'],
+      rules: {
+        'security/detect-non-literal-fs-filename': 'off',
```
melniiv, line 26: which file names are giving this errors?

melniiv (approved):

frontend/kesaseteli/employer/src/pages/index.tsx
nenonja: Most likely, I will try that it looks ok and text does not change during loading or anyhing.

frontend/shared/.eslintrc.js
nenonja: testcafe checks like:
```
        await t
          .expect(selectors.userName(name).exists)
          .ok(await getErrorMessage(t));
```
Because eslint thinks `exists` is related to use of `fs` but it isnt. Thus the rule is disabled for `*.components.ts` which are testcafe helper files.

PR #53, opened 2021-06-01
YJDH-106 | Fix kesäseteli review pipeline

klemmari1:

### DIFF
--- a/.github/workflows/bf-review.yml
+++ b/.github/workflows/bf-review.yml
@@ -4,6 +4,7 @@ on:
     paths:
     - 'backend/helsinki-benefit/**'
     - 'frontend/helsinki-benefit/**'
+    - 'frontend/shared/**'
     - '.github/workflows/bf-review.yml'
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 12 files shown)


.github/workflows/ks-review.yml
```diff
@@ -1,3 +1,4 @@
+# TODO: Fix this workflow
```
nenonja, line 1: todo can be removed

nenonja (commented):
Why frontend unit tests are not run at all? Perhaps because no changes to the frontend folder?
I can check it

.github/workflows/bf-review.yml
```diff
@@ -4,6 +4,8 @@ on:
     paths:
     - 'backend/helsinki-benefit/**'
     - 'frontend/helsinki-benefit/**'
+    - 'frontend/shared/**'
+    - 'frontend/*'
```
quyenlq, line 8: Hmn... Wouldn't this will trigger the Helsinki Benefit build when there is a change in `frontend/kesaseteli` ?

nenonja: as far as I understand it should not include subfolders: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

quyenlq: Ah, yeah, now I get it :) Thanks for the clarification

nenonja (approved):
ks-empl keeps failing to infamous pages directory error but otherwise I think this is good now and ready to merge

PR #60, opened 2021-06-02
KS-Backend | Update the OIDC mock view urls

klemmari1:
## Description :sparkles:

Update the url paths to match the original view paths.


haavikko (approved):
OK go ahead :-)

PR #62, opened 2021-06-02
YJDH-69 | Preliminary setup for e-authorizations implementation

klemmari1:
## Description :sparkles:

Features in this PR:
1. Decouple Company model from the EAuthorizationProfile model because the coupling is not really needed. The business id can be quieried using the organization roles endpoint when that is implemented in `oidc.utils.get_organization_roles`.
2. Refactor OIDC methods to support the upcoming changes.
3. Add OIDC views to their own package.
4. Create an env variable for the base OIDC OP url. Replace all the separate OIDC OP endpoint urls with a single
base url and use that to build the other urls.


.env.kesaseteli.example
```diff
@@ -6,12 +6,8 @@ HANDLER_URL=http://localhost:3100
 NEXT_PUBLIC_BACKEND_URL=http://localhost:8000
 
 # Authentication
+OIDC_OP_BASE_URL=https://tunnistus.test.hel.ninja/auth/realms/helsinki-tunnistus/protocol/openid-connect
```
haavikko, line 9: Feels like much easier to understand this way :+1:

backend/kesaseteli/oidc/tests/test_mock_views.py
```diff
@@ -3,7 +3,7 @@
 from django.test import override_settings
 from django.urls import path, reverse
 
-from oidc.mock_views import (
+from oidc.views.mock_views import (
```
haavikko, line 6: good idea to split into two modules

haavikko (approved):
Looks good to me... @quyenlq may want to check the changes with Company model (unless he already has), as he's working on the Company data model for helsinki-benefit

backend/kesaseteli/companies/models.py
```diff
@@ -1,10 +1,10 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
-from oidc.models import AbstractCompany
+from utils.models import UUIDModel
 
 
-class Company(AbstractCompany):
+class Company(UUIDModel):
```
quyenlq, line 7: Hmn... I thought we will share the AbstractCompany model with all the fields below? Or this will be the next step after this PR

klemmari1: I'll create another PR for that. Adding all the required stuff to shared.

quyenlq (approved):

PR #67, opened 2021-06-03
Renaming helsinki-benefit to benefit

quyenlq:
## Description :sparkles:
Renaming BE folder from `helsinki-benefit` to `benefit` for consistency.
Docker files and workflows are also updated accordingly


haavikko (approved):
This is OK by me. I'm a bit ambivalent on this - Yes, I'd like to use a shorter name too (and avoid the dash in the name), but using just "benefit" as the name is losing information - it's a generic term and there are so many different benefits. Whereas "Helsinki-benefit" is the actual name of this benefit. But since both you and Ivan are in favor of the change, I say let's go for it :-)

PR #68, opened 2021-06-03
Add setup.py for shared component, update requirement.txt to install package from repo folder

quyenlq:
## Description :sparkles:
This PR enables us to use the `shared` folder as a package, but the Django applications install this `shared` package locally from the repo, so any changes to the `shared` folder will affect immediately without release the package to external package distributor.

I added some code that I think can be shared already between two BE projects.

After this PR we can start contributing to the `shared` component and both projects can get the benefit from it immediately.


backend/shared/shared/models/abstract_models.py
```diff
@@ -0,0 +1,31 @@
...
+        abstract = True
+
+
+class AbstractCompany(models.Model):
+    """
+    TODO: Shared company fields go here
+    """
+
+    class Meta:
+        abstract = True
```
haavikko, line 31: The TimeStampedModel could be shared with Kesäseteli as well, I'd like to use it with Application model

```
class TimeStampedModel(models.Model):
    created_at = models.DateTimeField(auto_now_add=True, verbose_name=_("time created"))
    modified_at = models.DateTimeField(auto_now=True, verbose_name=_("time modified"))

    class Meta:
        abstract = True

```

haavikko (approved):
Looks good to me, before merging if you have time please add TimeStampedModel, unless there's reason not to :-)

klemmari1 (approved):
Looks good to me!

backend/shared/shared/models/abstract_models.py
quyenlq: Sure, I'll add it

PR #70, opened 2021-06-04
HL-36: data model for applications

haavikko:
## Description :sparkles:
* Application fields and some supporting models
* Add django-localflavor to requirements (IBANField)
* Add simple_history.middleware.HistoryRequestMiddleware to MIDDLEWARE so that the request's user is saved in history data
* some config like COMMERCIAL_COMPANY_FORMS could be moved to Company model
* more tests coming in another PR for rest api
*


backend/benefit/applications/enums.py
```diff
@@ -0,0 +1,19 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class ApplicationStatus(models.TextChoices):
+    DRAFT = "draft", _("Draft")
+    RECEIVED = "received", _("Receive")
```
klemmari1, line 7: _("Received")

backend/benefit/applications/tests/conftest.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from applications.tests.factories import ApplicationFactory, UserFactory
+from django.contrib.auth.models import Permission
+from rest_framework.test import APIClient
+
+
+@pytest.fixture
+def api_client():
```
klemmari1, line 8: Something I'm going to fix in kesäseteli is to change this as:
```
@pytest.fixture
def user():
    return UserFactory()


@pytest.fixture
def api_client(user):
    permissions = Permission.objects.all()
    user.user_permissions.set(permissions)
    client = APIClient()
    client.force_authenticate(user)
    return client
```
so that you can use the same user in the test if needed.

backend/benefit/applications/models.py
```diff
@@ -0,0 +1,243 @@
...
+                              verbose_name=_("ahjo batch"),
+                              related_name="applications",
+                              null=True,
+                              on_delete=models.SET_NULL,
+    )
+    """
+
+    bases = models.ManyToManyField("ApplicationBasis", related_name="applications")
+
+    history = HistoricalRecords(table_name="applications_application_history")
```
klemmari1, line 182: Is there a reason to set this up manually instead of using the `shared.shared.models.abstract_models.HistoricalModel`?

backend/benefit/helsinkibenefit/settings.py
```diff
@@ -107,6 +109,7 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "simple_history.middleware.HistoryRequestMiddleware",
```
klemmari1, line 112: Oh, I think I forgot to add this into kesäseteli. Kudos for reading the documentation :grin:

backend/benefit/applications/models.py
haavikko: I wanted to set the table_name for it, since I need to add the prefix (which, I notice now,  is missing ... fixed)

backend/benefit/applications/enums.py
haavikko: fixed

backend/benefit/applications/tests/conftest.py
haavikko: ok, changed the code to match kesäseteli code. Because the user model is different I guess it's not worth it to make these fixtures shared between projects

backend/benefit/helsinkibenefit/settings.py
haavikko: :-)

backend/benefit/applications/models.py
klemmari1: Ah yep, makes sense now.

quyenlq (approved):
Great work! Very detail comments in the models 👍

backend/benefit/utils/debug_util.py
```diff
@@ -0,0 +1,35 @@
+import functools
+import pdb
+import traceback
+
+from rest_framework.views import exception_handler
+
+
+def debug_on_exception(func):
```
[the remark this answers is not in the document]
klemmari1: +1

klemmari1 (approved):
LGTM!

quyenlq:
In case you might not notice, pytest failed because of a migration error

backend/benefit/applications/models.py
```diff
@@ -0,0 +1,262 @@
...
+        max_length=64, verbose_name=_("granter of the de minimis aid")
+    )
+    amount = models.DecimalField(
+        max_digits=8, decimal_places=2, verbose_name=_("amount of the de minimis aid")
+    )
+    granted_at = models.DateField(
+        verbose_name=_("benefit granted at"), blank=True, null=True
+    )
+
+    history = HistoricalRecords(table_name="applications_deminimisaid_history")
```
quyenlq, line 210: Should it be `bf_ applications_deminimisaid_history` too?

PR #72, opened 2021-06-07
Add users app, define AUTH_USER_MODEL

haavikko:
## Description :sparkles:

Before upgrading locally, must redo admin migrations because they depend on AUTH_USER_MODEL
python manage.py migrate admin zero
Pull latest code
python manage.py migrate


quyenlq (approved):
Looking good, there is a small lint issue though, approved already

backend/benefit/users/models.py
```diff
@@ -0,0 +1,8 @@
+import uuid
+
+from django.contrib.auth.models import AbstractUser
+from django.db import models
+
+
+class User(AbstractUser):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
quyenlq, line 8: Hmn, I forgot to ask, should we add a prefix to the db table name?

haavikko: Good catch! Added "bf_" prefix so table name would be "bf_users_user". (I *think* this was the prefix mentioned in Slack chat, but I can't find the message - if we had agreed on a different prefix, then ping me?

quyenlq: yes it is the bf_ prefix :)

PR #73, opened 2021-06-07
HL-15,16,44,46: Layout, stepper, generic form and wizard

melniiv:
## Description :sparkles:
- generic stepper implemented with independent action buttons
- styles and common components for forms
- form implementation for applicant step 1
- applicant wizard steps with steps with sections and basic translations


frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
```diff
@@ -0,0 +1,59 @@
...
+    handleSubmit,
+    handleBack,
+    t,
+    steps,
+    currentStep,
+    hasBack,
+    hasNext,
+  } = useComponent();
+
+  const DynamicFormStepComponent: ComponentType<DynamicFormStepComponentProps> = dynamic(
```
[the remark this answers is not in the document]
nenonja: Also This one is simple: https://codesandbox.io/s/github/devrnt/react-use-wizard/tree/main/examples/nextjs?file=/pages/index.js:1050-1054
You dont have to give hasBack and hasNext params. It does not preserve the step number if reloaded as the solution in previous comment

frontend/benefit/applicant/.eslintrc.js
```diff
@@ -3,6 +3,8 @@ const { join } = require('path');
 module.exports = {
   extends: ['auto'],
   rules: {
+    'no-secrets/no-secrets': ['error', { tolerance: 4.2 }],
+    'no-unsanitized/method': 'off',
     'no-void': 'off',
```
nenonja, line 8: i would still enable no-void rule, you have still added the void to several places: https://github.com/City-of-Helsinki/yjdh/pull/73/files#diff-3e08260afd1fc2ba62e1377185fd239d85426f2b80622e2a103b375a4c8065dfR113

frontend/benefit/applicant/src/components/applications/forms/application/step1/ApplicationFormStep1.tsx
```diff
@@ -0,0 +1,343 @@
...
+            onChange={formik.handleChange}
+            onBlur={formik.handleBlur}
+            aria-invalid={
+              !!getErrorMessage(APPLICATION_FIELDS.HAS_COMPANY_OTHER_ADDRESS)
+            }
+          />
+        </StyledFormGroup>
+        {formik.values.hasCompanyOtherAddress && (
+          <StyledFormGroup>
+            <TextInput
```
nenonja, line 85: Some repetition in these TextInputs, perhaps they could be refactored per field. But as this is not final result, this can be done later.

frontend/benefit/applicant/src/components/applications/forms/application/step1/extended.ts
```diff
@@ -0,0 +1,130 @@
...
+  [APPLICATION_FIELDS.CONTACT_PERSON_FIRST_NAME]: string;
+  [APPLICATION_FIELDS.CONTACT_PERSON_LAST_NAME]: string;
+  [APPLICATION_FIELDS.CONTACT_PERSON_PHONE]: string;
+  [APPLICATION_FIELDS.CONTACT_PERSON_EMAIL]: string;
+  [APPLICATION_FIELDS.DE_MINIMIS_AIDS_GRANTED]: string;
+  [APPLICATION_FIELDS.COLLECTIVE_BARGAINING_ONGOING]: string;
+  [APPLICATION_FIELDS.COLLECTIVE_BARGAINING_INFO]: string;
+};
+
+const useComponent = (): ExtendedComponentProps => {
```
nenonja, line 39: There is also some logic that could be extracted to common hook like useStep or something that takes the applicaton fields as parameter and returns common  functionality. But it can be done when starting doing second step.

frontend/benefit/applicant/src/constants.ts
```diff
@@ -1,5 +1,11 @@
 export const IS_CLIENT = typeof window !== 'undefined';
 
+export enum ROUTES {
```
nenonja, line 3: You can actually combine the locale and the route, for example:
http://localhost:3000/fi/application/new
http://localhost:3000/sv/application/new
Seem to work correctly out of the box.
On the other hand The problem in the language selector in the header is that it does not change the locale path parameter, so page refresh always defaults back to finnish,

nenonja (changes requested):
Great work in such a short time! The form pages and styles works as designed and didnt find any flaws from them. It seems also that you are very well oriented to work with formik so that's probably good way to go, even if I'm going to use reack-hook-form with Kesaseteli.
There is some comments, at least one thing than needs to be changed is the next/dynamic thing....
Also one important fix for the languageselector, but overall languages are working quite well.

frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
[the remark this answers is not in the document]
melniiv: i don't like the idea of the routing in wizard to be honest. that's overcomplication. preserving the state does not give you any advantage. You have to store the current step in backend anyway, on refresh, on edit application, on new application, etc. I didnt actually know that that the dynamic import does not work with literal strings. But in anyway, it can be easily replaced

frontend/benefit/applicant/src/components/applications/forms/application/step1/ApplicationFormStep1.tsx
melniiv: yes, there is repetition, i agree. but we are not going to create the form generator for one form. depending on field types there are different parameters, more generic you think, more overcomplication you get. Defining fields one by one, even with repetitive values, gives readability, compared to some config object you are looping through. if that was your concern.

frontend/benefit/applicant/.eslintrc.js
melniiv: i ll fix

frontend/benefit/applicant/.eslintrc.js
```diff
@@ -3,6 +3,8 @@ const { join } = require('path');
 module.exports = {
   extends: ['auto'],
   rules: {
+    'no-secrets/no-secrets': ['error', { tolerance: 4.2 }],
+    'no-unsanitized/method': 'off',
```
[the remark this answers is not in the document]
melniiv: yes, i got it, about the literal string

frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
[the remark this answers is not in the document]
nenonja: It is enough for now just to fix the dynamic import. The state of the wizard must be saved somewhere if you want to preserve it after refresh. Number of the step in the url is easiest way imho. It can also be a queryparam or hash I guess...

frontend/benefit/applicant/src/components/applications/forms/application/step1/ApplicationFormStep1.tsx
nenonja: If you feel like that then it's better to keep that way 👍

frontend/benefit/applicant/src/constants.ts
nenonja: You still have to save the selected language to the application state somehow if you want to preserve it for page refresh. If it's not in the url, then perhaps redux+localstorage is another option? I would prefer the url.

frontend/benefit/applicant/.eslintrc.js
nenonja: Oh I get mixed up. `'no-void' : 'off'` is good idea and use the `void` to resolve `@typescript-eslint/no-floating-promises` when floating promise is expected. So no need to do anything about it.

nenonja (approved):
Approved as most important comments are addressed. I didnt check the latest style changes very rigorously but look-and-feel of the ui seems all good to me.

PR #74, opened 2021-06-07
YJDH-69 | Eauthorizations OIDC flow

klemmari1:
## Description :sparkles:

Run local django server with SSL sertificate to get https callback urls that are required by suomi.fi eauthorizations. Also running the localhost as https also helps to bring the local environment closer to staging / production.

Add functionality for the suomi.fi eauthorizations flow. Upon successful helsinki profiili authentication, redirect the user straight to the suomi.fi valtuudet authorization init view that first handles the user registration for the valtuudet API and then redirects the user to the valtuudet authentication flow to choose the company on whose behalf they are doing business.

After receiving the "code" in EauthAuthenticationCallbackView, switch that to the auth tokens from the suomi.fi token endpoint. Save the tokens to EAuthorizationProfile for later use and to fetch the organization roles (and the business id) with `oidc.utils.get_organization_roles`.

Refs YJDH-69

## Issues :bug:

[YJDH-69](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-69)

TODO:
Use `oidc.utils.get_organization_roles` function to fetch the company business id to be used with the YTJ integration.

### DIFF
--- a/.env.kesaseteli.example
+++ b/.env.kesaseteli.example
@@ -1,9 +1,10 @@
 DEBUG=1
 APPLY_MIGRATIONS=1
 CREATE_SUPERUSER=1
+CORS_ORIGIN_ALLOW_ALL=1
 EMPLOYER_URL=http://localhost:3000
 HANDLER_URL=http://localhost:3100
-NEXT_PUBLIC_BACKEND_URL=http://localhost:8000
+NEXT_PUBLIC_BACKEND_URL=https://localhost:8000
 
 # Authentication
 OIDC_OP_BASE_URL=https://tunnistus.test.hel.ninja/auth/realms/helsinki-tunnistus/protocol/openid-connect
@@ -11,6 +12,11 @@ OIDC_RP_CLIENT_ID=
 OIDC_RP_CLIENT_SECRET=
 LOGIN_REDIRECT_URL=http://localhost:3000/
 LOGIN_REDIRECT_URL_FAILURE=http://localhost:3000/login?error=true
+
+EAUTHORIZATIONS_BASE_URL=https://asiointivaltuustarkastus.test.suomi.fi
+EAUTHORIZATIONS_CLIENT_ID=
+EAUTHORIZATIONS_CLIENT_SECRET=
+EAUTHORIZATIONS_API_OAUTH_SECRET=
 
 # Mock integrations
 MOCK_FLAG=1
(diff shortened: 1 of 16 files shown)


backend/kesaseteli/docker-entrypoint.sh
```diff
@@ -35,7 +35,7 @@ fi
 if [[ ! -z "$@" ]]; then
     "$@"
 elif [[ "$DEV_SERVER" = "1" ]]; then
-    python ./manage.py runserver 0.0.0.0:8000
+    python ./manage.py runserver_plus 0.0.0.0:8000 --cert-file /tmp/cert.crt
```
haavikko, line 38: Also update README.md?

.env.kesaseteli.example
```diff
@@ -3,11 +3,16 @@ APPLY_MIGRATIONS=1
...
+NEXT_PUBLIC_BACKEND_URL=https://localhost:8000
 
 # Authentication
 OIDC_OP_BASE_URL=https://tunnistus.test.hel.ninja/auth/realms/helsinki-tunnistus/protocol/openid-connect
 OIDC_RP_CLIENT_ID=
 OIDC_RP_CLIENT_SECRET=
 LOGIN_REDIRECT_URL=http://localhost:3000/
 LOGIN_REDIRECT_URL_FAILURE=http://localhost:3000/login?error=true
+
+EAUTHORIZATIONS_BASE_URL=https://asiointivaltuustarkastus.test.suomi.fi
```
haavikko, line 15: The Finnish language has good support for generating hostnames :-)

haavikko (commented):
Only started, need to continue later, sprint retro starting soon

quyenlq (approved):
LGTM! 👌 Hope this can be merged to shared component soon

backend/kesaseteli/docker-entrypoint.sh
haavikko: Just noticed there's another PR for nginx ssl termination (https://helsinkisolutionoffice.atlassian.net/browse/YJDH-109)

haavikko (approved):
Thanks for the walkthrough in slack call

backend/kesaseteli/docker-entrypoint.sh
klemmari1: Yes. We're probably going to switch to that instead of this runserver_plus. The nginx setup still needs a bit of testing so we'll transfer to that when it is tested and proven to work well.

PR #82, opened 2021-06-09
YHDH-109: Local https termination

mikkovihonen:
## Description :sparkles:

ADDS Local nginx with cert autogeneration for kesaseteli

## Issues :bug:

YHDH-109


klemmari1 (approved):
Tested & works!

PR #87, opened 2021-06-09
HL-53 | Company API 

quyenlq:
## Description :sparkles:
Added company API 
When the YTJ request failed to get data, the API will return latest company data saved in the database as a fallback data 

Most of the code is copied from Kesaseteli with small modification to fit Benefit needs


backend/benefit/common/utils.py
```diff
@@ -0,0 +1,8 @@
+def update_object(obj, data):
```
haavikko, line 1: good utility

backend/benefit/companies/api/v1/views.py
```diff
@@ -0,0 +1,62 @@
...
+        company = Company(**DUMMY_COMPANY_DATA)
+        company_data = CompanySerializer(company).data
+
+        return Response(company_data)
+
+    @transaction.atomic
+    def get(self, request: Request, format: str = None) -> Response:
+        if settings.MOCK_FLAG:
+            return self.get_mock(request, format)
+        business_id = request.META.get("HTTP_BUSINESS_ID") or "0877830-0"
```
haavikko, line 42: Is 0877830-0 there by accident? (I. Haanpää Oy)

backend/benefit/common/tests/conftest.py
```diff
@@ -1,10 +1,27 @@
...
 
 
 @pytest.fixture(autouse=True)
 def setup_test_environment(settings):
     factory.random.reseed_random("777")
     with freeze_time("2021-01-04"):
         yield
+
+
+@pytest.fixture
```
haavikko, line 16: Good to have these in the common conftest, I have identical in applications/tests which can be removed once these are merged

backend/benefit/companies/services.py
```diff
@@ -0,0 +1,28 @@
+from common.utils import update_object
+from companies.models import Company
+
+from shared.ytj.ytj_client import YTJClient
```
haavikko, line 4: As we aren't sure yet whether we the YTJClient functionality is sufficient for our purposes, after merging this PR we should investigate also the other options (palveluväylä, other YTJ apis).

haavikko (approved):
Looks good to me!

backend/benefit/companies/api/v1/views.py
klemmari1: This is something I used for testing purposes for YTJ.

backend/benefit/companies/services.py
quyenlq: Sure, this is a temporary solution until we have a decision of which API will be used

backend/benefit/companies/api/v1/views.py
quyenlq: Yeah I think we can remove it in our Benefit, I don't think we'll need this in our client

PR #88, opened 2021-06-09
YJDH-124: Fix browser tests by adding url as referrer to http headers

nenonja:
## Description :sparkles:
- Fix browser tests by adding url as referrer to http headers. Using this idea: https://stackoverflow.com/questions/61116376/set-referrer-for-firefox-on-test-cafe
- Make company page test simpler
- Commented out acceptance test step for benefit-applicant as there is none browser test.

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-109


melniiv (approved):
Great!

PR #92, opened 2021-06-10
YJDH-109 | Fix kesaseteli nginx setup

klemmari1:
## Description :sparkles:

Fix the nginx setup by adding an entrypoint file where the cert is created if it is missing and starting the nginx server.


nenonja (approved):
worked for me

PR #93, opened 2021-06-10
HL-16: Add wizard actions to each step

melniiv:
## Description :sparkles:
Missing actions are added so that user can go through all wizard and back.

### DIFF
--- a/frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
+++ b/frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
@@ -50,10 +50,10 @@ const PageContent: React.FC = () => {
       </StyledPageHeader>
       {currentStep === 1 && <ApplicationFormStep1 actions={actions} />}
       {currentStep === 2 && <ApplicationFormStep2 actions={actions} />}
-      {currentStep === 3 && <ApplicationFormStep3 />}
-      {currentStep === 4 && <ApplicationFormStep4 />}
-      {currentStep === 5 && <ApplicationFormStep5 />}
-      {currentStep === 6 && <ApplicationFormStep6 />}
+      {currentStep === 3 && <ApplicationFormStep3 actions={actions} />}
+      {currentStep === 4 && <ApplicationFormStep4 actions={actions} />}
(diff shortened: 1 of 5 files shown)


frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
```diff
@@ -50,10 +50,10 @@ const PageContent: React.FC = () => {
...
       {currentStep === 1 && <ApplicationFormStep1 actions={actions} />}
       {currentStep === 2 && <ApplicationFormStep2 actions={actions} />}
-      {currentStep === 3 && <ApplicationFormStep3 />}
-      {currentStep === 4 && <ApplicationFormStep4 />}
-      {currentStep === 5 && <ApplicationFormStep5 />}
-      {currentStep === 6 && <ApplicationFormStep6 />}
+      {currentStep === 3 && <ApplicationFormStep3 actions={actions} />}
+      {currentStep === 4 && <ApplicationFormStep4 actions={actions} />}
+      {currentStep === 5 && <ApplicationFormStep5 actions={actions} />}
+      {currentStep === 6 && <ApplicationFormStep6 actions={actions} />}
```
Nizar-Rahme, line 56: On a side note, I'm wondering if using the step number as a component name, as opposed to a descriptive one, would make it harder in the long run to maintain.

For example, if we decided to change the order of the steps, renaming these components would be quite tedious.

Nizar-Rahme (approved):
LGTM!

frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
melniiv: yes we might rename those. the initial idea was to make a dynamic loading, but Jaakko had a valid comment with nextjs docs refference that literal string with vars cannot be used for dynamic import. so i refactored it back, but didnt change the names. On the other hand, i dont think that the order of the steps will actually change, but the contents/fields might jump from one to another, which happened in PT project for instance. But yea, if you feel like renaming, we can do that in other PR

PR #96, opened 2021-06-10
Fix hot reload on docker-compose + clean up docker scripts

nenonja:
## Description :sparkles:
Hot reloading did not work for frontend services in docker-compose. This is now fixed-


klemmari1 (approved):
LGTM!

PR #100, opened 2021-06-11
HL-56: Application state base, grants state, stepper state

melniiv:


frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/extended.ts
```diff
@@ -31,10 +27,11 @@ type FormFields = {
   [DE_MINIMIS_AID_FIELDS.ISSUE_DATE]: string;
 };
 
-const useComponent = (submitForm: () => void): ExtendedComponentProps => {
+const useComponent = (): ExtendedComponentProps => {
   const { t } = useTranslation();
   const translationsBase = 'common:applications.sections.company';
-  const [isSubmitted, setIsSubmitted] = useState<boolean>(false);
+  const { application, setApplication } = React.useContext(ApplicationContext);
```
nenonja, line 33: Not sure if context is necessary if you plan to use `react-query`. You can use `useQuery` for getting `application` and `useMutation` for `setApplication` respectively.

nenonja: However data that is not synced to backend (ie. step number and selected language) could be in the context if you dont wish to use location params.

nenonja (commented):
Some somments, however this is good start to have some state management and can be refactored later if react-query will step in. One thing to note is the state is not persisted so everything is lost after page refresh.

nenonja:
When I tried https://helsinkibenefit-bf-appl-100.test.kuva.hel.ninja/ to create new application and filled the step 1 and moved to step 2 and then back to step.1 it didnt remember the values on step 1 anymore?

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/extended.ts
melniiv: well, the thing is that it will be data from different endpoints, as we have separate company data, then digital signing data, and so on and so on, which might not all be stored in one place. and the state + probably quite a lot of additional info (as you said current step and so on) will be stored in context. So basically context will be the whole state of the app, including application. Thats the current plan. But of course as less duplicates as possible, so i have to see the backend first to tell exactly. Query param we will use only for application id, so on refresh we fetch the data if needed.

melniiv:
yes, this pr is not about step 1. Its a base of the state, next PR will cover the step 1 so that you can actually go from 1 to second step and back and see the actual values.

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/extended.ts
nenonja: Different endpoints is not a problem for react-query, but yes it's probably safer to go this way and then refactor later if needed 👍

nenonja (approved):

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/extended.ts
melniiv: ok, i ll check it during the implementation

PR #102, opened 2021-06-14
Replace hard-coded variable values with HDS CSS custom properties

Nizar-Rahme:
## Description :sparkles:
- Replace hard-coded variable values with HDS CSS custom properties


nenonja (approved):
Looks good to me and application looks the same as before the change

PR #103, opened 2021-06-14
Error validation handling formik applicant

melniiv:
## Description :sparkles:
- added translations for different types of validation errors.
- added validation for deminimis form

## Issues :bug:
hds Date input component does not reset the value with resetForm (but is generic, does not work with setting the value with useState).

### DIFF
--- a/frontend/benefit/applicant/public/locales/en/common.json
+++ b/frontend/benefit/applicant/public/locales/en/common.json
@@ -162,30 +162,27 @@
   "menuToggleAriaLabel": "Menu",
   "form": {
     "validation": {
+      "required": "This field is required",
       "number": {
         "max": "Maximum value is {{max}}",
-        "min": "Value must be at least {{min}}",
-        "required": "This field is required"
+        "min": "Value must be at least {{min}}"
       },
       "string": {
         "email": "This field must be a valid email",
         "max": "This field can be up to {{max}} characters long",
         "min": "This field must be at least {{min}} characters long",
-        "required": "This field is required",
         "url": "This field must be a valid URL",
         "date": "This field must be a valid date",
         "time": "Time must be given in the following format hh:mm",
         "positiveNumber": "This field must be a valid number"
       },
       "date": {
-        "required": "This field is required",
         "mustNotInThePast": "This date must not be in the past",
         "min": "This date must be earliest {{min}}",
         "max": "This date must be latest {{max}}",
         "format": "This date is incorrect"
       },
       "time": {
-        "required": "This field is required",
         "min": "This time must be earliest {{min}}",
         "max": "This time must be latest {{max}}"
       }
--- a/frontend/benefit/applicant/public/locales/en/common.json
+++ b/frontend/benefit/applicant/public/locales/en/common.json
@@ -159,5 +159,36 @@
     "linkText": "hakemiseen vaadittaviin tietoihin ja tarvittaviin liitteisiin",
     "newApplicationBtnText": "Tee uusi hakemus"
   },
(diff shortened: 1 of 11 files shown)


frontend/benefit/applicant/public/locales/en/common.json
```diff
@@ -159,5 +159,36 @@
...
+      },
+      "date": {
+        "required": "This field is required",
+        "mustNotInThePast": "This date must not be in the past",
+        "min": "This date must be earliest {{min}}",
+        "max": "This date must be latest {{max}}",
+        "format": "This date is incorrect"
+      },
+      "time": {
+        "required": "This field is required",
```
Nizar-Rahme, line 188: `required` seems to be common property, should we move it a step higher?

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/DeMinimisAidForm.tsx
```diff
@@ -61,6 +61,7 @@ const DeMinimisAidForm: React.FC = () => {
               name={fields.deMinimisAidIssueDate.name}
               label={fields.deMinimisAidIssueDate.label}
               placeholder={fields.deMinimisAidIssueDate.placeholder}
+              language="fi"
```
Nizar-Rahme, line 64: I prefer if we define an `enum` of the supported languages and use its values instead (e.g: `Lang.FI`).

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/extended.ts
```diff
@@ -83,18 +102,12 @@ const useComponent = (): ExtendedComponentProps => {
...
-    // todo: implement error messages
-    // (getIn(formik.touched, fieldName) || isSubmitted) &&
-    // getIn(formik.errors, fieldName)
-    if (isSubmitted && fieldName) {
-      return fieldName;
-    }
-    return '';
-  };
+  const getErrorMessage = (fieldName: string): string | undefined =>
+    getErrorText(formik.errors, formik.touched, fieldName, t, isSubmitted);
```
Nizar-Rahme, line 106: Cannot the return type `string | undefined`  of `getErrorMessage` be inferred?  Does `getErrorText` return `undefined`?

Nizar-Rahme (changes requested):
Left a couple of comments, overall looks good!

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/DeMinimisAidForm.tsx
melniiv: yea we can. but it should we always the same value, independent of the locale we choose i guess

melniiv: yes

frontend/benefit/applicant/public/locales/en/common.json
melniiv: yes, it was from another projects, as they have already all translations. probably the idea was that the message for different types might be different. in practice not, so i fixed.

frontend/benefit/applicant/src/components/applications/forms/application/deMinimisAid/extended.ts
melniiv: yes, forgot to remove after refactoring, good point.

Nizar-Rahme (commented):
LGTM!

Nizar-Rahme (approved):

PR #105, opened 2021-06-16
Update company API to use query string instead of header

quyenlq:
## Description :sparkles:
Changing Company Get API to get business_id from path instead of header, which is more convenient for client side

Also add the CORS_ORIGIN_ALLOW_ALL = 1 to example .env

### DIFF
--- a/.env.benefit.example
+++ b/.env.benefit.example
@@ -4,6 +4,7 @@ CREATE_SUPERUSER=1
 EMPLOYER_URL=http://localhost:3000
 HANDLER_URL=http://localhost:3100
 NEXT_PUBLIC_BACKEND_URL=http://localhost:8000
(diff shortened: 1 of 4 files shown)


klemmari1:
Not sure if you want to follow the same logic here as kesäseteli but there we fetch the business id from the organization roles.

quyenlq:
> Not sure if you want to follow the same logic here as kesäseteli but there we fetch the business id from the organization roles.
> 
> https://github.com/City-of-Helsinki/yjdh/blob/develop/backend/kesaseteli/companies/api/v1/views.py#L66

Yeah, that'd be something we should follow, but since we don't have the authentication working now, this should be a temporary solution so FE can fetch company data.

haavikko (approved):
Looks good! Better to have the company identifier in the URL

PR #106, opened 2021-06-16
HL-38 | Add Employee model

quyenlq:
## Description :sparkles:
Add Employee Model


backend/benefit/applications/models.py
```diff
@@ -260,3 +281,71 @@ class Meta:
...
+    application = models.OneToOneField(
+        Application,
+        verbose_name=_("application"),
+        related_name="employee",
+        on_delete=models.CASCADE,
+    )
+
+    first_name = models.CharField(max_length=128, verbose_name=_("first name"))
+    last_name = models.CharField(max_length=128, verbose_name=_("last name"))
+    social_security_number = models.CharField(
```
haavikko, line 296: Kesäseteli uses an encrypted field for the SSN, should we do that too?

backend/benefit/applications/models.py
```diff
@@ -260,3 +281,71 @@ class Meta:
...
+        verbose_name=_("working hour"),
+        decimal_places=1,
+        max_digits=4,
+        blank=True,
+        null=True,
+    )
+
+    # TODO: Adding missing file fields or a separate Attachment relationship
+    # power_of_attorney = models.FileField()
+    # collective_bargaining_agreement = models.FieldField()
```
haavikko, line 343: I think that the collective_bargaining_agreement would be a freeform text input (not the actual file itself)

backend/benefit/applications/tests/factories.py
```diff
@@ -71,3 +72,25 @@ def bases(self, created, extracted, **kwargs):
...
+    phone_number = factory.Faker("phone_number")
+    email = factory.Faker("email")
+
+    employee_language = factory.Faker(
+        "random_element", elements=[v[0] for v in APPLICATION_LANGUAGE_CHOICES]
+    )
+    job_title = factory.Faker("job")
+    monthly_pay = factory.Faker("random_int", max=5000)
+    monthly_pay = factory.Faker("random_int", max=5000)
+    monthly_pay = factory.Faker("random_int", max=5000)
```
haavikko, line 92: typo?

haavikko (changes requested):
Some small things

backend/benefit/applications/tests/factories.py
quyenlq: Forgot to change the name, fixed now 😂

backend/benefit/applications/models.py
quyenlq: Good point, this should be encrypted, fixed

backend/benefit/applications/models.py
quyenlq: Ah, right, yeah I thought it was part of the attachments in the third step

haavikko (approved):
Looks good! Would be useful to add Django admin support for Employee before committing though?

klemmari1 (approved):
Looks good to me!

PR #107, opened 2021-06-17
YJDH-133 | KS-Backend: Refactor Company API logic

klemmari1:
## Description :sparkles:

- Create a company with just the name and the business id (fetched from suomi.fi) in case YTJ API does not find a company with the given business id.
- Set Company fields as optional: company_form, industry, street_address, postcode, city
- Add fields for manual user input: manual_company_form, manual_industry, manual_street_address, manual_postcode, manual_city

## Issues :bug:

[YJDH-133](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-133)


backend/kesaseteli/companies/api/v1/views.py
```diff
@@ -66,11 +66,12 @@ def get(self, request: Request, format: str = None) -> Response:
...
         try:
-            company = get_or_create_company_with_business_id(business_id)
-        except HTTPError:
-            return self.ytj_api_error
+            company = get_or_create_company_from_ytj_api(business_id)
         except ValueError:
             return self.ytj_response_error
+        except HTTPError:
+            name = organization_roles.get("name")
+            company = get_or_create_company_with_name_and_business_id(name, business_id)
```
haavikko, line 74: In case it is an transient error (YTJ API under too heavy load or something else), will the application logic re-try at some point (such as, when user performs some action in future?)

haavikko (approved):
Looks good to me! Just some ideas in comments but feel free to merge

backend/kesaseteli/companies/models.py
```diff
@@ -7,11 +7,33 @@
 class Company(UUIDModel):
```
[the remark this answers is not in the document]
quyenlq: Hmn, actually you can refactor Company to use AbstractCompany without any issue, I don't think there is any field in AbstractCompany which is not used in KS custom company.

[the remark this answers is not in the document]
klemmari1: Yea now it was clarified for us that some of those Company fields are not required aka `blank=True` which is probably not desirable for benefit.

backend/kesaseteli/companies/services.py
```diff
@@ -25,3 +25,20 @@ def get_or_create_company_with_business_id(business_id: str) -> Company:
     company_data = ytj_client.get_company_data_from_ytj_data(ytj_data)
```
quyenlq, line 25: Not sure if you noticed but there is a YTJClient in shared component, I think you can just use it and get rid of the YTJClient in the KS `companies` app. They are basically the same

backend/kesaseteli/companies/models.py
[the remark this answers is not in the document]
quyenlq: Ahh, right, if so I think we can just use our own Company for now, as both of Company models have diverged atm 👍

backend/kesaseteli/companies/services.py
klemmari1: Yep, I'll create a separate PR for that because I have to set up the shared code in the Dockerfile etc.

backend/kesaseteli/companies/api/v1/views.py
klemmari1: At the moment it is not retried. If the information is not found from YTJ / an error occurs, the user has to fill in the information themselves. We still have to discuss what happens in the case of an error and if we want to have some kind of special handling for that.

quyenlq (approved):
LGTM! 👌

PR #109, opened 2021-06-17
YJDH-98 | KS-Backend: Azure AD

klemmari1:
## Description :sparkles:

- Start using the shared package
- Handle invalid OIDC auth
- Add Azure ADFS library to Django

The authentication functionality is pretty much handled by the Azure ADFS library but fetching the user groups from Microsoft Graph API had to be implemented "manually". The users in HKI city might have so many groups that they wont all fit in the access token claims (because of the HTTP header size limitations).

## Issues :bug:

[YJDH-98](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-98)


.github/workflows/ks-py-coding-style.yml
```diff
@@ -39,7 +39,7 @@ jobs:
             ${{ runner.os }}-
 
       - name: Install dependencies
-        run: pip install -r backend/kesaseteli/requirements.txt -r backend/kesaseteli/requirements-dev.txt
+        run: cd backend/kesaseteli && pip install -r requirements.txt -r requirements-dev.txt
```
haavikko, line 42: Is this change about style or does it fix something?

backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -196,6 +203,26 @@
...
 
+# Azure ADFS
+LOGIN_URL = "django_auth_adfs:login"
+
+ADFS_CLIENT_ID = env.str("ADFS_CLIENT_ID")
+ADFS_CLIENT_SECRET = env.str("ADFS_CLIENT_SECRET")
+ADFS_TENANT_ID = env.str("ADFS_TENANT_ID")
+
+AUTH_ADFS = {
+    "AUDIENCE": ADFS_CLIENT_ID,
```
haavikko, line 214: for us, is AUDIENCE the same as ADFS_CLIENT_ID in all environments?

backend/shared/shared/oauth2/auth.py
```diff
@@ -0,0 +1,129 @@
...
+
+        response = requests.post(
+            url,
+            json=data,
+            headers=headers,
+        )
+        response.raise_for_status()
+
+        # This will be a list of UUIDs. We will use them as group names in Django.
+        groups = response.json()["value"]
```
haavikko, line 33: Could the group name have some kind of indication (prefix or something) that the group was created based on this query? "adfs-{UUID}"?

Also, if the values are supposed to be UUIDS, should it be validated, for example try to convert them to UUID?

backend/shared/shared/oauth2/auth.py
```diff
@@ -0,0 +1,129 @@
...
+        django_groups = [group.name for group in user.groups.all()]
+
+        if sorted(groups) != sorted(django_groups):
+            existing_groups = list(Group.objects.filter(name__in=groups).iterator())
+            existing_group_names = frozenset(group.name for group in existing_groups)
+            new_groups = [
+                Group.objects.get_or_create(name=name)[0]
+                for name in groups
+                if name not in existing_group_names
+            ]
```
haavikko, line 52: The code doesn't ever seem to remove groups, should it remove groups if groups have been removed from the user in AD?

.env.kesaseteli.example
```diff
@@ -18,5 +18,11 @@ EAUTHORIZATIONS_CLIENT_ID=
 EAUTHORIZATIONS_CLIENT_SECRET=
 EAUTHORIZATIONS_API_OAUTH_SECRET=
 
+ADFS_CLIENT_ID=
+ADFS_CLIENT_SECRET=
+ADFS_TENANT_ID=
+ADFS_LOGIN_REDIRECT_URL=
+ADFS_LOGIN_REDIRECT_URL_FAILURE=
```
haavikko, line 25: Where would these two redirect urls point in actual use?

backend/shared/shared/oauth2/views.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class HelsinkiOAuth2CallbackView(OAuth2CallbackView):
+    def get(self, request):
+        response = super().get(request)
+
+        if response.status_code == 302:
+            return redirect(settings.ADFS_LOGIN_REDIRECT_URL)
+        else:
+            return redirect(settings.ADFS_LOGIN_REDIRECT_URL_FAILURE)
```
haavikko, line 13: Should we log some info about the response if it's a failure, like response code? OAuth2CallbackView in django-auth-adfs uses 400, 401, 403 for separate error cases

haavikko (commented):
Just a few comments, the main one is about update_user_groups_from_graph_api. If I misunderstood the purpose, please add a comment to code

.github/workflows/ks-py-coding-style.yml
quyenlq: I can answer it, `requirement.txt` has a relative path to a `shared` folder, and since `pip` has a bug that will translate this relative path to an incorrect path (it calculates the path from the folder where we run the command, not the actual location of the `requirement.txt`) This is a workaround way to solve that issue

klemmari1: It helps when installing the shared package so that it will be found correctly from the relative path: `-e ../shared`

backend/kesaseteli/kesaseteli/settings.py
klemmari1: Yes. More info in https://django-auth-adfs.readthedocs.io/en/latest/azure_ad_config_guide.html#step-2-configuring-settings-py

backend/shared/shared/oauth2/auth.py
klemmari1: Good point!

I'll add the prefix there. I have to make sure that they are all UUIDs and I could implement the UUID validation as well.

backend/shared/shared/oauth2/auth.py
klemmari1: Yep, I copied this from the library but seems it was not thought out there as well :grin: I'll add group removes as well.

.env.kesaseteli.example
klemmari1: They should point to the handler UI success / error views. One thing I was thinking that should we create a separate django project for the handler as well, similarly as it is done in the frontend (at least for kesäseteli) but not sure if needed.

backend/shared/shared/oauth2/views.py
klemmari1: Yep, I'll add logging.

quyenlq (commented):
Maybe a README with some

backend/shared/shared/oauth2/auth.py
```diff
@@ -0,0 +1,129 @@
...
+            "oauth2/", "oauth2/v2.0/"
+        )
+        LOGGER.debug("Getting access token at: %s", provider_config.token_endpoint)
+        response = provider_config.session.post(
+            token_endpoint, data, timeout=settings.TIMEOUT
+        )
+        # 200 = valid token received
+        # 400 = 'something' is wrong in our request
+        if response.status_code == 400:
+            if response.json().get("error_description", "").startswith("AADSTS50076"):
```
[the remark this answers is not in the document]
klemmari1: This is basically copied from the library (from `django_auth_adfs.backend.AdfsBaseBackend.exchange_auth_code`), I just changed the path to use v2.0 and modified the data. Have to investigate myself where that comes from and I could add documentation about where the method comes from as well.

.github/workflows/ks-py-coding-style.yml
haavikko: ok!

backend/kesaseteli/kesaseteli/settings.py
haavikko: ok

.env.kesaseteli.example
haavikko: ok! These URLs would not be used if suomi.fi authentication is used.

It's a good question, whether the handler and applicant code should be split... in Helsinki-benefit at least, I hope the same API (with appropriate permission checks) can be used dor both

backend/shared/shared/oauth2/auth.py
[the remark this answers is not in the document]
klemmari1: Added docs to the function. Also added README: backend/shared/shared/azure_adfs/README.md

backend/shared/shared/oauth2/auth.py
klemmari1: Fixed

backend/shared/shared/oauth2/auth.py
klemmari1: Fixed

backend/shared/shared/oauth2/views.py
klemmari1: Fixed

quyenlq (approved):
Nicely done! 👌 💯

PR #115, opened 2021-06-21
HL-54 | OIDC shared component, added oidc & adfs authentication to Benefit API

quyenlq:
## Description :sparkles:
- Added ODIC to shared component
- Added ADFS authentication to Helsinki-benefit
- Minor changes to workflows, .env template


backend/shared/setup.py
```diff
@@ -21,9 +21,7 @@
     url="https://github.com/City-of-Helsinki/yjdh",
     author="City of Helsinki",
     author_email="dev@hel.fi",
-    install_requires=[
-        "Django",
-    ],
+    install_requires=["Django", "mozilla_django_oidc"],
```
klemmari1, line 24: Do we need to add `django_auth_adfs` and `django-extensions` here as well?

backend/benefit/helsinkibenefit/urls.py
```diff
@@ -11,7 +11,12 @@
 urlpatterns = [
     path("admin/", admin.site.urls),
     path("v1/", include((router.urls, "v1"), namespace="v1")),
-    path("v1/company/<str:business_id>", GetCompanyView.as_view()),
+    path(
+        "v1/company/<str:business_id>", GetCompanyView.as_view()
+    ),  # FIXME: Remove this later
+    path("v1/company/", GetCompanyView.as_view()),
+    path("oidc/", include("oidc.urls")),
```
klemmari1, line 18: Should this be `include("shared.oidc.urls")`?

quyenlq: I think both would work, as "oidc" is also declared in INSTALLED_APPS

haavikko:
I'd like to test the authentication flow locally, could you please provide instructions, how to set it up with suomi.fi test service?

haavikko:
Test coverage on some files is low, is it because the tests use mocks? Is it feasible/productive to try and increase the test coverage? Github doesn't let me see the coverage report ("Github API rate limit error") right now so I can't see the details.

klemmari1:
It gives you the "Github API rate limit error" if you are not logged in to codecov. It is strange that it shows that low since in the main branch that folder has 94.71% coverage before moving to shared.

https://codecov.io/gh/City-of-Helsinki/yjdh/tree/main/backend/kesaseteli

backend/shared/setup.py
quyenlq: What is the use of `django-extensions` there? I guess it's to enable the local https:// server? Is there any other purpose? If so I think we can just keep it in project requirement-dev.
I fixed `django_auth_adfs` btw

klemmari1: It would be for the `HourlyJob`: https://github.com/City-of-Helsinki/yjdh/pull/115/files#diff-f2014c778aaa7c004551dfb12563b6d347deffa2344c1ba22de99f9e4abddaadR2

quyenlq:
> It gives you the "Github API rate limit error" if you are not logged in to codecov. It is strange that it shows that low since in the develop branch that folder has 94.5% coverage before moving to shared.
> 
> https://codecov.io/gh/City-of-Helsinki/yjdh/tree/develop/backend/kesaseteli

Can it be that there are other tests in another place of KS?

klemmari1:
> Can it be that there are other tests in another place of KS?

I don't think so. Now that I checked, seems that it's not running the tests at all:
https://github.com/City-of-Helsinki/yjdh/runs/2893474610?check_suite_focus=true#step:6:44

klemmari1:
Ah, the `oidc` folder needs to be moved to `/backend/shared/shared/`

klemmari1:
I wonder if [this](https://github.com/City-of-Helsinki/yjdh/blob/hl-54/oidc-to-shared-component/backend/shared/shared/oidc/migrations/0002_eauthorizationprofile.py#L74) works or does it need to be:
```
to="shared.oidc.oidcprofile",
```

backend/shared/shared/oidc/migrations/0002_eauthorizationprofile.py
```diff
@@ -0,0 +1,80 @@
...
+                    models.DateTimeField(
+                        blank=True, null=True, verbose_name="refresh token expires"
+                    ),
+                ),
+                (
+                    "oidc_profile",
+                    models.OneToOneField(
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name="eauthorization_profile",
+                        to="oidc.oidcprofile",
```
klemmari1, line 74: I wonder if this works or does it need to be:
```
to="shared.oidc.oidcprofile",
```

quyenlq: That's an interesting question. I tested and this still works. The table name follows the format of <app_name>.<model_name> so this should be correct

klemmari1: Alrighty, just making sure. If it works then its :ok_hand:

klemmari1 (approved):
Great stuff! Lookin' good to me!

PR #120, opened 2021-06-24
YJDH-117 | Create new application

klemmari1:
## Description :sparkles:

Modify the logic of creating companies from eauth profile:
- Save the eauth profile as an one-to-one relation to the company.
- Separate function `get_or_create_company_from_eauth_profile` that handles the logic of the company creation.
- Handle mock calls.
- Adjust tests.
- Add mock authentication to the REST API auth 

Create new application:
- Allow creation of new applications from the application API. Create a new empty application by POSTing `{}` to the API.
- Handle mock calls to the creation endpoint.
- Adjust and create tests.

## Issues :bug:

[YJDH-117](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-117)


backend/kesaseteli/oidc/auth.py
```diff
@@ -120,6 +120,22 @@ def authenticate(self, request):
             return None
 
         if getattr(settings, "MOCK_FLAG", None):
+            from oidc.models import EAuthorizationProfile, OIDCProfile
```
klemmari1, line 123: I'll add these to the shared package when transferring to that (ASAP).

backend/kesaseteli/companies/services.py
```diff
@@ -40,5 +53,45 @@ def get_or_create_company_with_name_and_business_id(
...
+            company = Company.objects.filter(business_id=business_id).first()
+
+            if not company:
+                try:
+                    company = get_or_create_company_from_ytj_api(
+                        business_id, eauth_profile
+                    )
+                except ValueError:
+                    raise NotFound(detail="Could not handle the response from YTJ API")
+                except HTTPError:
```
haavikko, line 88: Maybe log a warning here?

backend/kesaseteli/applications/api/v1/serializers.py
```diff
@@ -50,3 +52,13 @@ def update(self, instance, validated_data):
             update_summer_vouchers_using_api_data(summer_vouchers_data, instance)
 
         return super().update(instance, validated_data)
+
+    @transaction.atomic
+    def create(self, validated_data):
+        user = self.context["request"].user
+        company = get_or_create_company_from_eauth_profile(
+            user.oidc_profile.eauthorization_profile
```
haavikko, line 60: I guess user.oidc_profile would always be set (required by db schema?) No other kinds of users?

backend/kesaseteli/companies/services.py
```diff
@@ -40,5 +53,45 @@ def get_or_create_company_with_name_and_business_id(
...
+            except HTTPError:
+                raise NotFound(
+                    detail="Unable to fetch organization roles from eauthorizations API"
+                )
+
+            business_id = organization_roles.get("identifier")
+
+            company = Company.objects.filter(business_id=business_id).first()
+
+            if not company:
```
haavikko, line 81: Just an idea - Would it make sense to add a check for missing company.ytj_json and retry to retrieve the data again, if it failed previously? (don't know - if it adds a lot of delay then it would certainly not make sense)

backend/shared/shared/oidc/auth.py
```diff
@@ -65,7 +66,10 @@ def authenticate(self, request, **kwargs):
         }
 
         # Get the token
-        token_info = self.get_token(token_payload)
+        try:
+            token_info = self.get_token(token_payload)
+        except HTTPError:
```
haavikko, line 71: When would this happen? Is it worth logging a warning here?

haavikko (approved):
Added a few questions/comments, address them as you see necessary and go ahead :-)

backend/kesaseteli/applications/api/v1/serializers.py
klemmari1: It would always be set because it is required by the `EAuthRestAuthentication` here: https://github.com/City-of-Helsinki/yjdh/blob/develop/backend/shared/shared/oidc/auth.py#L123. There should not be any other kinds of users at least for now.

backend/shared/shared/oidc/auth.py
klemmari1: This happens if an user tries to log in with the Azure ADFS login method and Django goes through all the authentication backends. This auth backend is before the ADFS backend so it would try this backend first and raise an error here. I don't think it needs a warning necessarily.

backend/kesaseteli/companies/services.py
klemmari1: Hmm, that would certainly make sense! The delay is not too bad so I'll add it.

backend/kesaseteli/companies/services.py
klemmari1: Yep, I'll add :+1:

PR #124, opened 2021-06-28
YDJH-125 | Use shared YTJ client

klemmari1:
## Description :sparkles:

Replace the usage of YTJ client with the shared YTJ client.

## Issues :bug:

[YJDH-125](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-125)


haavikko (approved):
Looks great! Negative productivity :-)

PR #130, opened 2021-07-01
YJDH-149 | Audit logging

klemmari1:
## Description :sparkles:

Implements a shared audit logger for the YJDH project.

Loosely based on: City-of-Helsinki/apartment-application-service#68
Format based on: [Helsinki profile audit log format](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/416972828/Helsinki+profile+audit+logging#Profile-audit-log---CRUD-events---JSON-content-and-format.)

This audit logger allows you to save log messages from Django Rest Famework CRUD events. The serializer works by defining a DRF model viewset that inherits `AuditLoggingModelViewSet`.

- Adds audit logging to the applicaiton viewset CRUD events.
- Add tests.
- Refactor testing tools to common `shared.utils.testing` package.

## Issues :bug:

[YJDH-149](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-149)

### DIFF
--- a/backend/kesaseteli/kesaseteli/settings.py
+++ b/backend/kesaseteli/kesaseteli/settings.py
@@ -61,6 +61,8 @@ env = environ.Env(
     AZURE_ACCOUNT_NAME=(str, ""),
     AZURE_ACCOUNT_KEY=(str, ""),
     AZURE_CONTAINER=(str, ""),
+    AUDIT_LOG_FILENAME=(str, ""),
+    AUDIT_LOG_ORIGIN=(str, ""),
 )
 if os.path.exists(env_file):
     env.read_env(env_file)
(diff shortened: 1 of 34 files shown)


backend/shared/shared/audit_log/viewsets.py
```diff
@@ -0,0 +1,104 @@
...
+from django.db.models import Model
+from django.db.models.base import ModelBase
+from rest_framework.viewsets import ModelViewSet
+from shared.audit_log import audit_logging
+from shared.audit_log.enums import Operation, Status
+
+User = get_user_model()
+
+
+class AuditLoggingModelViewSet(ModelViewSet):
```
haavikko, line 15: Very nice to have this in the shared code :-)
Requirements for audit logging in Helsinki benefit are still unclear but this should be very useful :-)

backend/shared/shared/audit_log/viewsets.py
```diff
@@ -0,0 +1,104 @@
...
+        target = None
+        lookup_value = self.kwargs.get(self.lookup_field, None)
+        if lookup_value is not None:
+            target = self.queryset.model.objects.filter(
+                **{self.lookup_field: lookup_value}
+            ).first()
+        return target or self.queryset.model
+
+    def _get_ip_address(self) -> str:
+        x_forwarded_for = self.request.META.get("HTTP_X_FORWARDED_FOR")
```
haavikko, line 99: Which header would the IBM staging/prod environment set?

haavikko: If IP address is not available and DEBUG=False, should that result in an exception?

backend/shared/shared/audit_log/viewsets.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def perform_update(self, serializer):
+        target_status_before = None
+        target_status_after = None
+
+        if hasattr(serializer.instance, "status"):
+            target_status_before = serializer.instance.status
+        super().perform_update(serializer)
+        if hasattr(serializer.instance, "status"):
+            target_status_after = serializer.instance.status
```
haavikko, line 61: Suggestion - implement if you feel like it: Only log status if target_status_before != target_status_after

backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -155,11 +157,59 @@
 CSRF_COOKIE_DOMAIN = env.str("CSRF_COOKIE_DOMAIN")
 CSRF_TRUSTED_ORIGINS = env.list("CSRF_TRUSTED_ORIGINS")
 
+# Audit logging
+AUDIT_LOG_ORIGIN = env.str("AUDIT_LOG_ORIGIN")
+AUDIT_LOG_FILENAME = env.str("AUDIT_LOG_FILENAME")
```
haavikko, line 162: If running in a kubernetes pod, we can't trust that any local file is really persisted -> need to develop some solution (with IBM?) for storing the audit log. Also, if running multiple pods in prod, need to merge the logs? But that's outside of the scope of this ticket I guess?

haavikko (approved):
Added a few comments, address them as you see suitable&go ahead with merge :-) Looks good!

backend/shared/shared/audit_log/viewsets.py
klemmari1: I don't know if IP address is such a vital information that it would raise an error if it's missing since we have the user information. I don't think the logging should raise errors anyways.

The header that IBM sets, we have to figure out, probably after the holidays.

backend/shared/shared/audit_log/viewsets.py
klemmari1: Sounds good, I'll add!

backend/kesaseteli/kesaseteli/settings.py
klemmari1: This is how helsinki profile has implemented it at the moment so let's go with this for now.

The audit log storing is still an open issue and might well change, we will know more next week / after the holidays about storing the logs and how we transfer them to elastic cloud. One option is that we will store the logs in database which would then affect this solution as well.

PR #131, opened 2021-07-01
YJDH-133 | Remove manual fields from company model

klemmari1:
## Description :sparkles:

Remove the manually fillable fields from company since these won't be used. If the information is not found from YTJ API, these fields will be left empty.

## Issues :bug:

[YJDH-133](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-133)


haavikko (approved):
Great to see negative productivity again :-)

PR #132, opened 2021-07-02
HL-67: add pay subsidy fields and validation rules for application page 2

haavikko:
## Description :sparkles:

* Add missing fields and validation rules
* As per the dicussion with handlers, leave the previously implemented application_logic in place for future extension
 
## Issues :bug:
HL-67


backend/benefit/applications/fixtures/application_basis.json
```diff
@@ -1,52 +1,2 @@
 [
-  {
```
klemmari1, line 2: Was this left as an empty list intentionally or can this file be deleted?

haavikko: hmm actually it can be deleted, good point. Fixed.

klemmari1 (approved):
Looks good to me!

PR #133, opened 2021-07-02
KS-Backend: Refactor tests to use common factories and fixtures

klemmari1:
## Description :sparkles:

Move all factories and fixtures to a common package that can be used in tests.


haavikko (approved):
Looks good to me!

PR #136, opened 2021-07-02
YJDH-128 | Refactor application API read & create

klemmari1:
## Description :sparkles:

Allow users to fetch only the applications that are linked to the current active company.

Allow the creation of only one DRAFT application.

Set the application default status to DRAFT.

## Issues :bug:

[YJDH-128](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-128)


backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -12,5 +14,23 @@ class ApplicationViewSet(AuditLoggingModelViewSet):
     )
```
haavikko, line 14: Does the queryset attribute need to remain defined, now that get_queryset() is defined? Can the select_related / prefetch_related be moved to get_queryset()?

haavikko (approved):
I don't quite know how the queryset / get_queryset thing plays out, please check it out&change if there is a real need to

backend/kesaseteli/applications/api/v1/views.py
klemmari1: Yea I'll move the prefetching to `get_queryset` to make it more clear.

klemmari1:
There is one browser test failing which might be because the browser tests are not yet adjusted to the new logic. Let's leave this open until that is sorted out.

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -1,16 +1,39 @@
...
+        )
+        user_company = self.request.user.oidc_profile.eauthorization_profile.company
+        return queryset.filter(company=user_company, status=ApplicationStatus.DRAFT)
+
+    def create(self, request, *args, **kwargs):
+        """
+        Allow only 1 (DRAFT) application per company.
+        """
+        if self.get_queryset().exists():
+            raise PermissionDenied("Company can have only one draft application")
```
nenonja, line 35: perhaps nitpick but permission denied usually means that user has no right to do an operation (403). But this is just illegal operation (400 or 422 or something).

klemmari1: Good point! I'll change this to `ValidationError` (400).

PR #156, opened 2021-07-23
HL-101 | Format Phone number field

quyenlq:


haavikko (approved):
Looks good to me! If it's simple to add, would be nice to have a testcase for invalid (foreign) phone number. But feel free to commit as-is

quyenlq:
> Looks good to me! If it's simple to add, would be nice to have a testcase for invalid (foreign) phone number. But feel free to commit as-is

I added some invalid inputs to the `test_application_post_invalid_employee_data` and `test_application_post_invalid_data`. I think that's enough to cover the case

PR #161, opened 2021-07-30
HL-85 | Add max file size upload validation, enable private blob storage

quyenlq:

### DIFF
--- a/backend/benefit/applications/api/v1/serializers.py
+++ b/backend/benefit/applications/api/v1/serializers.py
@@ -24,6 +24,7 @@ from dateutil.relativedelta import relativedelta
 from django.db import transaction
 from django.utils.translation import gettext_lazy as _
 from drf_spectacular.utils import extend_schema_field
+from helsinkibenefit.settings import MAX_UPLOAD_SIZE
 from rest_framework import serializers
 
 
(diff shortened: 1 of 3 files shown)


backend/benefit/helsinkibenefit/settings.py
```diff
@@ -274,6 +275,9 @@
 AZURE_ACCOUNT_NAME = env("AZURE_ACCOUNT_NAME")
 AZURE_ACCOUNT_KEY = env("AZURE_ACCOUNT_KEY")
 AZURE_CONTAINER = env("AZURE_CONTAINER")
+AZURE_URL_EXPIRATION_SECS = env("AZURE_URL_EXPIRATION_SECS")  # default 900s
+
+MAX_UPLOAD_SIZE = 13107200  # 10MB
```
haavikko, line 280: 13107200 / (1024*1024) == 12.5?

backend/benefit/applications/api/v1/serializers.py
```diff
@@ -98,6 +99,13 @@ class Meta:
         ]
         read_only_fields = ["created_at"]
 
+    def validate(self, data):
+        if data["attachment_file"].size > MAX_UPLOAD_SIZE:
+            raise serializers.ValidationError(
+                f"Upload file size cannot be greater than {MAX_UPLOAD_SIZE} bytes"
```
haavikko, line 105: Should error messages like these be localized? I see that I've myself been inconsistent with this

haavikko (commented):
Looks good! Added a couple of comments. Also, if it's easy to write a testcase for this, a test would be nice?

backend/benefit/applications/api/v1/serializers.py
haavikko: A separate jira issue was created for this yesterday

backend/benefit/helsinkibenefit/settings.py
quyenlq: Didn't remember why I came up with this number 😅 Yes it should be 10485760

haavikko (approved):
Looks good to me :-)

quyenlq:
> Looks good! Added a couple of comments. Also, if it's easy to write a testcase for this, a test would be nice?

Added a test for this 😋

PR #163, opened 2021-08-04
HL-118: add application step field

haavikko:
## Description :sparkles:

Add a field to the application model and API for the current/latest application step in UI.Useful for e.g.:

conditional validation (like, validate attachment requirements only when saving the attachments step)

When reloading the "edit application" page, show the previously selected application step.


quyenlq (approved):
Looks good to me 👍

PR #165, opened 2021-08-04
HL-109: Fix deminimis aid amount field type

melniiv:
## Description :sparkles:
For consistent styling (to follow the abstract design), the field is replaced with normal text field and will be validated with Yup.


Nizar-Rahme (approved):
LGTM! :shipit:

PR #168, opened 2021-08-04
Fixing dependabot pr builds: use '-' separator instead of '/' in dependabot pr branch names

nenonja:


nenonja:
I will merge this as it cannot be tested before with dependabot

PR #173, opened 2021-08-05
HL-112: Edit application layout changes

melniiv:
## Description :sparkles:
- created date value is shown in the applications drafts list instead of modified at value
- when application is in edit mode, the title of the application page should be Application
- when application is in edit mode, the creation date is shown under title as unique identifier
- when application is in edit mode, the helper text is shown under title to tell that its an edit mode

### DIFF
--- a/frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
+++ b/frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
@@ -14,7 +14,7 @@ import ApplicationFormStep6 from 'benefit/applicant/components/applications/form
 import React from 'react';
 import Container from 'shared/components/container/Container';
 import Stepper from 'shared/components/stepper/Stepper';
-import { formatDate } from 'shared/utils/date.utils';
+import { DATE_FORMATS, formatDate } from 'shared/utils/date.utils';
 
 import { usePageContent } from './usePageContent';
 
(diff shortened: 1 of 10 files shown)


frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
```diff
@@ -12,23 +14,37 @@ import ApplicationFormStep6 from 'benefit/applicant/components/applications/form
...
         <$HeaderItem>
           <Stepper steps={steps} activeStep={currentStep} />
         </$HeaderItem>
       </$PageHeader>
+      {id && application?.createdAt && (
+        <>
+          <$PageSubHeading>
+            {`${t('common:applications.pageHeaders.created')} ${formatDate(
+              new Date(application?.createdAt),
+              'dd.MM.yyyy. mm:ss'
```
Nizar-Rahme, line 40: I noticed that this format is used in many places, I suggest extracting it as a constant.

Nizar-Rahme (approved):
Left one comment to consider, otherwise, it looks good to me

frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
melniiv: yes, it is reusable. i ll fix

PR #174, opened 2021-08-05
HL-117 | Update localized messages

quyenlq:
## Description :sparkles:
- Update all message to be localized using gettext
- Add LOCALE_PATH to settings to store compiled messages later
- Change message structure so it can be formatted with variables after translation

How Django determine translation can be found here: https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#how-django-discovers-language-preference

### DIFF
--- a/backend/benefit/applications/api/v1/serializers.py
+++ b/backend/benefit/applications/api/v1/serializers.py
@@ -22,6 +22,7 @@ from companies.api.v1.serializers import CompanySerializer
 from companies.models import Company
 from dateutil.relativedelta import relativedelta
 from django.db import transaction
+from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
 from drf_spectacular.utils import extend_schema_field
 from helsinkibenefit.settings import MAX_UPLOAD_SIZE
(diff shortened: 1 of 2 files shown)


backend/benefit/applications/api/v1/serializers.py
```diff
@@ -57,7 +57,9 @@ def __call__(self, value, serializer_field):
             ):
                 raise serializers.ValidationError(
                     _(
-                        f"Application state transition not allowed: {application.status} to {value}"
+                        "Application state transition not allowed: {status} to {value}".format(
+                            status=application.status, value=value
+                        )
```
haavikko, line 62: When discussing this ticket I didn't realize that f strings are not supported by gettext... The issue is more complex than I thought 

Question about format(): does it work here? https://docs.djangoproject.com/en/3.2/topics/i18n/translation/ says that format() doesn't work on lazy translation objects and format_lazy should be used instead? Or should we do
"from django.utils.translation import gettext as _", as I think it's ok to translate the string immediately when raising ValidationErrors?

quyenlq: Aha, good catch, I think we should use the format_lazy instead of using gettext. There are someplace where we have to use gettext_lazy so it would end up us using two different gettext functions. I'll update the PR

haavikko (approved):
Looks good to me

PR #184, opened 2021-08-11
HL-87 | Add auto application number

quyenlq:
## Description :sparkles:
- Added a sequence to DB to generate application number. Sequence starts with a random 6 digits number.
- Make application_number readonly in Admin interface and API
- Auto use the value from DB sequence to fill the application number when new Application is created.


backend/benefit/applications/migrations/0013_application_number_auto_field.py
```diff
@@ -0,0 +1,30 @@
...
+            field=models.IntegerField(
+                default=applications.models._get_next_application_number,
+                verbose_name="application number",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="historicalapplication",
+            name="application_number",
+            field=models.IntegerField(
+                default=applications.models._get_next_application_number,
```
quyenlq, line 26: Hmn, now when I look at it, does it mean the historical_application will also use the same sequence? At least from the test it doesn't seem so

quyenlq: Well, I think Historical will never use the sequence, as it'll always use the value from the origin application, which means it never uses the default value from the model (i.e getting next value from the sequence)

backend/benefit/applications/admin.py
```diff
@@ -56,6 +57,7 @@ class ApplicationAdmin(admin.ModelAdmin):
     search_fields = (
         "id",
         "company__id",
+        "application_number",
```
haavikko, line 60: good catch

backend/benefit/applications/migrations/0013_application_number_auto_field.py
haavikko: ok good to check

haavikko (approved):
Looks good to me!

PR #188, opened 2021-08-13
HL-63 & HL-143 | Error messages and validation for Step 1

Nizar-Rahme:
## Description :sparkles:
- Fix styles on the first step to match the designs
- Add validation rules to all the fields 
- Add error messages and adjust their styles
- Scroll to the top when going to the second step


frontend/benefit/applicant/src/constants.ts
```diff
@@ -185,3 +187,9 @@ export enum VALIDATION_MESSAGE_KEYS {
 }
 
 export const dateRegex = /(?:(?:0?[1-9]|1\d|2\d)\.(?:0?[1-9]|1[0-2])|(?:30\.(?!02)(?:0[1-9]|1[0-2]))|(?:31\.(?:0[13578]|1[02])))\.(?:19|20)\d{2}/;
+export const addressRegex = /^([\d (),./A-Za-zÄÅÖäåö-]+)$/;
+export const companyAccRegex = /^FI\d{16}$/;
+export const phoneRegex = /^((\\+[1-9]{1,4}[ \\-]*)|(\\(\d{2,3}\\)[ \\-]*)|(\d{2,4})[ \\-]*)*?\d{3,4}?[ \\-]*\d{3,4}?$/;
```
nenonja, line 192: This doesnt seem to allow country code in the beginning, for example `+358401234567` should be valid (finnish number) but it isn't.
I'm not even sure is it reasonable to try to validate the phonenumber with regex, it's daunting task if we would want to allow country code: https://stackoverflow.com/questions/4338267/validate-phone-number-with-javascript/5059082#5059082

Accepted answer suggests to use specified library like: https://github.com/google/libphonenumber

frontend/benefit/applicant/src/constants.ts
```diff
@@ -185,3 +187,9 @@ export enum VALIDATION_MESSAGE_KEYS {
 }
 
 export const dateRegex = /(?:(?:0?[1-9]|1\d|2\d)\.(?:0?[1-9]|1[0-2])|(?:30\.(?!02)(?:0[1-9]|1[0-2]))|(?:31\.(?:0[13578]|1[02])))\.(?:19|20)\d{2}/;
```
nenonja, line 189: Where are these regex coming from? Perhaps link to source would help. Also you have to communicate to backend coders that they will use the same validation when sending application to backend.

frontend/benefit/applicant/src/constants.ts
```diff
@@ -185,3 +187,9 @@ export enum VALIDATION_MESSAGE_KEYS {
...
+export const addressRegex = /^([\d (),./A-Za-zÄÅÖäåö-]+)$/;
+export const companyAccRegex = /^FI\d{16}$/;
+export const phoneRegex = /^((\\+[1-9]{1,4}[ \\-]*)|(\\(\d{2,3}\\)[ \\-]*)|(\d{2,4})[ \\-]*)*?\d{3,4}?[ \\-]*\d{3,4}?$/;
+export const postalCodeRegex = /^\d{5}$/;
+export const namesRegex = /^[\w',.-][^\d!#$%&()*+/:;<=>?@[\\\]_{|}~¡¿÷ˆ]{2,}$/;
```
nenonja, line 194: This names regex seem to work quite well against these: https://www.w3.org/International/questions/qa-personal-names
Again some kyrydic russian names and japanese names does not work. The special character might crash the saving to database, should be asked about backend developers too what can be allowed. And also from PO too?

nenonja (commented):
regex needs some attention, perhaps we should talk about them in YJDH level because kesaseteli could use same validation. Also it concern also backend validation and product owners / designers might have opinions too? :)

frontend/benefit/applicant/src/constants.ts
nenonja: With 3rd party library there should be same solution for python too. https://github.com/google/libphonenumber has some port-libary for python so it should be used in the backend.

nenonja: One more option is to not to validate the phone number (except being non-empty) and count on the users that they will input it correctly. Valid value could be also `0401234567 during week and 0407654321 on weekends` if the number is for human reader. For automatic messages it is not...

frontend/benefit/applicant/src/components/applications/forms/application/step2/ApplicationFormStep2.tsx
```diff
@@ -46,6 +46,10 @@ const ApplicationFormStep2: React.FC<DynamicFormStepComponentProps> = ({
     subsidyOptions,
   } = useApplicationFormStep2(data);
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
melniiv: there is no route change, if i understood you correctly. just step change, so we should handle the scroll to top ourselves.

frontend/benefit/applicant/src/constants.ts
melniiv: +358 allowed value will be a a part of next PR, we just discussed it in the refinement, as the backend changed to support that just recently

frontend/benefit/applicant/src/constants.ts
melniiv: good point. though i never had problems with saving russian chars with utf-8 encoding

frontend/benefit/applicant/src/constants.ts
melniiv: i think it was my regex of dd/mm/yyyy, i definitely was testing couple of implementations and took most reliable one, but i cannot find the link. Regex is always the matter of improvising and improving, and there is no "correct" way, but i agree that backend and frontend should have the same rules.

frontend/benefit/applicant/src/constants.ts
nenonja: Kesäseteli will use only required as only validation for phonenumber but afaik there is different need for benefit application

frontend/benefit/applicant/src/constants.ts
nenonja: Some comment line to explain it would be good, at least to tell what is the expected date format

nenonja: It seems to be dd.mm.yyyy. Some corner cases like leap day cannot be validated with just regex so it needs some additional validation.

melniiv: you can actually disable direct typing to the hds datepicker (if its supported). then you cannot select the wrong date.

frontend/shared/src/constants.ts
```diff
@@ -1,2 +1,10 @@
 export const MAIN_CONTENT_ID = 'main_content';
 export const INVALID_FIELD_CLASS = 'invalid-field';
+
+// For the following regex constants, see: frontend/shared/src/__tests__/constants.test.ts
+export const ADDRESS_REGEX = /^([\d (),./A-Za-zÄÅÖäåö-]+)$/;
+export const COMPANY_BANK_ACCOUNT_NUMBER = /^FI\d{16}$/;
+export const PHONE_NUMBER_REGEX = /^((\+358[ -]*)+|(\\(\d{2,3}\\)[ -]*)|(\d{2,4})[ -]*)*?\d{3,4}?[ -]*\d{3,4}?$/;
+export const POSTAL_CODE_REGEX = /^\d{5}$/;
+export const NAMES_REGEX = /^[\w',.-][^\d!#$%&()*+/:;<=>?@[\\\]_{|}~¡¿÷ˆ]{2,}$/;
+export const CITY_REGEX = /^[A-Za-zÄÅÖ-\säåö]+$/;
```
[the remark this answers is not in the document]
Nizar-Rahme: Fixed

frontend/shared/src/__tests__/constants.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      it('should not match non-address strings', () => {
+        const nonAddress = '#@';
+
+        expect(nonAddress).not.toMatch(ADDRESS_REGEX);
+      });
+    });
+
+    describe('PHONE_NUMBER_REGEX', () => {
+      it('should match Finnish phone numbers', () => {
+        const phoneNumbers = [
```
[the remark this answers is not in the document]
Nizar-Rahme: I added a max limit to all the fields, not in regex though rather in Yup validation.

frontend/shared/src/__tests__/constants.test.ts
```diff
@@ -0,0 +1,195 @@
...
+  POSTAL_CODE_REGEX,
+} from '../constants';
+
+describe('constants', () => {
+  describe('regex', () => {
+    describe('NAMES_REGEX', () => {
+      it('should match Finnish first names, last names and full names', () => {
+        faker.locale = 'fi';
+
+        const firstName = faker.name.firstName();
```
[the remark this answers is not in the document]
Nizar-Rahme: Replaced them with static values.

frontend/shared/src/constants.ts
```diff
@@ -1,2 +1,10 @@
 export const MAIN_CONTENT_ID = 'main_content';
 export const INVALID_FIELD_CLASS = 'invalid-field';
+
+// For the following regex constants, see: frontend/shared/src/__tests__/constants.test.ts
+export const ADDRESS_REGEX = /^([\d (),./A-Za-zÄÅÖäåö-]+)$/;
+export const COMPANY_BANK_ACCOUNT_NUMBER = /^FI\d{16}$/;
+export const PHONE_NUMBER_REGEX = /^((\+358[ -]*)+|(\\(\d{2,3}\\)[ -]*)|(\d{2,4})[ -]*)*?\d{3,4}?[ -]*\d{3,4}?$/;
+export const POSTAL_CODE_REGEX = /^\d{5}$/;
+export const NAMES_REGEX = /^[\w',.-][^\d!#$%&()*+/:;<=>?@[\\\]_{|}~¡¿÷ˆ]{2,}$/;
```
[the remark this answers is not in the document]
Nizar-Rahme: >This does not seem to allow one or two character length for a name, although there exists at least last names with single character ie. Janne Ö
@nenonja I updated the expression to allow single-character strings.

 >Also there should not be punctuation characters ',.- in the beginning and end of the name, and no multiple
Yes, it's not a perfect regex, but when I attempted to cover these edge cases, I ended up with a very complex expression that is hard to read and wrap our heads around.

frontend/shared/src/constants.ts
```diff
@@ -1,2 +1,10 @@
 export const MAIN_CONTENT_ID = 'main_content';
 export const INVALID_FIELD_CLASS = 'invalid-field';
+
+// For the following regex constants, see: frontend/shared/src/__tests__/constants.test.ts
+export const ADDRESS_REGEX = /^([\d (),./A-Za-zÄÅÖäåö-]+)$/;
+export const COMPANY_BANK_ACCOUNT_NUMBER = /^FI\d{16}$/;
+export const PHONE_NUMBER_REGEX = /^((\+358[ -]*)+|(\\(\d{2,3}\\)[ -]*)|(\d{2,4})[ -]*)*?\d{3,4}?[ -]*\d{3,4}?$/;
+export const POSTAL_CODE_REGEX = /^\d{5}$/;
+export const NAMES_REGEX = /^[\w',.ÄÅÖäåö-][^\d!#$%&()*+/:;<=>?@[\\\]_{|}~¡¿÷ˆ]+$/;
+export const CITY_REGEX = /^[ -äåö]+$/;
```
nenonja, line 10: What happened to this `CITY_REGEX`? It used to be better, now it allows almost any value. Perhaps we should cover some cases in unit tests what should be invalid, like `!"#"%AAA` which now passes

frontend/shared/src/__tests__/constants.test.ts
```diff
@@ -0,0 +1,195 @@
...
+        ];
+
+        citiesFI.forEach((city) => {
+          expect(city).toMatch(CITY_REGEX);
+        });
+
+        citiesSV.forEach((city) => {
+          expect(city).toMatch(CITY_REGEX);
+        });
+      });
```
nenonja, line 139: As suggested in other comment, we could add some invalid values there too.

frontend/shared/src/__tests__/constants.test.ts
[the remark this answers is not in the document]
nenonja: Dont see the change?

nenonja (commented):
Did you push all the commits? There seems to be some changes missing perhaps?
Also CITY_REGEX has changed for the worse :( 
I would suggest to add also non-matching values to unit tests.

frontend/shared/src/constants.ts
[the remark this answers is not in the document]
nenonja: Something got broken here?

Nizar-Rahme:
>Did you push all the commits? There seems to be some changes missing perhaps?
Also CITY_REGEX has changed for the worse :(
I would suggest to add also non-matching values to unit tests.

Oops! Apparently not all the changes went through, maybe that was due to the conflicting file.
Now everything should be fine

frontend/shared/src/constants.ts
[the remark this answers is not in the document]
nenonja: That is understandable 👍

nenonja (approved):
Looks great! Perhaps there could be even more unit tests for non-matching cases because it's so easy to write them but it's fair enough and up to you to add them. I approve 👍  😅

PR #189, opened 2021-08-15
HL-119,120: Application step backend value to UI logic

melniiv:
## Description :sparkles:
- sets the application step when navigating the stepper
- sets the current application step when loading the existing application

### DIFF
--- a/frontend/benefit/applicant/.eslintrc.js
+++ b/frontend/benefit/applicant/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
   extends: ['auto'],
   rules: {
     'no-secrets/no-secrets': ['error', { tolerance: 4.2 }],
+    'sonarjs/cognitive-complexity': ['error', 20],
     'no-void': 'off',
     'unicorn/prefer-module': 'off',
     'jsx-a11y/anchor-is-valid': [
--- a/frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
+++ b/frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
@@ -11,6 +11,7 @@ import {
   Application,
   ApplicationData,
 } from 'benefit/applicant/types/application';
+import { getApplicationStepString } from 'benefit/applicant/utils/common';
 import { getErrorText } from 'benefit/applicant/utils/forms';
 import { FormikProps, useFormik } from 'formik';
 import { TFunction } from 'next-i18next';
@@ -81,6 +82,7 @@ const useApplicationFormStep2 = (
         {
           ...application,
           ...formik.values,
+          applicationStep: getApplicationStepString(step),
         },
         { deep: true }
       );
@@ -121,9 +123,9 @@ const useApplicationFormStep2 = (
   const getErrorMessage = (fieldName: string): string | undefined =>
     getErrorText(formik.errors, formik.touched, fieldName, t, isSubmitted);
 
-  const handleSubmit = (nextStep: number): void => {
+  const handleSubmitNext = (): void => {
     setIsSubmitted(true);
-    setStep(nextStep);
+    setStep(3);
     void formik.validateForm().then((errors) => {
       // todo: Focus the first invalid field
       const invalidFields = Object.keys(errors);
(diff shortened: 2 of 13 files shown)


frontend/benefit/applicant/.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   extends: ['auto'],
   rules: {
     'no-secrets/no-secrets': ['error', { tolerance: 4.2 }],
+    'sonarjs/cognitive-complexity': ['error', 20],
```
nenonja, line 7: Perhaps it would be better to split the overlong method (usePageContent) to smaller methods/hooks for more readable code, instead of increasing the limit?

frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
```diff
@@ -134,10 +136,17 @@ const useApplicationFormStep2 = (
     });
   };
 
-  const handleSubmitNext = (): void => handleSubmit(3);
-
-  const handleSubmitBack = (): void =>
-    setApplicationTempData({ ...applicationTempData, currentStep: 1 });
+  const handleSubmitBack = (): void => {
+    setStep(1);
+    const currentApplicationData: ApplicationData = snakecaseKeys(
```
nenonja, line 141: Why snakecase is needed? Or could we use snakecase also in frontend to prevent these conversions?

frontend/benefit/applicant/src/components/applications/pageContent/usePageContent.ts
```diff
@@ -62,8 +67,20 @@ const usePageContent = (): ExtendedComponentProps => {
       existingApplicationStatus !== 'loading'
     ) {
       setIsLoading(false);
+      const existingApplicationCurrentStep = getApplicationStepFromString(
+        existingApplication?.application_step || DEFAULT_APPLICATION_STEP
+      );
+      if (applicationTempData.currentStep !== existingApplicationCurrentStep) {
```
nenonja, line 73: is if-check necessary? could `setCurrentStep(existingApplicationCurrentStep);` just work always?

frontend/benefit/applicant/src/utils/common.ts
```diff
@@ -16,3 +16,14 @@ export const getLanguageOptions = (
 
   return createOptions(Object.values(SUPPORTED_LANGUAGES));
 };
+
+export const getApplicationStepFromString = (step: string): number => {
+  try {
+    return parseInt(step.split('_')[1], 10);
```
nenonja, line 22: Why radix parameter `10` is needed?

nenonja (commented):
Some comments :)

frontend/benefit/applicant/src/context/ApplicationContext.tsx
```diff
@@ -16,6 +17,7 @@ const ApplicationContext = React.createContext<ApplicationContextType>({
   },
   isLoading: true,
   setApplicationTempData: () => noop,
+  setCurrentStep: () => noop,
```
nenonja, line 20: `setCurrentStep: noop` is correct. 
Imho I would try to avoid lodash package and create the methods myself, I have bad experiences about the size of it and treeshaking not working properly.

nenonja:
Not sure how to test this. I tried to fill the application and proceed to next step but it gives an error:
https://helsinkibenefit-bf-appl-189.test.kuva.hel.ninja

frontend/benefit/applicant/src/utils/common.ts
melniiv: because linter complains if you dont define any

frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
melniiv: for consistency, i don't see a problem in conversion

frontend/benefit/applicant/src/components/applications/pageContent/usePageContent.ts
melniiv: i ll recheck, but i think it creates infinite loop

frontend/benefit/applicant/.eslintrc.js
melniiv: well, i face this problem already second time in the same file. its not really about splitting but about the number of logical operators. it does not matter if you move your code to utilities and have couple of extra || or && and still facing the same issue

frontend/benefit/applicant/src/context/ApplicationContext.tsx
melniiv: fixed

frontend/benefit/applicant/src/utils/common.ts
melniiv: default for decimal si 10, so i think its ok

frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
melniiv: basically, in my opinion, if we use everywhere else in react app the camelCase for props, we should not rely on django app conventions. We could have 10 integrations in different languages with different conventions, and its fine to map it in your application to common form.

Nizar-Rahme (approved):
It looks good to me

PR #196, opened 2021-08-17
YJDH-135 | Backend: Send language to helsinki profile and suomi.fi

klemmari1:
## Description :sparkles:

Fetches the language from GET parameters on the /authenticate call and sends that to helsinki profile and suomi.fi as a url query parameter. The language is saved in Django session during the auth flow.

## Issues :bug:

[YJDH-136](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-136)
[YJDH-137](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-137)

### DIFF
--- a/backend/shared/shared/oidc/views/eauth_views.py
+++ b/backend/shared/shared/oidc/views/eauth_views.py
@@ -79,6 +79,11 @@ class EauthAuthenticationRequestView(View):
             ).replace("http://", "https://"),
             "user": user_id,
         }
+
+        lang = request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+        if lang:
+            params["lang"] = lang
+
         query = urlencode(params)
 
         redirect_url = "{url}?{query}".format(url=auth_url, query=query)
@@ -92,10 +97,22 @@ class EauthAuthenticationCallbackView(View):
     http_method_names = ["get"]
 
     def login_success(self):
-        return HttpResponseRedirect(settings.LOGIN_REDIRECT_URL)
+        url = settings.LOGIN_REDIRECT_URL
+
+        lang = self.request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+        if lang:
+            url = f"{url}/{lang}/"
+
+        return HttpResponseRedirect(url)
 
     def login_failure(self):
-        return HttpResponseRedirect(settings.LOGIN_REDIRECT_URL_FAILURE)
+        url, error_path = settings.LOGIN_REDIRECT_URL_FAILURE.rsplit("/", 1)
+
+        lang = self.request.COOKIES.get(settings.LANGUAGE_COOKIE_NAME)
+        if lang:
+            url = f"{url}/{lang}/{error_path}"
+
+        return HttpResponseRedirect(url)
 
     def get_token_info(self, code):
         """Return token object as a dictionary."""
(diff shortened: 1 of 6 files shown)


backend/shared/shared/oidc/views/eauth_views.py
```diff
@@ -79,6 +79,11 @@ def get(self, request):
             ).replace("http://", "https://"),
             "user": user_id,
         }
+
+        lang = request.session.get("lang")
```
haavikko, line 83: Wondering if it would be useful to also set the `django_language` cookie? (that's the default for `LANGUAGE_COOKIE_NAME` 
My understanding is that Django Rest Framework would then use the cookie value for localized validation errors? (in Helsinki benefit, the frontend may display the validation errors to the end user so we need to localize them, but it depends on your use case)

backend/shared/shared/oidc/views/eauth_views.py
```diff
@@ -92,10 +97,22 @@ class EauthAuthenticationCallbackView(View):
...
+
+        return HttpResponseRedirect(url)
 
     def login_failure(self):
-        return HttpResponseRedirect(settings.LOGIN_REDIRECT_URL_FAILURE)
+        url, error_path = settings.LOGIN_REDIRECT_URL_FAILURE.rsplit("/", 1)
+
+        lang = self.request.session.pop("lang", None)
+        if lang:
+            url = f"{url}/{lang}/{error_path}"
```
haavikko, line 113: I assume this URL format has been agreed with UI devs?

haavikko (approved):
Looks good! Left a comment about django_language cookie

backend/shared/shared/oidc/views/eauth_views.py
klemmari1: This sounds like a good idea! I'll make some changes.

backend/shared/shared/oidc/views/eauth_views.py
klemmari1: Yes

backend/shared/shared/oidc/views/eauth_views.py
```diff
@@ -92,10 +97,22 @@ class EauthAuthenticationCallbackView(View):
     http_method_names = ["get"]
 
     def login_success(self):
-        return HttpResponseRedirect(settings.LOGIN_REDIRECT_URL)
+        url = settings.LOGIN_REDIRECT_URL
+
+        lang = self.request.session.get("django_language")
```
quyenlq, line 102: I believe the default `LANGUAGE_SESSION_KEY` is `_language`, and `LANGUAGE_COOKIE_NAME` is `django_language`. So if you still want to use session, maybe use `_language`  or  `from django.utils.translation import LANGUAGE_SESSION_KEY`

quyenlq (approved):
Looks good, just a minor comment, feel free to merge if disagree :)

backend/shared/shared/oidc/views/eauth_views.py
klemmari1: Good point! I should've looked more into this.

klemmari1: Seems that `LANGUAGE_SESSION_KEY` is deprecated and the cookie should be used.

https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.translation.LANGUAGE_SESSION_KEY:
> The language won’t be stored in the session in Django 4.0. Use the LANGUAGE_COOKIE_NAME cookie instead.

I'll change to that instead.

PR #197, opened 2021-08-18
KS-Backend: Replace django-cryptography

klemmari1:
## Description :sparkles:

Replace django-cryptography with django-searchable-encrypted-fields. Edit the migration `0001_initial.py` to remove the references to the old library. This should not affect any existing or new instances of the application.

## Issues :bug:

[YJDH-163](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-163)


backend/kesaseteli/applications/models.py
```diff
@@ -95,12 +95,10 @@ class SummerVoucher(HistoricalModel, UUIDModel):
...
-    employee_ssn = encrypt(
-        models.CharField(
-            max_length=32,
-            blank=True,
-            verbose_name=_("employee social security number"),
-        )
+    employee_ssn = EncryptedCharField(
+        max_length=32,
+        blank=True,
+        verbose_name=_("employee social security number"),
```
haavikko, line 101: If you need to enable "exact" queries by employee_ssn, do you also need to add a SearchField to the model in conjunction with the EncryptedField?

haavikko (approved):
Looks good to me! Depending on your requrements, you may want to add a SearchField to the model?

backend/kesaseteli/applications/models.py
klemmari1: Good point! As far as I know, we won't need to query the SSN field though.

quyenlq (approved):

PR #200, opened 2021-08-19
Fix frontend next build

nenonja:
## Description :sparkles:
It seems that next-build is broken in dev-mode after next.js v11.1.0 update. Here is the fix for it.


klemmari1 (approved):
Works for me :+1: GJ

PR #206, opened 2021-08-23
HL-128 | Add audit log sender

quyenlq:
## Description :sparkles:
- Added Elasticsearch client
- Task to index audit log to ElasticSearch every 15 minutes
- A utility task to clean sent audit log every month
- Cronjob config for review env
- Added created_at timestamp to AuditLogEntry model


backend/shared/shared/audit_log/tasks.py
```diff
@@ -0,0 +1,45 @@
...
+    ):
+        LOGGER.warning(
+            "Trying to send audit log to Elasticsearch without proper configuration, process skipped"
+        )
+        return
+    es = Elasticsearch(
+        cloud_id=settings.ELASTICSEARCH_CLOUD_ID,
+        api_key=(settings.ELASTICSEARCH_API_ID, settings.ELASTICSEARCH_API_KEY),
+    )
+    entries = AuditLogEntry.objects.filter(is_sent=False)
```
haavikko, line 28: Should there be an order_by() clause here?

backend/shared/shared/audit_log/tasks.py
```diff
@@ -0,0 +1,45 @@
...
+            entry.save()
+
+
+def clear_audit_log_entries(days_to_keep=30):
+    # Only remove entries older than `X` days
+    sent_entries = AuditLogEntry.objects.filter(
+        is_sent=True, created_at__lte=(timezone.now() - timedelta(days=days_to_keep))
+    )
+    for entry in sent_entries:
+        entry.delete()
```
haavikko, line 45: Do the entries need to be deleted one by one, or would a bulk operation be more efficient?

backend/shared/shared/audit_log/jobs/quater_hourly/send_audit_log.py
```diff
@@ -0,0 +1,12 @@
...
+
+from shared.audit_log.tasks import send_audit_log_to_elastic_search
+
+
+class Job(QuarterHourlyJob):
+    help = "Send AuditLogEntry to centralized log center every 15 minutes"
+
+    def execute(self):
+        if settings.ENABLE_SEND_AUDIT_LOG:
+            send_audit_log_to_elastic_search()
```
haavikko, line 12: What happens if send_audit_log_to_elastic_search throws an exception? Maybe add the call in a try-catch block and log exception content?

haavikko (approved):
Looks good! Please check the questions in comments, address as you see appropriate&go ahead with merging!

backend/shared/shared/audit_log/tasks.py
```diff
@@ -0,0 +1,45 @@
...
+        return
+    es = Elasticsearch(
+        cloud_id=settings.ELASTICSEARCH_CLOUD_ID,
+        api_key=(settings.ELASTICSEARCH_API_ID, settings.ELASTICSEARCH_API_KEY),
+    )
+    entries = AuditLogEntry.objects.filter(is_sent=False)
+
+    for entry in entries:
+        rs = es.index(
+            index=settings.ELASTICSEARCH_APP_AUDIT_LOG_INDEX, body=entry.message
```
klemmari1, line 32: Should we also index the django ID here so it is easier to find the correct logs from ES if that is ever needed.

```
rs = es.index(
            index=settings.ELASTICSEARCH_APP_AUDIT_LOG_INDEX, id=entry.id, body=entry.message
)
```

quyenlq: That'd work too. By default, an index id will be generated but having the index id as a reference to the object id might be useful somehow.

klemmari1: Yea, I was thinking if there is some special case in which it would be useful. For example there were some unnecessary logs sent that need to be deleted.

backend/shared/shared/audit_log/jobs/quater_hourly/send_audit_log.py
quyenlq: AFAIK, if the send audit log task fails, we want it to raise an error and that error will be captured to Sentry. I guess we just leave it to fail naturally here. There is nothing else we have to do rather than try to send the log in the next quarter-hour. Please correct me if I'm wrong :)

backend/shared/shared/audit_log/tasks.py
quyenlq: Yeah, that'd make more sense than delete them one by one. I'll fix it

klemmari1 (approved):
LGTM!

PR #210, opened 2021-08-23
YJDH-164 YJDH-173 | Add Attachment model & API

klemmari1:
## Description :sparkles:

Add a model for attachments and allow uploading attachments for summer vouchers.

Add a docker volume in `/backend/kesaseteli/media` for the attachments that is mounted in `/var/media/`. Add the media folder to gitignore.

## Issues :bug:

[YJDH-164](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-164)
[YJDH-173](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-173)


backend/kesaseteli/applications/models.py
```diff
@@ -115,7 +122,29 @@ class SummerVoucher(HistoricalModel, UUIDModel):
...
+    )
+    content_type = models.CharField(
+        max_length=100,
+        choices=ATTACHMENT_CONTENT_TYPE_CHOICES,
+        verbose_name=_("technical content type of the attachment"),
+    )
+    attachment_file = models.FileField(verbose_name=_("application attachment content"))
+
+    class Meta:
+        unique_together = [("summer_voucher", "attachment_type")]
```
haavikko, line 150: In Helsinki benefit I had to change attachment validation rules to allow multiple files of the same attachment type, because a single document might have multiple pages and the users would upload each page as a separate png file - don't know if the same applies here?

backend/kesaseteli/applications/models.py
```diff
@@ -1,11 +1,18 @@
...
 
-from applications.enums import ApplicationStatus
+from applications.enums import ApplicationStatus, AttachmentType
 from companies.models import Company
 
+ATTACHMENT_CONTENT_TYPE_CHOICES = (
+    ("application/pdf", "pdf"),
+    ("image/png", "png"),
+    ("image/jpeg", "jpeg"),
+    ("image/jpg", "jpg"),
```
haavikko, line 13: would image/jpg ever be used? I understand it's non-standard

haavikko (approved):
Looks good to me! Please check the question about unique_together, in case you have the same issue as Helsinki-benefit

backend/kesaseteli/applications/models.py
klemmari1: I have to ask about this.

backend/kesaseteli/applications/models.py
klemmari1: Yea, with some further investigation this should never occur. I'll remove it.

backend/kesaseteli/applications/models.py
klemmari1: Yea we will go with the same approach of allowing multiple attachments.

PR #212, opened 2021-08-24
HL-159 | Upgrade relevant dependencies to fix Styled-Components after NextJS upgrade

Nizar-Rahme:
## Description :sparkles:
- Remove redundant package-lock file
- Upgrade babel plugins
- Replace Babel config function with an object
- Use HDS footer dark theme
- Replace styled-components default import name


nenonja (approved):
Looks good!

PR #220, opened 2021-08-27
YJDH-171 | KS-Backend: Add field validation and queryset ordering

klemmari1:
## Description :sparkles:

Add validation for fields in the Rest API. Validate status and
that fields are not null when submitting the application.

Apply ordering for models Application, SummerVoucher and Attachment.
Add ordering field for summer vouchers so that the order can be
changed from the frontend.

## Issues :bug:

[YJDH-171](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-171)


haavikko (commented):
Looks good to me!

PR #226, opened 2021-08-30
Backend: Fix rare race condition in mock authentication

klemmari1:
## Description :sparkles:

Handle a rare race condition in the mocked rest authentication.

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "oidc_oidcprofile_user_id_key"
```

### DIFF
--- a/backend/shared/shared/oidc/auth.py
+++ b/backend/shared/shared/oidc/auth.py
@@ -3,6 +3,7 @@ import logging
 import requests
 from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
+from django.db.utils import IntegrityError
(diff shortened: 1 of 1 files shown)


haavikko (approved):
Looks good to me! The factories are only used for running test code, right?

klemmari1:
The factories here are used for the mocked login. So if you put `MOCK_FLAG=1` and go to the home page and click "log in" it won't redirect you to suomi.fi but do a local login. It will create a dummy user with dummy company, `OIDCProfile` and `EAuthorizationProfile`. This is used mainly by the browser tests but can be used by the developers as well.

PR #227, opened 2021-08-30
Yjdh-135: Preserve selected language through the suomifi login

nenonja:
## Description :sparkles:
Application will remember the previously selected langauge when coming back from suomifi login. Not only finnish but other languages too. Includes a new browser test that checks that the feature works with each foreign language (sv and en)

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-135


frontend/shared/browser-tests/utils/settings.ts
```diff
@@ -1,4 +1,7 @@
 export const getUrl = (baseUrl = '', path?: string): string =>
   `${baseUrl}${path?.startsWith('/') ? path : path ? `/${path}` : ''}`;
 
-export const SuomiFiAuthorizationUrl = 'https://testi.apro.tunnistus.fi/';
+export const SuomiFiAuthorizationUrls = [
```
nenonja, line 4: Interestingly enough, it seems that sometimes suomifi test environment points to a different address, so here is the list all possible addresses witnessed so far. These are used by browser tests.

PR #228, opened 2021-08-31
HL-149 | Export Application Batch to PDF before sending to AHJO

quyenlq:
## Description :sparkles:
- Add API endpoint to export application batch
- Each application in a batch will be generated to a single PDF files, the template will be decided based on the application type (company type, deminimis aid)
- Composed files are also generated, each will have two version: private and public, where the public pdf won't include employee data

NOTE: After the demo with PO, it turned out that the template should be updated so it can contain more information of the application, also design for the template might be considered later. So this PR is basically focus on the export function and its logic, any change to the template can be update later in a separated PR

## Screenshots :camera_flash:
Some example PDF:
Public composed file
[Liite 1 Helsinki-lisä 2021 koontiliite yhteisöille mallipohja julkinen.pdf](https://github.com/City-of-Helsinki/yjdh/files/7084038/Liite.1.Helsinki-lisa.2021.koontiliite.yhteisoille.mallipohja.julkinen.pdf)
Private composoed file
[Helsinki-lisä 2021 koontiliite yhteisöille mallipohja salassa pidettävä.pdf](https://github.com/City-of-Helsinki/yjdh/files/7084045/Helsinki-lisa.2021.koontiliite.yhteisoille.mallipohja.salassa.pidettava.pdf)

Single application file
[Liite 2 hakijakohtainen mallipohja yhteisöt R125000.pdf](https://github.com/City-of-Helsinki/yjdh/files/7084046/Liite.2.hakijakohtainen.mallipohja.yhteisot.R125000.pdf)

### DIFF
--- a/backend/benefit/applications/services/ahjo_integration.py
+++ b/backend/benefit/applications/services/ahjo_integration.py
@@ -1,0 +1,209 @@
+import os
+import zipfile
+from io import BytesIO
+
+import jinja2
+import pdfkit
+from applications.enums import ApplicationStatus, OrganizationType
+from django.utils import timezone
+
+PDF_PATH = os.path.join(os.path.dirname(__file__) + "/pdf_templates")
+TEMPLATE_ID_ASSOCIATION = "associations"
+TEMPLATE_ID_COMPANIES_BENEFIT_WITH_DE_MINIMIS_AID = (
+    "companies_benefit_with_de_minimis_aid"
+)
+TEMPLATE_ID_COMPANIES_BENEFIT_WITHOUT_DE_MINIMIS_AID = (
+    "companies_benefit_without_de_minimis_aid"
+)
+TEMPLATE_ID_COMPOSED_ACCEPTED_PUBLIC = "composed_accepted_public"
+TEMPLATE_ID_COMPOSED_ACCEPTED_PRIVATE = "composed_accepted_private"
+TEMPLATE_ID_COMPOSED_DECLINED_PUBLIC = "composed_declined_public"
+TEMPLATE_ID_COMPOSED_DECLINED_PRIVATE = "composed_declined_private"
+
+JINJA_TEMPLATES_COMPOSED = {
+    TEMPLATE_ID_COMPOSED_ACCEPTED_PUBLIC: {
+        "path": "composed_accepted_public.html",
+        "file_name": "Liite 1 Helsinki-lisä 2021 koontiliite yhteisöille "
+        "mallipohja julkinen.pdf",
+    },
+    TEMPLATE_ID_COMPOSED_ACCEPTED_PRIVATE: {
+        "path": "composed_accepted_private.html",
+        "file_name": "Helsinki-lisä 2021 koontiliite yhteisöille mallipohja "
+        "salassa pidettävä.pdf",
+    },
+    TEMPLATE_ID_COMPOSED_DECLINED_PUBLIC: {
+        "path": "composed_declined_public.html",
+        "file_name": "Helsinki-lisä 2021 kielteiset päätökset hakijakohtainen "
+        "mallipohja {start_number}-{end_number} salassa "
+        "pidettävä.pdf",
+    },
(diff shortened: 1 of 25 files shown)


backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -0,0 +1,213 @@
...
+from io import BytesIO
+
+import jinja2
+import pdfkit
+from applications.enums import ApplicationStatus, OrganizationType
+from common.exceptions import BenefitAPIException
+from django.utils import timezone
+from django.utils.translation import gettext_lazy as _
+
+PDF_PATH = os.path.join(os.path.dirname(__file__) + "/pdf_templates")
```
klemmari1, line 12: Should this path be defined in the settings?

backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -0,0 +1,213 @@
...
+
+    return pdf_files
+
+
+def generate_single_file(app):
+    # Association
+    if (
+        OrganizationType.resolve_organization_type(app.company.company_form)
+        == OrganizationType.ASSOCIATION
+    ):
```
haavikko, line 91: What about associations that do have business activities and have de minimis aid granted? Do they use the same template as other associations?

backend/benefit/applications/services/pdf_templates/associations.html
```diff
@@ -0,0 +1,40 @@
+<html>
+<meta http-equiv="Content-type" content="text/html; charset=utf-8" />
```
haavikko, line 2: Is it possible to factor the templates so that the common code (`<head>` and such) would be only defined once?

backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -0,0 +1,213 @@
...
+
+    with zipfile.ZipFile(mem_zip, mode="w", compression=zipfile.ZIP_DEFLATED) as zf:
+        for f in files:
+            zf.writestr(f[0], f[1])
+
+    return mem_zip.getvalue()
+
+
+def export_application_batch(batch):
+    if batch.applications.count() <= 0:
```
haavikko, line 205: Handling of the application batch status: When the data is exported, application batch status should already be SENT_TO_AHJO, or alternatively the status should be set to SENT_TO_AHJO here if it is still DRAFT. The Talpa guys had requirement that the applications should be immutable after the ahjo data is generated.
Also, we need to save the id of the person doing the state transfer to SENT_TO_AHJO, that is needed in Talpa. I guess that would be automatically stored in the history table of the application batch.

haavikko (changes requested):

backend/benefit/applications/services/ahjo_integration.py
quyenlq: Thanks for the information, I updated the validation and tests

quyenlq: Btw, I think storing the ID of person doing the state transfer should be done in a separated ticket leater, when we integrate the handler authentication API

backend/benefit/applications/services/pdf_templates/companies_benefit_with_de_minimis_aid.html
```diff
@@ -0,0 +1,24 @@
+{% block content %}
```
klemmari1, line 1: Are the other templates missing `{% extends "base.html" %}` ?

quyenlq: Nice catch, fixed them :)

backend/benefit/applications/services/ahjo_integration.py
haavikko: OK, I'll verify the issue with the admins, we can address it in a subsequent PR if needed

haavikko (approved):
Looks good to me!

PR #236, opened 2021-08-31
YJDH-165| KS-Backend: Add application API access control

klemmari1:
## Description :sparkles:

Add `user` ForeignKey for the applications. This will make sure that applications can only be viewed by the correct user & company combination. So if an user from the same company logs in, they cannot see an application that another user has created from the same company.

Delete all applications before adding the user foreign key. The applications are deleted because there is no reasonable default value for the user foreign key and there is no use for applications without an user.

Add permission checks for the REST API so that users can only see their own applications / summer vouchers / attachments.

## Issues :bug:

[YJDH-165](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-165)


haavikko (approved):
Looks good to me!

PR #252, opened 2021-09-03
HL-162: handle breaking changes in application

haavikko:
## Description :sparkles:
* Handle breaking changes in the application
* date range validation of previous benefits
* Handle validation of pay_subsidy_granted

## Issues :bug:
HL-162


backend/benefit/applications/migrations/0017_searchable_encrypted_social_security_number.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("applications", "0016_benefit_amount"),
+    ]
+
+    operations = [
+        migrations.RenameField(
```
quyenlq, line 14: According to `django-searchable-encrypted-field` document: https://gitlab.com/guywillett/django-searchable-encrypted-fields#migrations-add-searchencryptedfields-to-your-model-dont-alter-existing-fields. It is recommended to create a new encrypted field and migrate the old data to the new one.

Although we don't have any actual data anywhere, I think it's still better to at least create a whole new EncryptedField than renaming the old one

quyenlq (dismissed):

backend/benefit/applications/migrations/0017_searchable_encrypted_social_security_number.py
```diff
@@ -0,0 +1,31 @@
...
+            new_name="encrypted_social_security_number",
+        ),
+        migrations.AddField(
+            model_name="employee",
+            name="social_security_number",
+            field=encrypted_fields.fields.SearchField(
+                blank=True,
+                db_index=True,
+                encrypted_field_name="encrypted_social_security_number",
+                hash_key="ee235e39ebc238035a6264c063dd829d4b6d2270604b57ee1f463e676ec44669",
```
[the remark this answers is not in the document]
haavikko: Yes, fixed to use settings.SOCIAL_SECURITY_NUMBER_HASH_KEY here, as we probably should
have different value in staging/prod. Although the value in migration doesn't probably matter as it's not part of db schema, it might be confusing.

backend/benefit/applications/migrations/0017_searchable_encrypted_social_security_number.py
haavikko: OK good catch, I added a pair of migrations to delete/add the field. (removing and then adding the same field in the same migration didn't seem to work)

backend/benefit/common/utils.py
```diff
@@ -31,8 +35,82 @@ def xgroup(iter, n=2, check_length=False):
             last = []
 
 
+def pairwise(iterable):
+    "s -> (s0,s1), (s1,s2), (s2, s3), ..."
+    a, b = itertools.tee(iterable)
+    for elem in b:
```
klemmari1, line 41: What is this for loop doing?

quyenlq (approved):
Looks good 👍

backend/benefit/common/utils.py
haavikko: it removes the first element from b, so that in zip(a,b) the iterable a always returns the preceding element

PR #254, opened 2021-09-07
Employee consent attachment

haavikko:
## Description :sparkles:
NOTE: this PR contains a migration, so tests only pass after the preceding migration in hl-162/application-breaking-changes is merged

Add new attachment type for the consent attachment and validation for the attachment

## Issues :bug:
HL-133


quyenlq (approved):
Test failing but this PR looks good already

PR #255, opened 2021-09-07
HL-163: Benefit Handler project setup

melniiv:
## Description :sparkles:
- github actions missing for the handler side


frontend/benefit/handler/src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,22 @@
+import { useTranslation } from 'benefit/handler/i18n';
```
nenonja, line 1: This could be in shared folder if importing straight from `import { useTranslation } from 'from 'next-i18next';` Kesaseteli could also use it, every service just then implement their own translation texts.

frontend/benefit/handler/src/components/header/useHeader.ts
```diff
@@ -0,0 +1,56 @@
+import useLocale from 'benefit/handler/hooks/useLocale';
```
nenonja, line 1: Even this could be in shared folder if useTranslation comes from `next-i18next` and `getLanguageOptions` is also moved to shared? I think we have this same header login already copy-pasted three times per each service but the code is actually 99% same.

frontend/benefit/handler/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@frontend/bf-handler",
```
nenonja, line 2: It's good idea to distinct the name with `bf-handler` because there will be `ks-handler` for kesäseteli at some point

frontend/benefit/handler/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
+const { join } = require('path');
+
+module.exports = {
```
nenonja, line 3: to avoid copy-paste, this is only javascript object which could be partially imported from for example `frotend/.eslinrc.js` root folder and then overwrite the parts which are project-specified.
Something like
```
const baseConfig = require ('../../eslintrc.js`

module.exports = {
    ...baseConfig,
    rules: {
       ...baseConfig.rules,
       (custom rules)
    },
    overrides: [ ...baseConfig.overrides, (custom overrides) ],
}
```

frontend/benefit/handler/Dockerfile
```diff
@@ -0,0 +1,104 @@
+# =======================================
```
nenonja, line 1: dockerfile is also 95% copy-paste for each project. Perhaps we could make a jira ticket to use template and give the folder `benefit/handler` and some others as argument. Something like this: https://blog.dockbit.com/templating-your-dockerfile-like-a-boss-2a84a67d28e9

frontend/benefit/handler/next.config.js
```diff
@@ -0,0 +1,42 @@
+const webpack = require('webpack');
```
nenonja, line 1: Perhaps this next.config.js is exactly the same for each project? This could be also moved to `frontend` root folder to avoid copy-paste?

docker-compose.benefit.yml
```diff
@@ -47,6 +47,21 @@ services:
...
+    build:
+      context: ./frontend
+      dockerfile: benefit/handler/Dockerfile
+      target: development
+    volumes:
+      - ./frontend:/app
+      - /app/node_modules
+      - /app/.next
+    env_file:
+      - .env.benefit
```
nenonja, line 60: perhaps handler needs its own env params?

nenonja (commented):
Some comments how to avoid the sonarcube's alerted copy-paste. I would like to move the some common components to shared folder as we got them now three times in repo. Otherwise I think it's up to you how far to go avoiding copy-paste... I didnt try to setup this on my computer. Ultimately we need to see the review environment deploying and working to see that everything works correctly but it can be done later.

frontend/benefit/handler/Dockerfile
melniiv: yea, let create a ticket

frontend/benefit/handler/src/i18n.ts
```diff
@@ -0,0 +1,3 @@
+import { appWithTranslation, i18n, useTranslation } from 'next-i18next';
```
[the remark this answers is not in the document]
melniiv: fixed

frontend/benefit/handler/src/utils/common.ts
```diff
@@ -0,0 +1,18 @@
+import { TFunction } from 'next-i18next';
```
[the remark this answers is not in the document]
melniiv: fixed

frontend/benefit/handler/src/pages/_app.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <BackendAPIProvider
+      baseURL={getBackendDomain()}
+      headers={getHeaders(locale)}
+    >
+      <QueryClientProvider client={queryClient}>
+        <ThemeProvider theme={theme}>
+          <GlobalStyling />
+          <Layout>
+            <Header /> d
```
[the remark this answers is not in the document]
melniiv: yes, fixed

frontend/benefit/handler/src/components/header/useHeader.ts
melniiv: fixed

frontend/benefit/handler/src/components/footer/Footer.tsx
melniiv: fixed

frontend/benefit/handler/.eslintrc.js
melniiv: good point

frontend/benefit/handler/next.config.js
melniiv: not exactly the same

frontend/benefit/handler/.eslintrc.js
melniiv: but when comparing even kesaseteli one and applicant not that much in common, more customisations then "base". lets leave it for right now.

frontend/benefit/handler/src/hooks/useLocale.ts
```diff
@@ -0,0 +1,19 @@
+import { useTranslation } from 'benefit/handler/i18n';
```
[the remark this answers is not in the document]
melniiv: fixed

frontend/benefit/handler/next.config.js
melniiv: fixed

docker-compose.benefit.yml
melniiv: done

frontend/benefit/handler/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "next-i18next": "^8.4.0",
+    "next-plugin-custom-babel-config": "^1.0.5",
+    "next-transpile-modules": "^8.0.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-i18next": "^11.11.0",
+    "react-input-mask": "^2.0.4",
+    "react-loading-skeleton": "^2.2.0",
+    "react-query": "^3.17.0",
+    "react-toastify": "^8.0.2",
```
Nizar-Rahme, line 43: I suggest adding these packages only when they're needed.

frontend/benefit/handler/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "rexreplace": "^7.0.3",
```
Nizar-Rahme, line 44: I saw this package listed in other services, but are we using it?

frontend/benefit/handler/src/utils/forms.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/** Get error text
+ * @param {Object} errors
+ * @param {Object} touched
+ * @param {string} name
+ * @param {Function} t
+ * @param {boolean} isSubmitted
+ * @return {string}
+ */
+export const getErrorText = (
```
Nizar-Rahme, line 12: Should we move this to the shared folder?

Nizar-Rahme (approved):
I added a couple of comments you may consider, but overall, it's a good start 👍

frontend/benefit/handler/src/utils/forms.ts
melniiv: its formik related, in ks we are not using formik, but probably will in handler. i would want to keep shared independent (one thing to refactor is to remove nextjs dependency from there at some point as well)

frontend/benefit/handler/package.json
melniiv: fixed

frontend/benefit/handler/package.json
melniiv: fixed

frontend/benefit/handler/src/utils/forms.ts
Nizar-Rahme: Alright 👌

PR #260, opened 2021-09-08
YJDH-175 | Backend: Save organization roles in session

klemmari1:
## Description :sparkles:

Refactor get_organization_roles so that the roles are saved in session after login. After login the roles will be fetched from the session.

This is done because the access token that we get from suomi.fi is used only for getting these roles and the ttl of the token is only 10 minutes. So later when calling this function we can be assured that the roles are already fetched and there is no need to call the suomi.fi organizationRoles API.

## Issues :bug:

[YJDH-175](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-175)

### DIFF
--- a/backend/shared/shared/oidc/auth.py
+++ b/backend/shared/shared/oidc/auth.py
@@ -165,4 +165,9 @@ class EAuthRestAuthentication(SessionAuthentication):
         ):
             return None
 
+        # Store organization roles in session
+        from shared.oidc.utils import request_organization_roles
+
+        request_organization_roles(user.oidc_profile.eauthorization_profile, request)
+
         return user, auth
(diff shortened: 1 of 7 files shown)


backend/shared/shared/oidc/auth.py
```diff
@@ -165,4 +165,9 @@ def authenticate(self, request):
         ):
             return None
 
+        # Store organization roles in session
+        from shared.oidc.utils import request_organization_roles
+
+        request_organization_roles(user.oidc_profile.eauthorization_profile, request)
```
quyenlq, line 171: Should we also clear the session when user log out?

klemmari1: I think the session is flushed when user logs out, correct?

quyenlq: Ah, true 👍 my bad, I thought the backchannel log out is using a different method to log out

quyenlq (approved):
LGTM!

PR #262, opened 2021-09-08
HL-158 | Add Company department field

quyenlq:


backend/benefit/companies/models.py
```diff
@@ -8,6 +9,7 @@ class Company(AbstractCompany):
     bank_account_number = IBANField(
         include_countries=("FI",), verbose_name=_("bank account number")
     )
+    department = CharField(verbose_name=_("department"), max_length=256, blank=True)
```
haavikko, line 12: Do we need a department field in Company model and where does the value for this field come from?
One company would have multiple departments, so is the department useful as a single charfield?
Maybe, as part of Yritysprofiili, we'd implement department info, but maybe we'd make a separate design for it later?

haavikko (approved):
Check out the question about Company.department field, I'll leave it for your decision, but other than that, looks good to me!

backend/benefit/companies/models.py
quyenlq: Hmn, yeah I think you are right, I don't think we need the department field here. I'll fix it

PR #268, opened 2021-09-10
HL-154: data model for applicant terms

haavikko:
## Description :sparkles:

Data model for applicant terms.
In order to implement approval of Terms of service later on, common functionality is in AbstractTermsApproval.
API with validation and more tests coming in another pull request.

## Issues :bug:
HL-154

## Testing :alembic:
backend/benefit/terms/tests/test_models.py


backend/benefit/terms/models.py
```diff
@@ -0,0 +1,115 @@
...
+    terms = models.ForeignKey(
+        Terms,
+        verbose_name=_("terms"),
+        on_delete=models.CASCADE,
+    )
+    """
+    API validates that the selected_applicant_consents must belong to the terms of this approval
+    """
+    selected_applicant_consents = models.ManyToManyField(
+        ApplicantConsent,
```
klemmari1, line 94: Missing verbose name.

Also missing `related_name` unless the default is OK.

backend/benefit/terms/models.py
```diff
@@ -0,0 +1,115 @@
...
+    approved_at = models.DateTimeField(verbose_name=_("timestamp of approval"))
+    approved_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        verbose_name=_("user who approved the terms"),
+    )
+    terms = models.ForeignKey(
+        Terms,
+        verbose_name=_("terms"),
+        on_delete=models.CASCADE,
```
klemmari1, line 88: Missing related_name unless the default is OK.

backend/benefit/terms/models.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class AbstractTermsApproval(UUIDModel, TimeStampedModel):
+    approved_at = models.DateTimeField(verbose_name=_("timestamp of approval"))
+    approved_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        verbose_name=_("user who approved the terms"),
+    )
+    terms = models.ForeignKey(
```
klemmari1, line 85: "API validates that the selected_applicant_consents must belong to the terms of this approval"

I might be missing something here but if there can be multiple `selected_applicant_consents` should there also be multiple `terms`? So is the logic here that there are a set of terms that all need to be consented by the user?

What I'm getting at here is that should `terms` be a `ManyToManyField` as well?

backend/benefit/terms/models.py
haavikko: Added verbose_name, good catch

the default related_name is ok, also when using ManyToManyField in abstract base, would need to use "%(app_label)s_%(class)s" substitution in the related_name, I left it out to simplify it.

backend/benefit/terms/models.py
haavikko: I let the system generate related_name here, too, so don't need to use app_label substitutio

backend/benefit/terms/models.py
haavikko: Yes, it's a bit unclear in the code, so I'm adding the following as a comment in the code:

The "terms approval" process in UI has two steps:
1. User is shown a PDF file with terms
2. User needs to click a set of mandatory checkboxes in order to proceed

In discussion with the product owners, we gathered a few requirements for the backend:
1. The terms might change as a result of political decisions by the city, or for other reasons
2. The set of checkboxes shown to the applicant may change, too
3. We need to store the exact version of terms that was shown to the user
4. It's not enough to just store the terms, we need to store also the fact that the user checked the required checkboxes (even if the checkboxes are all mandatory)

So, here, the AbstractTermsApproval is used the store the fact that:
1. The user has been shown the terms pdf document (=one Terms object)
2.  The user has checked all the checkboxes (=selected_applicant_consents) defined by that Terms object (this is actually redundant data, but I guess it helps us meet some auditing requirements). If we didn't have that requirement, the data model would work fine with just a ForeignKey to Terms object, as the applicant_consents can be accessed through it.

backend/benefit/terms/models.py
haavikko: Oops forgot to push the changes yesterday, done now

backend/benefit/terms/models.py
klemmari1: Okay, makes sense :+1:

klemmari1 (approved):
Looks good after the style test is fixed :+1:

backend/benefit/terms/models.py
```diff
@@ -0,0 +1,135 @@
...
+        Terms,
+        verbose_name=_("terms"),
+        on_delete=models.CASCADE,
+    )
+    """
+    API validates that the selected_applicant_consents must belong to the terms of this approval
+    """
+    selected_applicant_consents = models.ManyToManyField(
+        ApplicantConsent,
```
quyenlq, line 94: I wonder if should we add Company FK to this model as well because it's needed for the `TERMS_OF_SERVICE` terms_type

quyenlq (approved):
Looks great! It's good that we can use one Terms model in both Terms of Service and Applicant Terms

PR #269, opened 2021-09-10
KS-Backend: Keep IDs on SummerVoucher updates

klemmari1:
## Description :sparkles:

Frontend relies on the summer voucher IDs staying the same between updates. Add a list serializer for the summer vouchers so that we can update the existing objects instead of deleting everything and creating new objects.

## Issues :bug:

-


haavikko (approved):
Looks good! Yes, the same issue exists with DeMinimisAid in Helsinki-benefit, similar approach would be useful for us too.

PR #270, opened 2021-09-10
HL-144 & HL-145 & HL-31 | Validation, Refactoring Styles and Typings

Nizar-Rahme:
## Description :sparkles:
- Made the typings of steps fields more strict. Each step can access its own fields; otherwise, it will throw an error.
- Got rid of most of the styled containers. 
- Implemented `$Grid` & `$GridCell` components that will function as wrappers to almost all the other elements and apply consistent styles across the application.
- Added validation to the second step of the application.
- As a bonus, the application now looks acceptable on mobile devices.


frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
```diff
@@ -35,9 +42,62 @@ type ExtendedComponentProps = {
   getDefaultSelectValue: (fieldName: keyof Application) => OptionType;
 };
 
+const ssnValidationTable = [
```
melniiv, line 45: https://www.npmjs.com/package/finnish-ssn
Could we use it?

frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
```diff
@@ -35,9 +42,62 @@ type ExtendedComponentProps = {
...
+
+  if (!validateSsn(number, checkCharacter)) return false;
+
+  return true;
+};
+
 const useApplicationFormStep2 = (
   application: Application
-): ExtendedComponentProps => {
+): UseApplicationFormStep2Props => {
```
melniiv, line 100: What is the point of renaming to "Use..."?

melniiv (approved):
looks good!

frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
Nizar-Rahme: I replaced our implementation with this dependency.

frontend/benefit/applicant/src/components/applications/forms/application/step2/useApplicationFormStep2.ts
Nizar-Rahme: Nothing really 😅 
I saw some other hooks following this naming convention; so I thought, why not this one as well?

PR #271, opened 2021-09-13
WIP: YJDH-198 | Improve mock login

klemmari1:
## Description :sparkles:

This will be released probably after MVP since this is not crucial. There is a browser test that needs to be adjusted for this as well.


klemmari1:
This is more of a quality of life improvement for testing the service with the MOCK_FLAG set on. At least one browser test is depending on the old implementation so those should be adjusted to this new implementation if this is ever released.

PR #272, opened 2021-09-13
Fix snapshot test

quyenlq:
## Description :sparkles:
Update snapshot data.
If the test failed randomly next time, we can consider switching to another method

### DIFF
--- a/backend/benefit/applications/tests/snapshots/snap_test_ahjo_integration.py
+++ b/backend/benefit/applications/tests/snapshots/snap_test_ahjo_integration.py
@@ -52,9 +52,9 @@ snapshots[
         <td>Y125007</td>
         <td>Example.</td>
         <td>8906333-4</td>
-        <td>Kimberly</td>
-        <td>Baker</td>
-        <td>2021-07-02</td>
+        <td>Renee</td>
+        <td>Hartman</td>
+        <td>2021-01-18</td>
     </tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


haavikko (approved):
This looks good and fixes the immediate issue, but I'm not sure if this works for the long term, if more tests are added.

I tested running individual tests like this, and it results in failure, so I think the values generated by the faker library depend on the ordering of the test executions. 

`pytest backend/benefit/applications/tests/ -ra -vvv -s --cov=. -k test_generate_composed_template_htm`

Maybe if reseed_random was called before running each test, then the execution order of tests would not matter? https://factoryboy.readthedocs.io/en/stable/recipes.html#recipe-random-management

quyenlq:
> This looks good and fixes the immediate issue, but I'm not sure if this works for the long term, if more tests are added.
> 
> I tested running individual tests like this, and it results in failure, so I think the values generated by the faker library depend on the ordering of the test executions.
> 
> `pytest backend/benefit/applications/tests/ -ra -vvv -s --cov=. -k test_generate_composed_template_htm`
> 
> Maybe if reseed_random was called before running each test, then the execution order of tests would not matter? https://factoryboy.readthedocs.io/en/stable/recipes.html#recipe-random-management

That could be one option, the other option is to not using the snapshot test but try to assert the data from the dynamic factories data generated inside the test. I'll make a separated ticket to fix this issue later

PR #277, opened 2021-09-14
Fix pytest to avoid random failed test

quyenlq:
## Description :sparkles:
Need to get rid of the snapshot test because it's not appropriate to use in this case. The `ahjo_application_number` is generated purely by the order of the test, so even we reseed every time the test is run, it's still a problem because we can't reset the PGSQL sequence.

Also changed some fields in the Factory, to avoid unexpected validation logic


haavikko (approved):
Looks good to me!

PR #278, opened 2021-09-15
YJDH-197 | KS-Employer: Login page

klemmari1:
## Description :sparkles:

Add login page styles and translations.

Modify the Layout shared component so that headingText is optional.

Invision: https://invis.io/WT10T1E97XKB#/453766715_Koti_Login

## Issues :bug:

[YJDH-197](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-197)

## Screenshots :camera_flash:

Screenshot from 2021-09-15 11-18-49

Screenshot from 2021-09-15 11-19-53

Screenshot from 2021-09-15 11-20-42

### DIFF
--- a/frontend/kesaseteli/employer/public/locales/fi/common.json
+++ b/frontend/kesaseteli/employer/public/locales/fi/common.json
@@ -11,6 +11,13 @@
     "retry": "Lataa sivu uudelleen",
     "logout": "Kirjaudu ulos"
   },
+  "loginPage": {
+    "logoutMessageLabel": "Olet kirjautunut ulos",
+    "errorLabel": "Tapahtui tuntematon virhe. Kirjaudu uudelleen sisään",
+    "sessionExpiredLabel": "Käyttäjäsessio vanhentui. Kirjaudu uudelleen sisään",
+    "infoLabel": "Palvelun käyttäminen edellyttää vahvaa tunnistautumista",
+    "infoContent": "Qui consequat qui tempor ut. Sit veniam esse in consectetur eiusmod id aliquip excepteur deserunt enim eiusmod proident amet incididunt. Minim occaecat laborum enim sint in cupidatat. Mollit elit aliqua laboris qui ipsum laboris. Labore incididunt Lorem labore commodo."
+  },
   "application": {
     "new": "Uusi hakemus",
     "common_error": "Tapahtui tuntematon virhe",
--- a/frontend/kesaseteli/employer/public/locales/sv/common.json
+++ b/frontend/kesaseteli/employer/public/locales/sv/common.json
@@ -11,6 +11,13 @@
     "retry": "SV Lataa sivu uudelleen",
     "logout": "SV Kirjaudu ulos"
   },
+  "loginPage": {
+    "logoutMessageLabel": "Olet kirjautunut ulos SWE",
+    "errorLabel": "Tapahtui tuntematon virhe. Kirjaudu uudelleen sisään SWE",
+    "sessionExpiredLabel": "Käyttäjäsessio vanhentui. Kirjaudu uudelleen sisään SWE",
+    "infoLabel": "Palvelun käyttäminen edellyttää vahvaa tunnistautumista SWE",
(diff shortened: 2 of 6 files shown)


frontend/kesaseteli/employer/src/pages/login.tsx
```diff
@@ -1,25 +1,52 @@
...
+
 const Login: NextPage = () => {
   useClearQueryParams();
+  const { t } = useTranslation();
   const {
     query: { logout, error, sessionExpired },
   } = useRouter();
+  const login = useLogin();
+
+  let notificationLabel = t(`common:loginPage.infoLabel`)
```
nenonja, line 23: ususally using `let` should be avoided, perhaps we could use method here: `const notificationKey = getNotificationLabelKey(query)` ans put the logic into method. then use it `t(notificationKey)`

frontend/kesaseteli/employer/src/pages/login.tsx
```diff
@@ -1,25 +1,52 @@
...
+
   return (
     <Container>
-      <Layout headingText="Työnantajan liittymä">
-        {logout && <p>Olet kirjautunut ulos</p>}
-        {error && <p>Tapahtui tuntematon virhe. Kirjaudu uudelleen sisään.</p>}
-        {sessionExpired && (
-          <p>Käyttäjäsessio vanhentui. Kirjaudu uudelleen sisään.</p>
-        )}
+      <Layout headingText="">
```
nenonja, line 36: empty values are discouraged, perhaps you could change the Layout component so that it allows headingText to be optional

nenonja (commented):
Looks good! Few minor comments, which should be easily addressed

frontend/kesaseteli/employer/public/locales/fi/common.json
```diff
@@ -11,6 +11,14 @@
     "retry": "Lataa sivu uudelleen",
     "logout": "Kirjaudu ulos"
   },
+  "loginPage": {
+    "logoutLabel": "Olet kirjautunut ulos",
```
nenonja, line 15: you could put `SWE` in the end of the sentence so it's easier to see that the translation works.

frontend/kesaseteli/employer/public/locales/sv/common.json
```diff
@@ -11,6 +11,14 @@
     "retry": "SV Lataa sivu uudelleen",
     "logout": "SV Kirjaudu ulos"
   },
+  "loginPage": {
+    "logoutLabel": "Olet kirjautunut ulos",
```
nenonja, line 15: logoutLabel and loginLabel are duplicates, you could use the same keys for them as the header does

frontend/kesaseteli/employer/src/pages/login.tsx
```diff
@@ -1,25 +1,57 @@
...
+  const { t } = useTranslation();
   const {
     query: { logout, error, sessionExpired },
   } = useRouter();
+  const login = useLogin();
+
+  const getNotificationLabelKey = (): string => {
+    let notificationLabel = `common:loginPage.infoLabel`
+    if (logout) {
+      notificationLabel = `common:loginPage.logoutLabel`
```
nenonja, line 26: nitpick but you could return instantly so no need to use let variable.

nenonja: Also, we could re-use `common:header.loginLabel` and `common:header.logoutLabel`. The name header is misleading although it could be renamed to `common:user.logoutLabel` or something

nenonja (approved):
One small comment, but approved already because nitpick only ✅ . Also It would be easier to re-review if you do separate fix commit for previous review fixes.

frontend/kesaseteli/employer/src/pages/login.tsx
klemmari1: Yea, I'll return instantly :+1: 

logoutLabel is a bit different here:
```
  "loginPage": {
    "logoutLabel": "Olet kirjautunut ulos",
    ...
  },
```
```
  "header": {
    "logoutLabel": "Kirjaudu ulos",
    ...
  },
```

klemmari1: I renamed `loginPage.logoutLabel` -> `loginPage.logoutMessageLabel` to make it a bit clearer.

PR #279, opened 2021-09-15
Fix validation to return values

quyenlq:
## Issues :bug:
Missing return statement causing the serializer field will be always None after validation


klemmari1 (approved):
LGTM!

haavikko (approved):
Looks good to me!

PR #288, opened 2021-09-21
HL-146: add DUMMY_COMPANY_FORM to change created company type during development

haavikko:
## Description :sparkles:
Set 
MOCK_FLAG=1
DUMMY_COMPANY_FORM=ry

in the env file to change the company form returned from backend

Move organization_type field from ApplicationSerializer to CompanySerializer so the value is accessible to the UI when application is being created.

## Issues :bug:

HL-146

## Screenshots :camera_flash:

image


klemmari1 (approved):
LGTM!

quyenlq (approved):
Good solution to temporary fix the company type problem 👍

PR #293, opened 2021-09-21
HL-187 | Update AHJO template

quyenlq:
## Screenshots :camera_flash:
Checkout this ticket https://helsinkisolutionoffice.atlassian.net/secure/RapidBoard.jspa?rapidView=204&projectKey=HL&modal=detail&selectedIssue=HL-187


backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -31,34 +26,27 @@
         "file_name": "Helsinki-lisä 2021 koontiliite yhteisöille mallipohja "
```
haavikko, line 26: I think "mallipohja" (=template) maybe shouldn't be part of the actual output file name, as that file is not a template

backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -71,119 +59,96 @@ def _get_template(path):
...
+    file_name = template_config["file_name"].format(company_name=company.name)
+    temp = _get_template(template_config["path"])
+    html = temp.render(apps=apps)
+    single_pdf = pdfkit.from_string(html, False)
+    return file_name, single_pdf, html
+
+
+def generate_single_approved_file(company, apps):
+    # FIXME: Need to change the logic later when we have multiple benefit per application
+    # Association without business activity
```
haavikko, line 94: I asked the PO's in Kanslia teams chat to provide an example where this scenario occurs

backend/benefit/applications/services/pdf_templates/base.html
```diff
@@ -4,21 +4,41 @@
...
+
     body {
-        font-family: Arial, sans-serif;
+        font-family: ArialMT, sans-serif;
     }
     </style>
 </head>
 <body>
+<div>
+    <img src="https://www.hel.fi/wps/wcm/connect/b3717cdb-45a7-4575-946b-5ea799506c80/1/HELSINKI_Tunnus_MUSTA.png?MOD=AJPERES" width="80" height="37">
```
haavikko, line 38: The POs have told that a hel.fi re-implementation is going on, this URL probably would change. Though I don't know where this file should be stored - does the pdf renderer find it if it's stored on local disk?

backend/benefit/applications/services/pdf_templates/benefit_declined.html
```diff
@@ -0,0 +1,23 @@
+{% extends "base.html" %}
+
+{% block content %}
+    <p>Työllisyydenhoidon Helsinki-lisä, kielteiset päätökset työnantajille</p>
+    <table aria-label="AHJO report""AHJO report" style="border: 1px solid black">
```
haavikko, line 5: Is report""AHJO  a typo?

haavikko (approved):
Looks good to me!
Please check comments, update if needed&merge at will :-)

backend/benefit/applications/services/pdf_templates/base.html
quyenlq: Unfortunately no, the pdfkit won't be able to render the file if src is relative path to the template. The only thing that works is backend hosting the image as a static file, then we use the url to that file here, but it should be over complicated as URL are different between env. 

I think we can fix it later when we launch the production, then we can host this logo in cloud storage and replace this with the URL there.

I'll leave this as it for now

backend/benefit/applications/services/pdf_templates/benefit_declined.html
quyenlq: It is, I'll fix it :)

backend/benefit/applications/services/ahjo_integration.py
quyenlq: Thanks, I'll update it

PR #295, opened 2021-09-21
HL-193 | Add employee consent validation

quyenlq:
## Description :sparkles:
Add consent validation so only one employee consent is allowed to upload


haavikko (approved):
Looks good to me!

PR #299, opened 2021-09-23
YJDH-203 | KS-Backend: Add cron jobs to clear old attachments & applications

klemmari1:
## Description :sparkles:

There is a need to delete attachments that are older than a year and applications older than 5 years. Add cron jobs to delete them.

## Issues :bug:

[YJDH-203](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-203)


haavikko (approved):
Looks good and  will be useful for Helsinki-Benefit too!

PR #300, opened 2021-09-24
YJDH-208 | KS-Backend: Roll back to sentry logging

klemmari1:
## Description :sparkles:

According to the latest information, we will be using a centralized Helsinki Sentry service for the error logging instead of the Azure application insights.

## Issues :bug:

[YJDH-208](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-208)


quyenlq (approved):
LGTM! 😂

PR #302, opened 2021-09-27
* HL-154: output full URLs instead of paths for applicant terms.

haavikko:
Additionally:
* fix issue with is_handler when DISABLE_AUTHENTICATION is set
* Add blank=True to the ApplicationBatch foreignkey, so testing through Django admin is easier.
* Add a debug_env URL to ease debugging of environment issues in Azure. Will be useful when setting up production env, as we probably won't get console access to pod


klemmari1 (approved):
LGTM!

PR #306, opened 2021-09-27
Backend: Small fixes for PLATTA

klemmari1:
## Description :sparkles:

- Fix Azure ADFS settings so that the whole application won't crash if the setup is missing.
- Update django environ so that the database url can be read correctly in PLATTA environment.

## Issues :bug:

-

### DIFF
--- a/backend/kesaseteli/kesaseteli/settings.py
+++ b/backend/kesaseteli/kesaseteli/settings.py
@@ -248,9 +248,9 @@ EAUTHORIZATIONS_API_OAUTH_SECRET = env.str("EAUTHORIZATIONS_API_OAUTH_SECRET")
 # Azure ADFS
 LOGIN_URL = "django_auth_adfs:login"
 
-ADFS_CLIENT_ID = env.str("ADFS_CLIENT_ID")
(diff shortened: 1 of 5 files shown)


backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -248,9 +248,9 @@
 # Azure ADFS
 LOGIN_URL = "django_auth_adfs:login"
 
-ADFS_CLIENT_ID = env.str("ADFS_CLIENT_ID")
-ADFS_CLIENT_SECRET = env.str("ADFS_CLIENT_SECRET")
-ADFS_TENANT_ID = env.str("ADFS_TENANT_ID")
+ADFS_CLIENT_ID = env.str("ADFS_CLIENT_ID") or "client_id"
```
quyenlq, line 251: I think env.str("ADFS_TENANT_ID", "client_id") works as well

quyenlq (approved):

backend/kesaseteli/kesaseteli/settings.py
klemmari1: Seemed that `env.str("ADFS_TENANT_ID", "client_id")` did not work if this was set explicitly: `ADFS_TENANT_ID=`

PR #316, opened 2021-09-29
HL-209: add the immediate manager checkbox to data model

haavikko:
## Description :sparkles:

Add field for storing the "immediate manager check" checkbox value.
* Initial value is None
* For companies, the value needs to stay None (as the applicant is never shown the checkbox in UI)
* For associations, the value needs to be changed to True before application is submitted.

## Issues :bug:
HL-209


klemmari1 (approved):
Looks good to me after the style tests have been fixed :+1:

PR #317, opened 2021-09-29
YJDH-218 | KS-Backend: Disable django admin in production

klemmari1:
## Description :sparkles:

Add a setting to disable django admin in the production. By default django admin is enabled but it can be disabled by setting ENABLE_ADMIN=False.

Small fix included:
- Fix typo in cron job path.

## Issues :bug:

[YJDH-218](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-218)


haavikko (approved):
Looks good to me!

PR #318, opened 2021-09-30
Upgrade django-environ so that database url can be read correctly in PLATTA environment

haavikko:
## Description :sparkles:
same change was done in Kesäseteli

### DIFF
--- a/backend/benefit/requirements.txt
+++ b/backend/benefit/requirements.txt
@@ -40,7 +40,7 @@ django-auth-adfs==1.7.0
     #   yjdh-backend-shared
 django-cors-headers==3.7.0
     # via -r requirements.in
-django-environ==0.4.5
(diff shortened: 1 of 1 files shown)


klemmari1:
Just to clarify, upgrading `django-environ` did not solve anything in the end, it was just testing if it would solve the problem. In the end, the problem was to add some env variables `PSQL_HOST` `PSQL_PORT` that were removed for some reason. Those are used by pgbouncer.

klemmari1 (approved):
I'll approve this but probably won't solve the problem if you have DB connection issues.

haavikko:
Yes, I was thinking the timeouts yesterday were due to this, but that was a transient error

haavikko:
Anyway it's good to have the same versions in helsinki-benefit and kesäseteli

PR #321, opened 2021-09-30
Fix CORS errors in PLATTA.

haavikko:
Also re-ordering of the settings.py lines so that comparison with Kesäseteli is easier in the future


klemmari1 (approved):
LGTM!

klemmari1:
BTW a good idea might be to have shared settings file that we extend in the projects so that we can keep the common things updated as well.

PR #322, opened 2021-10-01
Backend: Improve audit logging str

klemmari1:
## Description :sparkles:

Improve the audit log str so that it shows the status of the event and "UNKNOWN" instead of empty values.


haavikko (approved):
Looks good to me!

PR #328, opened 2021-10-04
HL-78 | Refactoring and testing

Nizar-Rahme:


frontend/benefit/applicant/src/hooks/useFormActions.ts
```diff
@@ -0,0 +1,124 @@
...
+  const getData = (values: Application, step: number): ApplicationData =>
+    snakecaseKeys(
+      {
+        ...values,
+        applicationStep: getApplicationStepString(step),
+      },
+      { deep: true }
+    );
+
+  const handleNext = (currentValues: Application): void => {
```
nenonja, line 104: If you would start to use Wizard.tsx from the shared folder, you could also use useWizard-hook which gives these step functions "for free". Also some other boilerplate could be removed, like indexing on steps.

PR #332, opened 2021-10-05
KS-Backend: Fix 502 error in production

klemmari1:
## Description :sparkles:

With some long callback URLs, uwsgi will throw 502 errors with the message:

`[WARNING] unable to add HTTP_X_FORWARDED_HOST=... to uwsgi packet, consider increasing buffer size.`

Increase the buffer size to (try to) resolve this.


haavikko (approved):
OK!

PR #334, opened 2021-10-06
KS-Backend: Refine attachments & logout

klemmari1:
## Description :sparkles:

Small adjustments to improve the attachment handling. For example we won't check if the file exists with os.path.isfile since we are using the blob storage.

Use the django logout view instead of the Azure ADFS logout to log the user out from the excel download view. This is because if using the Azure logout, the user will be logged out of all sites related to the ADFS login.


haavikko (approved):
Looks good to me!

PR #342, opened 2021-10-08
Backend: Fix refreshing tokens

klemmari1:
## Description :sparkles:

Fix the refreshing of helsinki profile access token. No need to create a new OIDCProfile instance but instead need to update the existing instance.


haavikko (approved):
Looks good to me!

PR #346, opened 2021-10-11
Handle case where DUMMY_COMPANY_FORM is set but empty

haavikko:
## Description :sparkles:
If the variable is listed in the configmap but doesn't have a value


klemmari1 (approved):
LGTM!

PR #348, opened 2021-10-11
HL-209 & HL-200|  Validation for Association Fields & Add the Required Checkbox

Nizar-Rahme:
## Description :sparkles:
- Fix validation and styles for association fields
- Add the missing checkbox 
- Fix styels
 
## Screenshots :camera_flash:
image


frontend/benefit/applicant/src/components/applications/forms/application/step1/utils/validation.ts
```diff
@@ -58,16 +61,35 @@ export const getValidationSchema = (t: TFunction): Yup.SchemaOf<Step1> =>
...
     [APPLICATION_FIELDS_STEP1_KEYS.ASSOCIATION_HAS_BUSINESS_ACTIVITIES]:
       Yup.boolean()
         .nullable()
-        .when(APPLICATION_FIELDS_STEP1_KEYS.ORGANIZATION_TYPE, {
-          is: ORGANIZATION_TYPES.ASSOCIATION,
-          then: Yup.boolean().required(t(VALIDATION_MESSAGE_KEYS.REQUIRED)),
+        .test({
+          message: t(VALIDATION_MESSAGE_KEYS.REQUIRED),
+          test: (val) => {
+            if (
```
nenonja, line 70: kinda hard to follow the logic. Does this mean that there is for example radiobutton input for `ASSOCIATION_HAS_BUSINESS_ACTIVITIES` but only if organization type is `ASSOCIATION` ? And in this case it is required to select true or false value for the radiobutton? With other organization types there is no validation (perhaps no input either) and it always return true. 
yup can be quite complex sometimes...

frontend/benefit/applicant/src/components/applications/forms/application/step1/utils/validation.ts
```diff
@@ -58,16 +61,35 @@ export const getValidationSchema = (t: TFunction): Yup.SchemaOf<Step1> =>
...
+      Yup.boolean()
+        .nullable()
+        .test({
+          message: t(VALIDATION_MESSAGE_KEYS.REQUIRED),
+          test: (val) => {
+            if (
+              organizationType?.toLowerCase() ===
+              ORGANIZATION_TYPES.ASSOCIATION.toLowerCase()
+            )
+              return val === true;
```
nenonja, line 89: and same for here but input seem to be a required checkbox which must be selected. Otherwise value is invalid.

frontend/benefit/applicant/src/components/applications/forms/application/step1/utils/validation.ts
Nizar-Rahme: Exactly how you described it 👌

PR #353, opened 2021-10-12
YJDH-153 | KS-Employer: Add Sentry logging

klemmari1:
## Description :sparkles:

- Add initial Sentry setup for the Employer UI.
- KS-Backend: Add support for a header used by the frontend sentry: "sentry-trace".

## Issues :bug:

[YJDH-153](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-153)

### DIFF
--- a/frontend/kesaseteli/employer/sentry.server.config.js
+++ b/frontend/kesaseteli/employer/sentry.server.config.js
@@ -1,0 +1,19 @@
+// This file configures the initialization of Sentry on the server.
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs';
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
+const SENTRY_ENVIRONMENT = process.env.SENTRY_ENVIRONMENT || process.env.NEXT_PUBLIC_SENTRY_ENVIRONMENT || "development";
+
+Sentry.init({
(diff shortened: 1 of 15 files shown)


haavikko (approved):
Looks good to me! Maybe if you have time, document something about SENTRY_AUTH_TOKEN configuration in different environments, like is it required and where to configure it

frontend/kesaseteli/employer/sentry.server.config.js
```diff
@@ -0,0 +1,19 @@
+// This file configures the initialization of Sentry on the server.
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs';
```
nenonja, line 5: Is this enough or should we wrap the server also with sentry? https://github.com/City-of-Helsinki/yjdh/blob/develop/frontend/shared/src/server/next-server.js#L27-L44

with this https://docs.sentry.io/platforms/javascript/guides/nextjs/:

```
import { withSentry } from "@sentry/nextjs";

const handler = async (req, res) => {
  res.status(200).json({ name: "John Doe" });
};

export default withSentry(handler);
```

nenonja (commented):
Quickly read through and I didnt test this in anyway... 
One thing to remember is to check on staging and production envs that the sentry works on them too.

frontend/kesaseteli/employer/sentry.server.config.js
nenonja: as we spoke, no need to do this

PR #369, opened 2021-10-15
Yjdh-224: Final translations

nenonja:
## Description :sparkles:
Let's set up the final translations of the application. Also there is some minor code changes to show new texts. Still wip


nenonja (commented):
Exceptionally no need to approve, just translations

PR #372, opened 2021-10-15
HL-228 | Remove the link from the Helsinki logo on the header

Nizar-Rahme:
## Description :sparkles:
- Remove the link from the Helsinki logo on the header

## Testing :alembic:
- The Helsinki logo on the header is no longer clickable


nenonja (approved):
Just wonder why do you want to drop the link? For The code it LGTM.

PR #383, opened 2021-10-19
KS-Backend: Handle YTJ api request timeout

klemmari1:
## Description :sparkles:

Do not raise an error for the user if YTJ API times out.


haavikko (approved):
Looks good to me! On question, is it always RequestException that is thrown, is there any need to catch also HTTPErrors?

PR #384, opened 2021-10-19
Backend: Change ADFS login username to email

klemmari1:
## Description :sparkles:

Use the `email` parameter for the adfs login username which should be unique.


haavikko (approved):
OK great, good that you fixed ot for Helsinki-lisä too :-)

PR #385, opened 2021-10-19
KS-Backend: Add invoicer fields to excel

klemmari1:
## Description :sparkles:

Add invoicer fields to excel if application has a separate invoicer.


haavikko (approved):
Looks good to me!

PR #386, opened 2021-10-19
YJDH-212: Serial number / invoicer fields bug fix

nenonja:
## Description :sparkles:

- Fix serial number and invoicer input saving bugs
- refactor business logic to separate hook
- useToggle hook inspired by this: https://tkdodo.eu/blog/use-state-vs-use-reducer


Nizar-Rahme (approved):
LGTM!

PR #387, opened 2021-10-20
Limit phone number field length

quyenlq:
## Description :sparkles:
Added length limit and validation to the phone number field
Technically the `PhoneNumberField` would be enough to validate phone number length since any number which is less than 6 or more than 13 in length is already an invalid phone number and not possible to save to the DB anyway. But I just added this for making this consistent with FE validation


haavikko (approved):
Looks good to me!

PR #392, opened 2021-10-20
KS-Employer: NextJS Sentry setup fix

klemmari1:
## Description :sparkles:

Try to fix the Sentry setup so that we would get messages from dev/test/stg/prod environments.


frontend/kesaseteli/employer/src/pages/_app.tsx
```diff
@@ -22,17 +22,9 @@ import { ThemeProvider } from 'styled-components';
 
 import { BackendEndPoints, getBackendDomain } from '../backend-api/backend-api';
 
-// Sentry logger
```
nenonja, line 25: I wonder if we just still keep the react-query logger here, or didnt it work?
https://react-query.tanstack.com/reference/setLogger

PR #418, opened 2021-10-27
HL-206: calculator API

haavikko:
## Description :sparkles:

Do you guys think we should have a separate endpoint for the handlers?
Currently the endpoint is the same, but the "calculation", "batch" and "pay_subsidies" fields in application are not available for applicants.

## Issues :bug:
HL-206

## Testing :alembic:
backend/benefit/calculator/tests/test_calculator_api.py
backend/benefit/calculator/tests/test_previous_benefits_api.py


backend/benefit/common/delay_call.py
```diff
@@ -0,0 +1,56 @@
...
+def _call_all_pending():
+    if _local.pending_calls is None:
+        raise Exception("This function must be used only with update_at_end")
+    for k, func in list(_local.pending_calls.items()):
+        func()
+    _local.pending_calls.clear()
+
+
+@contextlib.contextmanager
+def do_delayed_calls_at_end():
```
quyenlq, line 37: Nice function!

quyenlq (approved):
Great work! 💯

PR #425, opened 2021-10-28
Yjdh-277: Improved Error Summary

nenonja:
## Description :sparkles:
Improved error summary:
- use hds ErrorSummary which has autofocus
- links from error message to input field
- also on step 1
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-277
## Screenshots :camera_flash:
Screenshot 2021-10-28 at 3 40 33 PM
Screenshot 2021-10-28 at 12 18 31 PM


melniiv (approved):
Looks good. But we have the same in bf. Should we style those links somehow? With default styling it looks weird

nenonja:
> Looks good. But we have the same in bf. Should we style those links somehow? With default styling it looks weird

Yes we def could do that. This is the kesäseteli component https://github.com/City-of-Helsinki/yjdh/pull/425/files#diff-f01411c41f4264a94ee6b1c631c6aaf929d2e598a5af6f1e0a6cf6f848b0dbcd
but it (or some part of it) could be moved to shared folder with styles

PR #435, opened 2021-11-01
YJDH-296 | Backend: Security improvements

klemmari1:
## Description :sparkles:

Improvements based on the security audit:

- Set the secure flag on the CSRF cookie.
- Limit the info returned by the `/oidc/userinfo` endpoint to given name and family name.

## Issues :bug:

[YJDH-296](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-296)

The security audit also mentioned (minor issue) that the httpOnly flag should be set on the CSRF cookie. However this complicates the communication from frontend to backend so much that here I think the benefits of not setting it outweigh the risks.

See more info from:

https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-CSRF_COOKIE_HTTPONLY

> Although the setting offers little practical benefit, it’s sometimes required by security auditors.

> If you enable this and need to send the value of the CSRF token with an AJAX request, your JavaScript must pull the value from a hidden CSRF token form input instead of from the cookie.


haavikko (approved):
Looks good to me!

PR #439, opened 2021-11-03
YJDH-293 | KS-Employer: Add modify datepicker validation

klemmari1:
## Description :sparkles:

Change the onBlur event to onChange to validate the datepicker value correctly even when the date is picked with the datepicker element instead of writing it.

## Issues :bug:

[YJDH-293](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-293)


nenonja:
Seem to work! I just noted that for example "20" or "2000" is ok for the ui. I think there is something wrong with the regex

nenonja:
> Seem to work! I just noted that for example "20" or "2000" is ok for the ui. I think there is something wrong with the regex

Let's do this in separate pr

PR #440, opened 2021-11-04
HL-279 | Add message model

quyenlq:
## Description :sparkles:
Added message model for messaging function and note function


backend/benefit/messages/tests/factories.py
```diff
@@ -0,0 +1,14 @@
+import factory
+from messages.models import Message
+
+
+class MessageFactory(factory.django.DjangoModelFactory):
+    content = factory.Faker("sentence", nb_words=32)
+    sender = factory.SubFactory("applications.tests.factories.UserFactory")
+    application = factory.SubFactory("applications.tests.factories.ApplicationFactory")
```
haavikko, line 8: If the application has messages the application would not be in draft status anymore. PR https://github.com/City-of-Helsinki/yjdh/pull/443 has HandlingApplicationFactory thta might be used here, it's not merged yet though

haavikko (approved):
Looks good to me!

backend/benefit/messages/tests/factories.py
quyenlq: Thanks, fixed

PR #441, opened 2021-11-04
HL-307: Fix refetch in get queries

melniiv:
## Description :sparkles:
Fixes the issue with unsaved changes cleared when switching tabs

### DIFF
--- a/frontend/benefit/applicant/src/hooks/useApplicationQuery.ts
+++ b/frontend/benefit/applicant/src/hooks/useApplicationQuery.ts
@@ -18,9 +18,7 @@ const useApplicationQuery = (
           ),
     {
       enabled: Boolean(id),
-      retry: false,
-      refetchOnWindowFocus: false,
-      refetchOnReconnect: false,
+      staleTime: Infinity,
     }
   );
 };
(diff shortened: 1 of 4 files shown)


frontend/benefit/applicant/src/hooks/useApplicationQuery.ts
```diff
@@ -16,7 +16,12 @@ const useApplicationQuery = (
         : handleResponse<ApplicationData>(
             axios.get(`${BackendEndpoint.APPLICATIONS}${id}/`)
           ),
-    { enabled: Boolean(id), retry: false }
+    {
```
nenonja, line 19: Not sure if the disabling is the best option, refetching has good things also. In react-hook-form I used reset-function to keep the data on form: 
https://github.com/City-of-Helsinki/yjdh/blob/03d1819671203094d8ad4d0432c9f981b76ea8fc/frontend/kesaseteli/employer/src/components/application/ApplicationForm.tsx#L21
and 
https://github.com/City-of-Helsinki/yjdh/blob/develop/frontend/kesaseteli/employer/src/hooks/application/useResetApplicationFormValues.ts
Not sure what's equivalent on formik, is it this one: https://formik.org/docs/api/formik#resetform-nextstate-partialformikstatevalues--void

nenonja (commented):
If the comment below doesnt work, then you need to disable refetching, but then it's better to disable on global level instead of repeating in every query:
https://react-query.tanstack.com/guides/window-focus-refetching
(put all the options in the same way to there)

nenonja (approved):
Inifinty is there but does it work now?
I cannot test this because the review environment is broken (login does not work). You should probably ask backend developers to do same kind of mocked login as kesäseteli has, to make the review environment workable. I could set this up locally but it takes a lot of time and usually it still wont deploy (new settings missing etc). Review environment would be favorable. I approve it now as the code looks good now.

PR #447, opened 2021-11-06
HL-282 | Message API

quyenlq:

### DIFF
--- a/backend/benefit/helsinkibenefit/urls.py
+++ b/backend/benefit/helsinkibenefit/urls.py
@@ -12,7 +12,12 @@ from drf_spectacular.views import (
     SpectacularRedocView,
     SpectacularSwaggerView,
 )
-from rest_framework import routers
+from messages.views import (
+    ApplicantMessageViewSet,
+    HandlerMessageViewSet,
+    HandlerNoteViewSet,
+)
+from rest_framework_nested import routers
 from terms.api.v1.views import ApproveTermsOfServiceView
 from users.api.v1.views import CurrentUserView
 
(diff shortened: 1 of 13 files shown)


backend/benefit/helsinkibenefit/urls.py
```diff
@@ -12,7 +12,12 @@
     SpectacularRedocView,
     SpectacularSwaggerView,
 )
-from rest_framework import routers
+from messages.views import (
+    ApplicantMessageViewSet,
+    HandlerMessageViewSet,
+    HandlerNoteViewSet,
+)
+from rest_framework_nested import routers
```
haavikko, line 20: drf-nested-routers seems suitable for our purposes

backend/benefit/messages/views.py
```diff
@@ -0,0 +1,53 @@
...
+from messages.serializers import MessageSerializer, NoteSerializer
+from rest_framework import viewsets
+from rest_framework.exceptions import NotFound
+from users.utils import get_company_from_request
+
+
+class ApplicantMessageViewSet(viewsets.ModelViewSet):
+
+    serializer_class = MessageSerializer
+    permission_classes = [BFIsAuthenticated, TermsOfServiceAccepted, IsMessageAuthor]
```
haavikko, line 16: Since the applicant is operating on behalf of their company, should they see all of their company's messages?

haavikko (commented):
I'll continue review later today, kids woke up :-)

backend/benefit/messages/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+
+    class Meta:
+        model = Message
+        fields = [
+            "id",
+            "created_at",
+            "modified_at",
+            "content",
+            "message_type",
```
haavikko, line 22: Is it so that also the handler UI doesn't need the "sender" field?

backend/benefit/messages/views.py
```diff
@@ -0,0 +1,53 @@
...
+        if settings.DISABLE_AUTHENTICATION:
+            return Application.objects.get(
+                id=self.kwargs["application_pk"]
+            ).messages.get_messages_qs()
+        company = get_company_from_request(self.request)
+        try:
+            application = company.applications.get(id=self.kwargs["application_pk"])
+        except Application.DoesNotExist:
+            raise NotFound(_("Application not found"))
+        if company == application.company:
```
haavikko, line 28: Wouldn't check would always be True (as we are querying company.applications) so code can be simplified?

backend/benefit/messages/permissions.py
```diff
@@ -0,0 +1,18 @@
...
+    message = _("You don't have permission to change this message")
+
+    def has_object_permission(self, request, view, obj):
+        if (
+            settings.DISABLE_AUTHENTICATION
+            or request.method in permissions.SAFE_METHODS
+        ):
+            return True
+        if request.user.is_handler():
+            return obj.sender == request.user
```
haavikko, line 17: Maybe we should confirm with Joonas if this restriction applies? Should the handlers be able to delete/edit each other's messages?

backend/benefit/messages/permissions.py
```diff
@@ -0,0 +1,18 @@
+from django.conf import settings
+from django.utils.translation import gettext_lazy as _
+from rest_framework import permissions
+from users.utils import get_company_from_request
+
+
+class IsMessageAuthor(permissions.BasePermission):
```
haavikko, line 7: Maybe reconsider the name of the class. For applicants, we're not checking for authorship, but company

haavikko (approved):
Looks good! Removed my previous comments as they are actually addressed, but added a few minor ones

backend/benefit/messages/serializers.py
quyenlq: At least from the design I don't see the place where sender is needed. However, the NoteSerializer has `sender`

backend/benefit/messages/views.py
quyenlq: Nice catch, removed this part

backend/benefit/messages/permissions.py
quyenlq: Make sense, I changed the name to `HasMessagePermission`

backend/benefit/messages/permissions.py
quyenlq: So Jonathan confirm that we do not allow update, only delete is allowed. I update this part to follow the specification. Relevant unittests are also updated

PR #448, opened 2021-11-08
Fix issue with do_delayed_calls_at_end when using multiple threads

haavikko:
## Description :sparkles:

update_now_or_later results in error when multiple threads are used, as threadlocal variable is not initialized


quyenlq (approved):
LGTM!

PR #462, opened 2021-11-09
Bump package versions 09.11.2021

nenonja:


frontend/benefit/applicant/package.json
```diff
@@ -34,9 +34,9 @@
     "graphql": "^15.5.0",
     "hds-react": "^1.6.0",
     "lodash": "^4.17.21",
-    "next": "^12.0.2",
```
nenonja, line 37: next 12 isnt compatible with styled-components yet: https://nextjs.org/blog/next-12 
> The new Rust compiler is backwards compatible. If you have an existing Babel configuration, you will automatically be opted out. We have plans to port parsing for popular libraries like styled-components, emotion, and relay soon. If you're using a custom Babel setup, please share your configuration.

PR #464, opened 2021-11-10
YJDH-319: better validation error messages

nenonja:
## Description :sparkles:
Unify error messages in error notificaton box to match error messages next to field.


Nizar-Rahme (approved):
LGTM! 👍

PR #468, opened 2021-11-12
YJDH-222: Browser test for application sending

nenonja:
## Description :sparkles:
Browser test that creates the whole application with two employments with several attachment files and send it.
Bunch of bugs fixed along the way, at least:
- serial number selector bug (regression bug)
- summary page didnt show all data (regression bug)

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-222


Nizar-Rahme (approved):
Well done 👍

PR #469, opened 2021-11-13
HL-324: Add handler types, update lists

melniiv:
## Description :sparkles:
-Extend handler application type with calculation
-Show handlers name in the applications list in the handlers front page

### DIFF
--- a/frontend/benefit/handler/src/types/application.d.ts
+++ b/frontend/benefit/handler/src/types/application.d.ts
@@ -7,7 +7,7 @@ import {
   EMPLOYEE_KEYS,
 } from '../constants';
 
-export interface EmployeeData {
+export type EmployeeData = {
   id?: string;
   first_name: string;
   last_name: string;
@@ -25,7 +25,7 @@ export interface EmployeeData {
   commission_amount: string;
   commission_description: string;
   created_at?: string;
-}
+};
 
 export type Employee = {
   id?: string;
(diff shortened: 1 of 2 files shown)


frontend/benefit/handler/src/types/application.d.ts
```diff
@@ -154,6 +154,39 @@ export type ApproveTerms = {
   selectedApplicantConsents: string[];
 };
 
+// handler
+
+export interface RowData {
```
nenonja, line 159: These seems to be identical to other application type except for the casing, perhaps you could try this: https://stackoverflow.com/questions/60269936/typescript-convert-generic-object-from-snake-to-camel-case
Also I would use `type` instead of `interface` I think that `interface` is better only if supporting object oriented way of doing things.

nenonja (commented):
One comment

frontend/benefit/handler/src/types/application.d.ts
melniiv: great idea! but i would leave it for separate PR, when we move the common types to shared from applicant and handler. But i ll fix interface to type in this one

melniiv: fixed

nenonja: Ok, it's should not be a big thing to do as typescript would complain if it doesnt work. But let's leave it later.

nenonja (approved):

PR #470, opened 2021-11-15
HL-265: application review layout

melniiv:
## Description :sparkles:
Implements the layout and styling for the application review page:
-header
-review section
-review section with actions

Example:

image

### DIFF
--- a/frontend/benefit/handler/src/pages/application/index.tsx
+++ b/frontend/benefit/handler/src/pages/application/index.tsx
@@ -1,0 +1,32 @@
+import ApplicationReview from 'benefit/handler/components/applicationReview/ApplicationReview';
+import AppContext from 'benefit/handler/context/AppContext';
+import { GetStaticProps, NextPage } from 'next';
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations';
+import * as React from 'react';
+import { useEffect } from 'react';
+import theme from 'shared/styles/theme';
+
+const ApplicantIndex: NextPage = () => {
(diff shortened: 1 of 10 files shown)


frontend/benefit/handler/src/pages/application/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { GetStaticProps, NextPage } from 'next';
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations';
+import * as React from 'react';
+import { useEffect } from 'react';
+import theme from 'shared/styles/theme';
+
+const ApplicantIndex: NextPage = () => {
+  const { setIsNavigationVisible, setLayoutBackgroundColor } =
+    React.useContext(AppContext);
+
```
nenonja, line 12: Perhaps this could be cleaner code?
```
const setPageSettings = React.useCallback(() => {
    setIsNavigationVisible(false);
    setLayoutBackgroundColor(theme.colors.white);
}, [setIsNavigationVisible, setLayoutBackgroundColor]);

useEffect(() => {
    setPageSettings();
    return () => setPageSettings();
}, [setPageSettings]);
```

nenonja (approved):
One minor comment, but also approved. This seems to be first version with dummy data.

frontend/benefit/handler/src/pages/application/index.tsx
melniiv: lets see how many configurations we have. maybe we can do the pageConfigs hook or something. its a first try

PR #482, opened 2021-11-19
HL-225: Fix summary pay subsidy section logic

melniiv:
## Description :sparkles:
image
bug fixed: if apprenticeship program is No, the %es were not shown at all in the summary page.


Nizar-Rahme (approved):
LGTM!

PR #499, opened 2021-11-24
Fix company API when disable authentication

quyenlq:
## Description :sparkles:
There was a bug that when DISABLE_AUTHENTICATION set to `True`. Company API won't be able to extract the business_id from the user request. So it always returns 404

Small fix to change the logic so there is 1 company always be returned if authentication disabled


haavikko (approved):
Looks good to me!

PR #502, opened 2021-11-25
Add missing migrations

quyenlq:
## Description :sparkles:
Some migrations were missing when changing the models.
These do not affect the DB but will be generated by django when running `makemigrations`


haavikko (approved):
Looks good to me! Just noticed the same myself

PR #503, opened 2021-11-25
HL-328 + HL-329 | Add read receipt to Message API

quyenlq:
## Description :sparkles:
- Added seen_by_handler & seen_by_applicant into message model, since we don't need to know exactly who read the message
- Added `unread_messages_count` to application serializer. This will count all the unread message according to user persective. For example: applicant will see count of message which is unread by applicant, while handler see count of message which they haven't seen.
- The new field is annotated during the `get_queryset` method for better performance.
- Whenever someone query the message view per application, all messages seen status will be changed, depends on who read the message


backend/benefit/applications/api/v1/serializers.py
```diff
@@ -693,6 +695,9 @@ class Meta:
         allow_blank=True,
         help_text="Company contact person phone number normalized (start with zero, without country code)",
     )
+    unread_messages_count = serializers.IntegerField(
```
haavikko, line 698: Good, should be helpful for UI devs

backend/benefit/messages/views.py
```diff
@@ -19,6 +20,11 @@ class ApplicantMessageViewSet(viewsets.ModelViewSet):
         HasMessagePermission,
     ]
 
+    @transaction.atomic
```
haavikko, line 23: good idea to have atomic block here

backend/benefit/applications/api/v1/views.py
```diff
@@ -162,14 +164,23 @@ class ApplicantApplicationViewSet(BaseApplicationViewSet):
     permission_classes = [BFIsAuthenticated, TermsOfServiceAccepted]
     filterset_class = ApplicantApplicationFilter
 
+    def _annotate_unread_messages_count(self, qs):
```
haavikko, line 167: good idea to use annotation

haavikko (approved):
Looks good to me!

PR #506, opened 2021-11-26
kesaseteli/employer: Add let's encrypt staging -certificates + fix browser tests

nenonja:
- Add certificate hashes from review environment to the browser-test script flag
- Fix production build bug that loads still the previous draft application after clicking "create new application"

### DIFF
--- a/frontend/kesaseteli/employer/package.json
+++ b/frontend/kesaseteli/employer/package.json
@@ -13,8 +13,8 @@
     "test:debug-dom": "cross-env DEBUG_PRINT_LIMIT=1000000 yarn test",
     "test:staged": "yarn test --watchAll=false --findRelatedTests",
     "test:coverage": "yarn test:debug-dom --verbose --coverage",
-    "browser-test": "testcafe \"chrome --window-size='1249,720'\" testcafe.dotenv.js browser-tests/",
-    "browser-test:ci": "testcafe \"chrome:headless --disable-gpu --window-size='1249,720'\" --screenshots path=report,fullPage=true --video report --reporter spec,custom,html:report/index.html testcafe.dotenv.js browser-tests/"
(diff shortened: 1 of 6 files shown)


mikkovihonen:
I think we've added the cert injection to wrong place and it also seems that the file system doesn't allow this.  
Let's take a step back and see if this would work 
https://source.chromium.org/chromium/chromium/src/+/master:services/network/public/cpp/network_switches.cc;l=35?q=kIgnoreCertificateErrorsSPKIList&ss=chromium

nenonja:
Everything is green! This hasn't happened for a month or so! Let's keep this now forever 🎉

frontend/kesaseteli/employer/package.json
```diff
@@ -13,8 +13,8 @@
     "test:debug-dom": "cross-env DEBUG_PRINT_LIMIT=1000000 yarn test",
     "test:staged": "yarn test --watchAll=false --findRelatedTests",
     "test:coverage": "yarn test:debug-dom --verbose --coverage",
-    "browser-test": "testcafe \"chrome --window-size='1249,720'\" testcafe.dotenv.js browser-tests/",
-    "browser-test:ci": "testcafe \"chrome:headless --disable-gpu --window-size='1249,720'\" --screenshots path=report,fullPage=true --video report --reporter spec,custom,html:report/index.html testcafe.dotenv.js browser-tests/"
+    "browser-test": "testcafe 'chrome --window-size=\"1249,720\" --ignore-certificate-errors-spki-list=\"8sg/cl7YabrOFqSqH+Bu0e+P27Av33gWgi8Lq28DW1I=,gJt+wt/T3afCRkxtMMSjXcl/99sgzWc2kk1c1PC9tG0=,zrQI2/1q8i2SRPmMZ1sMntIkG+lMW0legPFokDo3nrY=\"' testcafe.dotenv.js browser-tests/",
```
nenonja, line 16: review environment certificate hashes are added explicitly to make the browser test ignore specific certificate errors.

Nizar-Rahme (approved):

PR #509, opened 2021-11-30
HL-319: apprenticeship benefit calculation, TrainingCompensation model and API

haavikko:
## Description :sparkles:

Calculate apprenticeship benefit

## Issues :bug:
HL-319

## Testing :alembic:
backend/benefit/calculator/tests/Helsinki-lisä laskurin testitapaukset.xlsx


backend/benefit/calculator/rules.py
```diff
@@ -157,70 +170,77 @@ def get_maximum_monthly_pay_subsidy(self):
...
-                self._create_row(
-                    SalaryBenefitSubTotalRow, start_date=start_date, end_date=end_date
-                )
 
+        if len(date_ranges) > 1:
             self._create_row(
                 DateRangeDescriptionRow,
-                start_date=start_date,
-                end_date=end_date,
+                start_date=sub_range.start_date,
```
quyenlq, line 237: This looks quite weird because `sub_range` is a local variable inside the for-loop. Is this supposed to be correct? Or should it be `date_ranges[0].start_date` and `date_ranges[-1].end_date`?

haavikko: The goal is that, if the the calculation is done separately for multiple date ranges, to add for each time range an
"Ajalta dd.mm.yyyy - dd.mm.yyyy" row. In the UI design, page "palkan - kahdella ajalla", you can see the two separate rows

image

from:
https://futuricedesigners.invisionapp.com/console/share/T734X5SPR2/813764061

But if there is only one time range in the calculation, then the row "Ajalta dd.mm.yyyy - dd.mm.yyyy" is redundant, that's why the row is created only if len(date_ranges) > 1

haavikko: Oops, my previous comment was wrong, there's indeed a bug in the calculation of the "Koko ajalta" row, fixed

quyenlq (approved):
image

PR #514, opened 2021-12-01
YJDH-335 | Add /v1/schools/ endpoint with hardcoded data

karisal-anders:
## Description :sparkles:

Add /v1/schools/ endpoint with hardcoded data 

Add tests for the /v1/schools/ endpoint:
 - test_schools_list_status_ok
 - test_schools_list_returns_list
 - test_schools_list_returns_non_empty
 - test_schools_list_returns_active_school_count
 - test_schools_list_returns_string_collection
 - test_schools_list_returns_non_empty_strings_collection
 - test_schools_list_returns_unique_elements_collection
 - test_schools_list_returns_sorted_collection

NOTE: The hardcoded school values in /v1/schools/ are placeholder data.

Refs YJDH-335

## Issues :bug:

YJDH-335


backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -21,14 +23,33 @@
...
+    _name_fi = Func(
+        "name",
+        function="fi-FI-x-icu",  # fi_FI.UTF8 would be best but wasn't available
+        template='(%(expressions)s) COLLATE "%(function)s"',
+    )
+
+    queryset = School.objects.all().order_by(_name_fi.asc())
+    serializer_class = SchoolSerializer
+
+    def get_permissions(self):
```
quyenlq, line 48: Can you use class attribute permission_classes here? Should be much shorter
```
class SchoolListView(ListAPIView):
    # PostgreSQL specific functionality:
    # - Custom sorter for name field to ensure finnish language sorting order.
    # - NOTE: This can be removed if the database is made to use collation fi_FI.UTF8
    _name_fi = Func(
        "name",
        function="fi-FI-x-icu",  # fi_FI.UTF8 would be best but wasn't available
        template='(%(expressions)s) COLLATE "%(function)s"',
    )

    queryset = School.objects.all().order_by(_name_fi.asc())
    serializer_class = SchoolSerializer
    permission_classes = [AllowAny]
```

backend/kesaseteli/applications/models.py
```diff
@@ -16,6 +16,18 @@
 from companies.models import Company
 
 
+class School(HistoricalModel, TimeStampedModel, UUIDModel):
```
quyenlq, line 19: Do we have any use-case for School Historical Model?

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -21,14 +23,33 @@
...
-from applications.models import Application, SummerVoucher, YouthApplication
+from applications.models import Application, School, SummerVoucher, YouthApplication
 from common.utils import DenyAll
 
 
+class SchoolListView(ListAPIView):
+    # PostgreSQL specific functionality:
+    # - Custom sorter for name field to ensure finnish language sorting order.
+    # - NOTE: This can be removed if the database is made to use collation fi_FI.UTF8
+    _name_fi = Func(
```
quyenlq, line 39: Nice one!

backend/kesaseteli/applications/tests/test_schools_api.py
```diff
@@ -0,0 +1,61 @@
...
+def test_schools_list_returns_non_empty_list_of_unique_elements(api_client):
+    response = api_client.get(get_schools_api_url())
+    assert response.status_code == status.HTTP_200_OK
+    assert type(response.json()) == list
+    assert len(response.json()) > 0
+    assert len(set(response.json())) == len(response.json())
+
+
+@pytest.mark.django_db
+def test_schools_list_returns_non_empty_sorted_list(api_client):
```
quyenlq, line 56: General comment about these tests. They look quite duplicated to me, seems like the next test is a superset of the previous test. For example, while nothing changes in the request, you don't have to assert `assert type(response.json()) == list` in the second test if the first test already did that. Each test should be isolated and test a specific expectation

backend/kesaseteli/applications/tests/test_schools_api.py
```diff
@@ -0,0 +1,61 @@
...
+    response = api_client.get(get_schools_api_url())
+    assert type(response.json()) == list
+    assert len(response.json()) > 0
+
+
+@pytest.mark.django_db
+def test_schools_list_returns_non_empty_string_list(api_client):
+    response = api_client.get(get_schools_api_url())
+    assert type(response.json()) == list
+    assert len(response.json()) > 0
```
quyenlq, line 32: How about counting the element then comparing it to `Schools.objects.count()`?

backend/kesaseteli/applications/tests/test_schools_api.py
karisal-anders: I'll make the tets more isolated.

backend/kesaseteli/applications/tests/test_schools_api.py
karisal-anders: Taking this into account.

backend/kesaseteli/applications/api/v1/views.py
karisal-anders: Thanks! Adapted from this [Stackoverflow](https://stackoverflow.com/a/36423340) answer.

backend/kesaseteli/applications/models.py
karisal-anders: Not really, probably inactive/active would be simpler. I'll look into this.

backend/kesaseteli/applications/api/v1/views.py
karisal-anders: I tried that initially and doublechecked it just now. Using the permission_classes = [AllowAny] gives 403 Forbidden when accessing https://localhost:8000/v1/schools/ so no or not only.

backend/kesaseteli/applications/models.py
karisal-anders: Taken into account by removing HistoricalModel and adding deleted_at member variable and using it for filtering for active schools only.

quyenlq (approved):

PR #515, opened 2021-12-01
HL-251 | Fix - De Minimis grant has an initial selected value

Nizar-Rahme:
## Description :sparkles:
- Fix mismatching application types
- Remove the unnecessary application template query
- Fix save logic on the first step for new applications
- Fix Attachment types for shared, Benefit, and Kesäseteli modules

## Testing :alembic:
- Create a new application
- The question about the De Minimis grant should have no pre-selected value

**Note:** There was a bug when saving **new** applications using the "save and quit" button **on the first step**, this also got fixed in the same task.


frontend/benefit/applicant/src/hooks/useFormActions.tsx
```diff
@@ -29,21 +29,19 @@ const useFormActions = (application: Application): FormActions => {
...
-    }
-  }, [newApplication?.id, router]);
+  const { mutateAsync: createApplication, error: createApplicationError } =
+    useCreateApplicationQuery();
+
+  const createApplicationAndAppendId = async (
+    data: ApplicationData
+  ): Promise<void> => {
+    const newApplication = await createApplication(data);
+    void router.replace({
```
nenonja, line 39: router replace causes page refresh, but you can prevent it by passing `{ shallow: true }`. I'm not sure though which is the use case here.

nenonja (changes requested):
One minor comment... otherwise refactoring with attachments look good. Still kesaseteli component tests and deploying has failed, you should address them. New Browser test coverage some part of the attachments but does not try to add wrong type of file for example so it needs some manual testing for kesaseteli (I Can do it when deploy works).

frontend/benefit/applicant/src/hooks/useFormActions.tsx
Nizar-Rahme: Good point 👍 
Fixed

Nizar-Rahme:
> One minor comment... otherwise refactoring with attachments look good. Still kesaseteli component tests and deploying has failed, you should address them. New Browser test coverage some part of the attachments but does not try to add wrong type of file for example so it needs some manual testing for kesaseteli (I Can do it when deploy works).

I made TestCafe uses the same TS version that is used in the applications, which fixed the failure in building browser tests.

nenonja (approved):
It seems that everything is green now ✅

PR #516, opened 2021-12-02
HL-321 | Integrate YRTTI API

quyenlq:
## Description :sparkles:
YRTTI API Integration to query association info via Palveluväylä
Documentation of YRTTI API can be found: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/7562068053/Working+document+Integration+to+Palveluv+yl+from+app+developer+s+perspective

- Added a YRTTIClient to `shared` component, although AFAIK Kesaseteli does not support application for association.
- Update company query logic so it'll try to get association info if the first query to company return nothing.
- Update unittests

### DIFF
--- a/backend/benefit/helsinkibenefit/urls.py
+++ b/backend/benefit/helsinkibenefit/urls.py
@@ -59,9 +59,6 @@ urlpatterns = [
     path("v1/", include(handler_app_router.urls)),
     path("v1/terms/approve_terms_of_service/", ApproveTermsOfServiceView.as_view()),
     path("v1/company/", GetCompanyView.as_view()),
-    path(
-        "v1/company/<str:business_id>", GetCompanyView.as_view()
-    ),  # FIXME: Remove this later
     path("v1/users/me/", CurrentUserView.as_view()),
     path("oidc/", include("shared.oidc.urls")),
     path("oauth2/", include("shared.azure_adfs.urls")),
--- a/backend/shared/shared/yrtti/yrtti_client.py
+++ b/backend/shared/shared/yrtti/yrtti_client.py
@@ -1,0 +1,72 @@
+import requests
+from django.conf import settings
+
+TARGET_ASSOCIATION_NAME_TYPE = "P"
+TARGET_ASSOCIATION_NAME_LANGUAGE = "FI"
+TARGET_ASSOCIATION_NAME_STATUS = "R"
+
+
+class YRTTIClient:
+    def __init__(self):
+        if not all([settings.YRTTI_BASIC_INFO_PATH, settings.YRTTI_TIMEOUT]):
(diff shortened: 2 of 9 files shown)


backend/shared/shared/yrtti/yrtti_client.py
```diff
@@ -0,0 +1,69 @@
...
+    def _get_active_association_name(self, name_info: dict) -> str:
+        # There will be the case where an association has more than one name
+        # Here we only pick up latest name available
+        return [
+            name
+            for name in name_info
+            if name["AssociationNameType"] == TARGET_ASSOCIATION_NAME_TYPE
+            and name["AssociationNameLanguage"] == TARGET_ASSOCIATION_NAME_LANGUAGE
+            and name["AssociationNameStatus"] == TARGET_ASSOCIATION_NAME_STATUS
+        ][0]
```
haavikko, line 69: Is there always at least one name? Raise meaningful error if no name in data?

backend/shared/shared/yrtti/yrtti_client.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_association_data_from_yrtti_data(self, yrtti_data: dict) -> dict:
+        """
+        Get the required company fields from YRTTI data.
+        """
+        association_name = self._get_active_association_name(
+            yrtti_data["AssociationNameInfo"]
+        )
+        # There might be a list of addresses, but it is impossible to identify which one is the primary address from
+        # the data so we just select the first one. Most of the time there is only 1 address
```
haavikko, line 29: This solution is OK with me, good to have the comment as documentation

haavikko (approved):
Please check the comments and update code if you think it needs to, merge at will!

At the testing stage, we should test with more associations if at all possible

backend/benefit/helsinkibenefit/urls.py
```diff
@@ -59,9 +59,6 @@
     path("v1/", include(handler_app_router.urls)),
     path("v1/terms/approve_terms_of_service/", ApproveTermsOfServiceView.as_view()),
     path("v1/company/", GetCompanyView.as_view()),
-    path(
-        "v1/company/<str:business_id>", GetCompanyView.as_view()
-    ),  # FIXME: Remove this later
```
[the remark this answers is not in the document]
quyenlq: FE guys confirmed that this is not needed anymore

backend/shared/shared/yrtti/yrtti_client.py
quyenlq: Nice catch, from the API doc it is nullable but I don't think it'll ever happen. Updated this so there will be an error if name not found

PR #520, opened 2021-12-03
YJDH-347 | Make social security number check stricter

karisal-anders:
## Description :sparkles:

Add check that the Finnish social security number has no whitespace and
is uppercase. Affects YouthApplication's social security number.

Add more test cases in test_youth_applications_api.py for valid and
invalid social security numbers.

Refs YJDH-347

## Issues :bug:

YJDH-347

### DIFF
--- a/backend/kesaseteli/common/utils.py
+++ b/backend/kesaseteli/common/utils.py
@@ -16,21 +16,41 @@ class DenyAll(BasePermission):
         return False
 
 
+def has_whitespace(value):
+    value_without_whitespace = "".join(value.split())
+    return value != value_without_whitespace
+
+
+def is_uppercase(value):
+    """
+    Is the value all uppercase? Returns True also if there are no alphabetic characters.
+    """
+    return value == value.upper()
+
+
 def validate_finnish_social_security_number(value):
     """
-    Raise a ValidationError if the given value is not a Finnish social security number.
+    Raise a ValidationError if the given value is not an uppercase Finnish social
+    security number with no whitespace.
     """
-    if not is_valid_finnish_social_security_number(value):
+    if (
+        not is_valid_finnish_social_security_number(value)
+        or has_whitespace(value)
+        or not is_uppercase(value)
+    ):
         raise ValidationError(
-            _("%(value)s is not a valid Finnish social security number"),
+            _(
+                "%(value)s is not a valid Finnish social security number. "
+                "Make sure to have it in uppercase and without whitespace."
+            ),
             params={"value": value},
         )
 
 
 def validate_optional_finnish_social_security_number(value):
     """
-    Raise a ValidationError if the given value is not None, an empty string or a Finnish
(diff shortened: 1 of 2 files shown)


backend/kesaseteli/common/utils.py
```diff
@@ -16,21 +16,38 @@ def has_permission(self, request, view):
...
     """
-    Raise a ValidationError if the given value is not a Finnish social security number.
+    Raise a ValidationError if the given value is not an uppercase Finnish social
+    security number with no whitespace.
     """
-    if not is_valid_finnish_social_security_number(value):
+    if (
+        not is_valid_finnish_social_security_number(value)
+        or has_whitespace(value)
+        or not is_uppercase(value)
```
quyenlq, line 36: You can just use `not value.isupper()`

backend/kesaseteli/common/utils.py
```diff
@@ -16,21 +16,38 @@ def has_permission(self, request, view):
...
 def validate_finnish_social_security_number(value):
     """
-    Raise a ValidationError if the given value is not a Finnish social security number.
+    Raise a ValidationError if the given value is not an uppercase Finnish social
+    security number with no whitespace.
     """
-    if not is_valid_finnish_social_security_number(value):
+    if (
+        not is_valid_finnish_social_security_number(value)
+        or has_whitespace(value)
```
quyenlq, line 35: `or " " in value` should do the trick here

backend/kesaseteli/common/utils.py
karisal-anders: Thanks for the tip! Taking into account.

backend/kesaseteli/common/utils.py
karisal-anders: That's a bit different. I want to make sure there is no whitespace at all, not just space characters.

backend/kesaseteli/common/utils.py
karisal-anders: Actually it's not the same. "123".isupper() == False which is weird. I'm changing is back to the former way...

karisal-anders: I changed this back and added documentation to is_uppercase function that it's not same as string.isupper

karisal-anders:
**Now the checks passed.**

Check "(KS) Build & Review & Acceptance tests / KS Acceptance tests (ks-empl) (pull_request)" **failed** a at least **twice** but passed on a third or fourth run I think.

backend/kesaseteli/common/utils.py
quyenlq: ah okay. Got it 👍

quyenlq (approved):
LGTM! Thanks for clarification

karisal-anders:
> LGTM! Thanks for clarification

You're welcome and thanks for the review!

PR #525, opened 2021-12-08
YJDH-330: save youth application to backend

nenonja:
## Description :sparkles:
call backend to save the application to the backend. Including tests.


Nizar-Rahme (approved):
Looks good to me 👍

PR #531, opened 2021-12-09
Fix the attachment type on the Handler side

Nizar-Rahme:
## Description :sparkles:
- Fix the types on the attachments on the Handler side

**Note:** now only shared components have the type Attachment, but service-specific components have the corresponding type (e.g. BenefitAttachment).

### DIFF
--- a/frontend/benefit/handler/src/components/attachmentsListView/AttachmentsListView.tsx
+++ b/frontend/benefit/handler/src/components/attachmentsListView/AttachmentsListView.tsx
@@ -7,10 +7,10 @@ import {
 } from 'shared/components/benefit/summaryView/SummaryView.sc';
 import { $GridCell } from 'shared/components/forms/section/FormSection.sc';
 import theme from 'shared/styles/theme';
-import Attachment from 'shared/types/attachment';
+import { BenefitAttachment } from 'shared/types/attachment';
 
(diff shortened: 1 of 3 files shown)


nenonja:
changes to the handler folder does not trigger any linting, building or tests?

nenonja (commented):
Changes to the handler folder does not trigger any linting, building/deploying or tests?

Nizar-Rahme:
> Changes to the handler folder does not trigger any linting, building/deploying or tests?

Yeah, it doesn't seem to trigger them. I'll check how we can fix this.

nenonja:
> > Changes to the handler folder does not trigger any linting, building/deploying or tests?
> 
> Yeah, it doesn't seem to trigger them. I'll check how we can fix this.

I think that the workflow yml files are simply missing for handler.

Nizar-Rahme:
> I think that the workflow yml files are simply missing for handler.

It looks like that this is the case, but we'll add them in a separate PR.

nenonja (approved):

nenonja:
> > I think that the workflow yml files are simply missing for handler.
> 
> It looks like that this is the case, but we'll add them in a separate PR.

Good idea... I approved this

melniiv (approved):

PR #538, opened 2021-12-13
HL-335 and HL-364: months_used and months_remaining values, use PastBenefit info in validation

haavikko:
HL-335: add months_used and months_remaining to application API.
HL-364: use PreviousBenefit (in addition to previously accepted
applications) objects when validating application duration.

## Issues :bug:
HL-335 and HL-364

## Testing :alembic:
backend/benefit/applications/tests/test_benefit_aggregation.py


backend/benefit/applications/api/v1/serializers.py
```diff
@@ -897,97 +916,41 @@ def _validate_date_range(self, start_date, end_date, benefit_type):
...
+        if validation_errors := get_past_benefit_info(
+            self.instance,
+            company,
+            social_security_number,
+            start_date,
+            end_date,
+            apprenticeship_program,
+        ).validation_errors:
+            error_text = ", ".join(str(v) for v in validation_errors)
+            raise serializers.ValidationError({"start_date": error_text})
```
karisal-anders, line 953: Is the validation error here always related specifically to start_date or can it be related to something else?

backend/benefit/applications/benefit_aggregation.py
```diff
@@ -0,0 +1,168 @@
...
+        past_benefit_info.months_remaining = None
+    else:
+        past_benefit_info.months_remaining = max(
+            Decimal(0), BENEFIT_MAX_DURATION - past_benefit_info.months_used
+        )
+
+    if (
+        not apprenticeship_program
+        and past_benefit_info.months_used + duration_in_months(start_date, end_date)
+        > Application.BENEFIT_MAX_MONTHS
```
karisal-anders, line 75: What's the difference between BENEFIT_MAX_DURATION and BENEFIT_MAX_MONTHS? They're both 12 at the moment.

The code is ok as it is here and couldn't the if-clause also be done like this:
`
if(not apprenticeship_program and duration_in_months(start_date, end_date) > past_benefit_info.months_remaining)
`

backend/benefit/applications/benefit_aggregation.py
```diff
@@ -0,0 +1,168 @@
...
+            _("Benefit can not be granted before 24-month waiting period expires")
+        )
+    return past_benefit_info
+
+
+def _get_past_benefits(application, company, social_security_number, end_date):
+    """
+    Return a list containing two types of objects:
+    * PreviousBenefits
+    * Applications that are in ACCEPTED status
```
karisal-anders, line 88: Typing doesn't seem to be used much in the project but if you wanted
to add that based on the comments then shouldn't the return type be:
`List[Union[PreviousBenefit, Application]]`?

It's up to you whether you want to add it.

backend/benefit/applications/benefit_aggregation.py
```diff
@@ -0,0 +1,168 @@
...
+
+def _get_benefits_relevant_for_validation(past_benefits, start_date):
+    """
+    :param past_benefits: list of PreviousBenefit and/or Application objects
+    :param start_date:
+    :return:
+    """
+    if not past_benefits:
+        return []
+    most_recent_benefit = past_benefits[0]
```
karisal-anders, line 128: This relies on the past_benefits to be in a certain order. This could perhaps be mentioned in the function's documentation that the past_benefits parameter must have its benefits in a certain order or the code could be changed to find the most recent benefit even if the list wouldn't have the most recent benefit as its first element.

backend/benefit/applications/benefit_aggregation.py
```diff
@@ -0,0 +1,168 @@
...
+        > Application.BENEFIT_MAX_MONTHS
+    ):
+        # granting this benefit would exceed the limit
+        past_benefit_info.validation_errors.append(
+            _("Benefit can not be granted before 24-month waiting period expires")
+        )
+    return past_benefit_info
+
+
+def _get_past_benefits(application, company, social_security_number, end_date):
```
karisal-anders, line 84: Documenting that this function returns its values in a certain order might be good for using it later where the values are already presumed to be in a certain order (e.g. in _get_benefits_relevant_for_validation).

backend/benefit/applications/benefit_aggregation.py
```diff
@@ -0,0 +1,168 @@
...
+    :param start_date:
+    :return:
+    """
+    if not past_benefits:
+        return []
+    most_recent_benefit = past_benefits[0]
+    if (
+        not most_recent_benefit
+        or most_recent_benefit.end_date
+        < start_date - relativedelta(months=BENEFIT_WAITING_PERIOD_MONTHS)
```
karisal-anders, line 132: I have a hard time reading this without being confused as I read the line
`or most_recent_benefit.end_date`
on its own.

Perhaps adding parentheses or doing something else
stylistically might help to make this more readable,
or putting the `start_date - ...` part into a variable and
then using that in the comparison to make it fit on one
line might help.

I leave it up to you to decide what to do here, if anything.

backend/benefit/applications/tests/test_benefit_aggregation.py
```diff
@@ -0,0 +1,363 @@
...
+    for class_name, previous_start_date, previous_end_date in previous_benefits:
+        # previous, already granted benefits for the same employee+company
+        if class_name == Application:
+            decided_application = DecidedApplicationFactory()
+            # the type of the previous benefit and the apprenticeship_program value do not matter when
+            # calculating the remaining available benefit time (Teams discussion)
+            decided_application.benefit_type = random.choice(
+                [BenefitType.SALARY_BENEFIT, BenefitType.EMPLOYMENT_BENEFIT]
+            )
+            decided_application.apprenticeship_program = random.choice([True, False])
```
karisal-anders, line 314: Does this produce reproducible tests in the review environment?
I can see that in backend/benefit/common/tests/conftest.py there's
```
@pytest.fixture(autouse=True)
def setup_test_environment(settings):
    factory.random.reseed_random("777")
    with freeze_time("2021-06-04"):
        yield
```
which makes the tests at least under backend/benefit/common/tests reproducible
but it doesn't affect the tests in this test's directory, right?

It'd probably be better to make the random choices always be the same random
choices when the tests are run so that the tests don't randomly fail and succeed
if some values happen to be corner cases. What do you think?

backend/benefit/calculator/tests/factories.py
```diff
@@ -136,7 +136,7 @@ class PreviousBenefitFactory(factory.django.DjangoModelFactory):
         "pydecimal", left_digits=3, right_digits=2, min_value=0
     )
     total_amount = factory.LazyAttribute(
-        lambda o: duration_in_months(o.start_date, o.end_date)
+        lambda o: duration_in_months(o.start_date, o.end_date) * o.monthly_amount
```
karisal-anders, line 139: Has the meaning of the total_amount field changed? If it has then ok, adding the multiplication by monthly_amount seems logical. If it has not then was the multiplication left out previously by accident?

backend/benefit/applications/api/v1/serializers.py
haavikko: Rest framework stores errors under some key. It's possible to use a generic error also, but that loses all context of the error. In the frontend, the error message structure is at the moment flattened AFAIK, so the key doesn't have effect on the display of the error.

backend/benefit/applications/benefit_aggregation.py
haavikko: good idea

backend/benefit/applications/benefit_aggregation.py
haavikko: good idea

backend/benefit/applications/benefit_aggregation.py
haavikko: Good point, need to fix

backend/benefit/applications/benefit_aggregation.py
haavikko: good idea

backend/benefit/calculator/tests/factories.py
haavikko: yes, it was a bug

karisal-anders: Ok, this is clear

backend/benefit/applications/tests/test_benefit_aggregation.py
haavikko: fixed

backend/benefit/applications/api/v1/serializers.py
haavikko: As discussed, in future this error will be turned into a warning, and the issue can be addressed at thata time.

backend/benefit/applications/benefit_aggregation.py
haavikko: fixed

backend/benefit/applications/benefit_aggregation.py
haavikko: fixed

backend/benefit/applications/benefit_aggregation.py
haavikko: fixed

backend/benefit/applications/benefit_aggregation.py
haavikko: fixed

backend/benefit/applications/benefit_aggregation.py
haavikko: fixed

backend/benefit/applications/api/v1/serializers.py
karisal-anders: > As discussed, in future this error will be turned into a warning, and the issue can be addressed at thata time.

👌

backend/benefit/applications/benefit_aggregation.py
karisal-anders: There are more places with similar readability issues but I don't see them as crucial, I leave them up to you whether you want to change them or not. This is ok as it is now for me.

backend/benefit/applications/tests/test_benefit_aggregation.py
karisal-anders: Sorry, I gave you false information. I just had a talk with @quyenlq in another PR on this same subject, see the following comments: https://github.com/City-of-Helsinki/yjdh/pull/545#discussion_r771393436, https://github.com/City-of-Helsinki/yjdh/pull/545#discussion_r771400958, https://github.com/City-of-Helsinki/yjdh/pull/545#discussion_r772105156.

Now it actually runs the `backend/benefit/common/tests/conftest.py`'s `setup_test_environment` for each parametrization of `test_application_with_previously_accepted_applications_and_previous_benefits` so
```python
decided_application.benefit_type = random.choice(
                [BenefitType.SALARY_BENEFIT, BenefitType.EMPLOYMENT_BENEFIT]
            )
```
will always give the same value.

I thought this would've generated the same random values for each parametrization of the function `test_application_with_previously_accepted_applications_and_previous_benefits` i.e.
 - parametrization 0 → random value X₀
 - parametrization 1 → random value X₁
 - …
 - parametrization N → random value Xₙ

So that X₀, X₁, …, Xₙ would be random but stay the same between different pytest test runs.
But it became apparent to me this morning that this is not so as I doublechecked my own code in another PR.

backend/benefit/applications/api/v1/serializers.py
```diff
@@ -716,6 +717,24 @@ def get_applicant_terms_in_effect(self, obj):
         else:
             return None
 
+    def get_past_benefit_info(self, obj):
+        aggregated_info = get_past_benefit_info(
```
quyenlq, line 721: function name is a bit confusing here 😅

backend/benefit/calculator/models.py
```diff
@@ -345,6 +345,10 @@ class PreviousBenefit(UUIDModel, TimeStampedModel):
 
     history = HistoricalRecords(table_name="bf_calculator_previousbenefit_history")
 
+    @property
```
quyenlq, line 348: This appears quite a lots in different models. Would it be better to create a mixin?

backend/benefit/common/tests/conftest.py
```diff
@@ -11,6 +13,7 @@
 @pytest.fixture(autouse=True)
 def setup_test_environment(settings):
     factory.random.reseed_random("777")
+    random.seed(777)
```
quyenlq, line 16: is it leftover? Or what is the purpose of it?

quyenlq (approved):
LGTM! Some minor comments, feel free to merge :)

backend/benefit/applications/api/v1/serializers.py
haavikko: :-D once you see it, you can't unsee it... I'll change it to something more unambiguous

backend/benefit/common/tests/conftest.py
haavikko: Had discussion about randomization of tests, yes this is leftover from that, removed

backend/benefit/calculator/models.py
haavikko: Good idea!

PR #544, opened 2021-12-13
HL-171 | Add support to multiple time ranges in ahjo template

quyenlq:
## Description :sparkles:
Add support to create multiple rows per application in AHJO PDF
This change only affects applications that have two separated benefit periods. In that case we render the sub total sums of the benefit instead of the total sum.
Reusing the `CalculationRow` requires me to add `start_time` and `end_time` to the Row model, otherwise, I'll need to recalculate them from `PaySubsidy` or `TrainingCompensation` every time.
## Screenshots :camera_flash:

Sample HTML file generated from the test, PDF file will look almost the same
Screenshot 2021-12-13 at 23 53 36


backend/benefit/applications/services/pdf_templates/benefit_with_de_minimis_aid.html
```diff
@@ -14,16 +14,32 @@
...
+            {% if app.ahjo_rows.count() < 2 %}
             <tr>
                 <td>{{ app.ahjo_application_number }}</td>
                 <td>{{ app.company_name }}</td>
                 <td>{{ app.company.business_id }}</td>
                 <td>{{ app.employee.first_name }} {{ app.employee.last_name }}</td>
                 <td>{{ app.start_date.strftime("%d.%m.%Y") }} - {{ app.end_date.strftime("%d.%m.%Y")}}</td>
                 <td>{{ app.get_benefit_type_display() }}</td>
-                <td class="text-right">{{ "%.0f €" | format(app.benefit_amount/(app.duration_in_months | round | int))}} </td>
+                <td class="text-right">{{ "%.0f €" | format(app.benefit_amount/(app.duration_in_months | round(2)))}} </td>
```
haavikko, line 25: Should we actually use app.calculation.duration_in_months, as that the handlers might grant the benefit for a shorter time range than the applicant asked for?

backend/benefit/calculator/models.py
```diff
@@ -426,6 +426,8 @@ def __init__(self, *args, **kwargs):
     amount = models.DecimalField(
         max_digits=7, decimal_places=2, verbose_name=_("row amount")
     )
+    start_date = models.DateField(blank=True, null=True, verbose_name=_("Start date"))
+    end_date = models.DateField(blank=True, null=True, verbose_name=_("End date"))
```
haavikko, line 430: +1 for adding start_date and end_date in the CalculationRow. I considered it myself but there wasn't a strong enough use case for it at the time, but for the Ahjo handling it's good to have it here.

haavikko (approved):
One comment about duration_in_months, please check that, but merge at will!

backend/benefit/applications/services/pdf_templates/benefit_with_de_minimis_aid.html
quyenlq: Yes, nice catch. `app.duration_in_months` was used because previously we didn't have a separated calculation model. Fixed now

PR #547, opened 2021-12-14
Frontend validation: Change max length of youth names

nenonja:

### DIFF
--- a/frontend/kesaseteli/youth/src/__tests__/index.test.tsx
+++ b/frontend/kesaseteli/youth/src/__tests__/index.test.tsx
@@ -61,8 +61,8 @@ describe('frontend/kesaseteli/youth/src/pages/index.tsx', () => {
       const indexPageApi = getIndexPageApi();
       await indexPageApi.expectations.pageIsLoaded();
 
-      indexPageApi.actions.typeInput('first_name', 'a'.repeat(257)); // max limit is 257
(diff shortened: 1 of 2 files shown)


melniiv (approved):

frontend/kesaseteli/youth/src/__tests__/index.test.tsx
```diff
@@ -61,8 +61,8 @@ describe('frontend/kesaseteli/youth/src/pages/index.tsx', () => {
       const indexPageApi = getIndexPageApi();
       await indexPageApi.expectations.pageIsLoaded();
 
-      indexPageApi.actions.typeInput('first_name', 'a'.repeat(257)); // max limit is 257
-      indexPageApi.actions.typeInput('last_name', 'a'.repeat(257)); // max limit is 257
+      indexPageApi.actions.typeInput('first_name', 'a'.repeat(129)); // max limit is 128
+      indexPageApi.actions.typeInput('last_name', 'a'.repeat(129)); // max limit is 128
```
karisal-anders, line 65: Just out of curiosity why is there one more here than in the maxLength in the YouthForm.tsx? I mean 129 instead of 128?

You changed the comment to say 128 but use repeat(129), was that intentional? Previously the comment used the same value as in the repeat...

PR #550, opened 2021-12-15
Add yjdh-tet initial local environment bootstrapping and review setup

mattikl:


.github/workflows/te-review.yml
```diff
@@ -0,0 +1,280 @@
...
+            dockerfile: ./frontend/tet/employer/Dockerfile
+            port: 3000
+          - service: te-yout
+            context: ./frontend
+            dockerfile: ./frontend/tet/youth/Dockerfile
+            port: 3000
+          - service: te-admn
+            context: ./frontend
+            dockerfile: ./frontend/tet/admin/Dockerfile
+            port: 3000
```
tuovinensanttu, line 108: `database: false/true` are missing from these. See [kesäseteli](https://github.com/City-of-Helsinki/yjdh/blob/develop/.github/workflows/ks-review.yml#L88:L103).

frontend/tet/admin/.eslintrc.js
```diff
@@ -0,0 +1,57 @@
...
+        specialLink: ['hrefLeft', 'hrefRight'],
+        aspects: ['invalidHref', 'preferButton'],
+      },
+    ],
+    'import/no-extraneous-dependencies': [
+      'error',
+      // Use package.json from both this package folder and root.
+      { packageDir: [__dirname, join(__dirname, '../../')] },
+    ],
+    'no-void': 'off',
```
nenonja, line 23: there were some rules recently added to the other projects as new major version of eslint was taken into use.

frontend/tet/youth/src/pages/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const Home: NextPage = () => (
+  <div className={styles.container}>
+    <Head>
+      <title>TET Job Search - Admin</title>
+      <meta name="description" content="Generated by create next app" />
+      <link rel="icon" href="/favicon.ico" />
+    </Head>
+
+    <main className={styles.main}>
```
nenonja, line 15: This is probably hello world type of thing which can be like this, but for the future is good to know that at least in other existing projects we are using StyledComponents instead of css modules.

.github/workflows/te-review.yml
```diff
@@ -0,0 +1,284 @@
...
+            dockerfile: ./backend/docker/tet.Dockerfile
+            port: 8000
+          - service: te-empl
+            context: ./frontend
+            dockerfile: ./frontend/tet/employer/Dockerfile
+            port: 3000
+          - service: te-yout
+            context: ./frontend
+            dockerfile: ./frontend/tet/youth/Dockerfile
+            port: 3000
```
nenonja, line 65: ports should be different. in docker-compose you have 3000, 3001, and 3002. You could use same here and other places too.

docker-compose.tet.yml
```diff
@@ -0,0 +1,95 @@
...
+       dockerfile: tet/employer/Dockerfile
+       target: development
+     volumes:
+       - ./frontend:/app
+       - /app/node_modules
+       - /app/.next
+     env_file:
+       - .env.tet
+     container_name: te-empl
+
```
nenonja, line 46: add
```
    environment:
      - PORT=3000
```
You can check how the ports are  mapped in docker-compose.employer

frontend/tet/youth/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@frontend/te-yout",
+  "version": "1.0.0",
+  "scripts": {
+    "dev": "PORT=3000 node start.js",
```
nenonja, line 5: check that the port is same here as in docker-compose and te-review.yml file

frontend/tet/youth/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "build": "next build",
+    "start": "NODE_ENV=production PORT=3000 node start.js ",
```
nenonja, line 7: also here

frontend/tet/employer/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@frontend/te-empl",
+  "version": "1.0.0",
+  "scripts": {
+    "dev": "PORT=3000 node start.js",
```
nenonja, line 5: check that the port is same here as in docker-compose and te-review.yml file

frontend/tet/admin/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@frontend/te-admn",
+  "version": "1.0.0",
+  "scripts": {
+    "dev": "PORT=3000 node start.js",
```
nenonja, line 5: check that the port is same here as in docker-compose and te-review.yml file

nenonja (commented):
Some comments left... check at least that each service has different localhost ports and is correctly set in package scripts, te-review.yml and docker-compose file (perhaps somewhere else too)

nenonja:
Also there is some lint failures because develop just upgraded to the latest version of eslint

.github/workflows/te-review.yml
mattikl: Our reasoning was to use always port 3000 inside the frontend containers (least surprising) and then map to different ports outside the containers as needed. But we can of course change this so that all YJDH projects implement this in the same way.

frontend/tet/admin/.eslintrc.js
mattikl: Thanks, I will review the rules under tet apps!

PR #552, opened 2021-12-16
HL-350: fix suomi.fi logout. Add logout_callback url.

haavikko:
## Description :sparkles:

Implement logout using redirect to city profile, instead of a background request.

Also silence Sonarcube warning about healthcheck URLs


karisal-anders:
SonarCloud Code Analysis is failing because of code duplication but the code duplication is only between helsinkibenefit and kesaseteli's settings.py files and that's fine by me.

backend/shared/shared/oidc/views/hki_views.py
```diff
@@ -57,38 +56,47 @@ def login_failure(self):
...
-            response = requests.post(
-                settings.OIDC_OP_LOGOUT_ENDPOINT,
-                data=payload,
-                verify=self.get_settings("OIDC_VERIFY_SSL", True),
-                timeout=self.get_settings("OIDC_TIMEOUT", None),
-                proxies=self.get_settings("OIDC_PROXY", None),
+            query = urlencode(params)
+
+            redirect_url = "{url}?{query}".format(
+                url=settings.OIDC_OP_LOGOUT_ENDPOINT, query=query
```
karisal-anders, line 80: Would using urllib.parse.urljoin or some other URL handling function here be safer than using format?

backend/shared/shared/oidc/views/hki_views.py
```diff
@@ -57,38 +56,47 @@ def login_failure(self):
...
+    """This callback is called after the suomi.fi logout has been performed at the city profile"""
 
-        return HttpResponse("OK", status=200)
+    http_method_names = ["get"]
+
+    def get(self, request):
+        # As of 2021-12, the city profile does not provide any error/status codes along with the
+        # callback. If such parameters are added in the future, we would handle them here.
+        # Now we just assume that the logout has been done successfully and redirect to
+        # the logout landing URL in the frontend.
```
karisal-anders, line 98: Nice clarifying comment! 👍

backend/shared/shared/oidc/views/hki_views.py
haavikko: fixed

karisal-anders:
`The KS Acceptance tests (ks-empl)` are failing, see https://github.com/City-of-Helsinki/yjdh/runs/4579175889?check_suite_focus=true

@nenonja What's your take on the above failing tests?

quyenlq (approved):
LGTM!

nenonja:
ks-empl logout browser test is red?

nenonja (changes requested):
Browser test gives correct result, the Logout is broken in ks-empl.
Start browser with this command (for macosx):
`
open -a Google\ Chrome --args --ignore-certificate-errors-spki-list="8sg/cl7YabrOFqSqH+Bu0e+P27Av33gWgi8Lq28DW1I=,gJt+wt/T3afCRkxtMMSjXcl/99sgzWc2kk1c1PC9tG0=,zrQI2/1q8i2SRPmMZ1sMntIkG+lMW0legPFokDo3nrY="
`

Then go to https://ks-empl-552.test.kuva.hel.ninja/fi
And try to login and then logout. It gives cors-error.

If you compare to this older environment, there the logout still works:
https://ks-empl-540.test.kuva.hel.ninja/login

frontend/kesaseteli/employer/src/components/header/__tests__/Header.test.tsx
```diff
@@ -37,9 +37,6 @@ describe('frontend/kesaseteli/employer/src/components/header/Header.tsx', () =>
     await waitFor(() =>
       expect(queryClient.getQueryData('user')).toBeUndefined()
     );
-    expect(spyRouterPush).toHaveBeenCalledWith(
```
nenonja, line 40: check that it redirects now to backend

frontend/kesaseteli/employer/src/components/header/__tests__/Header.test.tsx
```diff
@@ -37,9 +37,6 @@ describe('frontend/kesaseteli/employer/src/components/header/Header.tsx', () =>
     await waitFor(() =>
```
nenonja, line 37: this seems to be false positive check, because query data is no longer cleared, it should have been failed. Something is wrong with the test.

frontend/kesaseteli/employer/src/components/header/Header.tsx
```diff
@@ -36,17 +35,11 @@ const Header: React.FC = () => {
...
   const userQuery = useUserQuery();
-  const logoutQuery = useLogoutQuery();
-  const logout = React.useCallback(
-    () => logoutQuery.mutate({}, { onError }),
-    [logoutQuery, onError]
-  );
+  const logout = useLogout();
 
-  const isLoading = userQuery.isLoading || logoutQuery.isLoading;
+  const { isLoading, isSuccess, data } = userQuery;
```
nenonja, line 42: you should not destruct the react-query object because type-narrowing is lost: https://tkdodo.eu/blog/react-query-and-type-script#type-narrowing
If eslint complains, just use `eslint-disable-next-line` for the rule

frontend/kesaseteli/employer/src/components/header/Header.tsx
```diff
@@ -59,12 +52,12 @@ const Header: React.FC = () => {
...
         !isLoading
           ? {
-              isAuthenticated: !isLoginPage && userQuery.isSuccess,
+              isAuthenticated: !isLoginPage && isSuccess,
               loginLabel: t('common:header.loginLabel'),
               logoutLabel: t('common:header.logoutLabel'),
               onLogin: login,
               onLogout: logout,
-              userName: userQuery.isSuccess ? userQuery.data.name : undefined,
+              userName: isSuccess ? data?.name : undefined,
```
nenonja, line 60: here you can see that type narrowing didnt work and you have to use question mark `data?` even when data should be not empty.

frontend/kesaseteli/employer/src/components/header/Header.tsx
```diff
@@ -36,17 +35,11 @@ const Header: React.FC = () => {
...
-  const onError = useErrorHandler(false);
-
   const login = useLogin();
   const userQuery = useUserQuery();
-  const logoutQuery = useLogoutQuery();
-  const logout = React.useCallback(
-    () => logoutQuery.mutate({}, { onError }),
-    [logoutQuery, onError]
-  );
+  const logout = useLogout();
```
nenonja, line 40: The problem is that the useLogoutQuery cleans cache after success and now it does not do it:
If you check the `useLogoutQuery.tsx` there is success callback:
```
      onSuccess: () => {
        void queryClient.removeQueries();
        clearLocalStorage('application');
        void goToPage('/login?logout=true');
      },
```
goToPage call is no longer needed but we still need to clear local storage and remove react-query cache after succesfull logout. Perhaps you can do it on the `login.tsx` page on `useEffect`when `logout` value is true which means that we have just logged out succesfully.

nenonja (changes requested):
One important comment and few smaller ones.
Generally `onSuccess` callback is no longer called and the functionality should be moved to login.tsx page. I think every service has different `onSuccess` methods in `useLogoutQuery.ts` After the functionality is moved, we can remove all `useLogoutQuery.ts` files as they are no longer used.

nenonja (approved):
Looks good to me and I saw that localstorage was now cleared on logout. I think that Tet-team should review and test the logout on their service too: @mattikl @pekkvil

.github/workflows/ks-review.yml
```diff
@@ -144,6 +144,7 @@ jobs:
           echo "K8S_SECRET_CSRF_TRUSTED_ORIGINS=.test.kuva.hel.ninja" >> $GITHUB_ENV
           echo "K8S_SECRET_LOGIN_REDIRECT_URL=${{ env.EMPLOYER_URL }}" >> $GITHUB_ENV
           echo "K8S_SECRET_LOGIN_REDIRECT_URL_FAILURE=${{ env.EMPLOYER_URL }}/login?error=true" >> $GITHUB_ENV
+          echo "K8S_SECRET_LOGOUT_REDIRECT_URL=${{ env.EMPLOYER_URL }}/login?logout=true" >> $GITHUB_ENV
```
nenonja, line 147: Do we need these changes also to `te-review.yml` and `bf-review.yml` ?

.env.kesaseteli.example
```diff
@@ -15,6 +15,7 @@ OIDC_RP_CLIENT_ID=
 OIDC_RP_CLIENT_SECRET=
 LOGIN_REDIRECT_URL=https://localhost:3000/
 LOGIN_REDIRECT_URL_FAILURE=https://localhost:3000/login?error=true
+LOGOUT_REDIRECT_URL=https://localhost:3000/login?logout=true
```
nenonja, line 18: add this to `.env.tet.example` too?

nenonja (changes requested):
Actuall I just noticed two (small) comments more...

.env.kesaseteli.example
ngocdh236: Oh I forgot to remove that, apparently only backend needs that variable, oh and we don't have example files for backend?

.github/workflows/ks-review.yml
ngocdh236: @haavikko is this needed? if yes I can fix it now, thank you

.env.kesaseteli.example
ngocdh236: Oh they are in same file, I will fix now

.github/workflows/ks-review.yml
haavikko: Good catch, I think it's needed so thta login works in review environments

mattikl (approved):
TET settings were missing the settings that were added to benefit and kesaseteli. I added them and verified that TET admin is working correct. It only uses `oidc` auth when `NEXT_PUBLIC_MOCK_FLAG` is set. However, TET employer will use oidc authentication, so getting this working in TET project is important.

.github/workflows/te-review.yml
```diff
@@ -146,6 +146,7 @@ jobs:
           echo "K8S_SECRET_CSRF_TRUSTED_ORIGINS=.test.kuva.hel.ninja" >> $GITHUB_ENV
           echo "K8S_SECRET_AD_LOGIN_REDIRECT_URL=${{ env.ADMIN_URL }}" >> $GITHUB_ENV
           echo "K8S_SECRET_AD_LOGIN_REDIRECT_URL_FAILURE=${{ env.ADMIN_URL }}/login?error=true" >> $GITHUB_ENV
+          echo "K8S_SECRET_LOGOUT_REDIRECT_URL=${{ env.EMPLOYER_URL }}/login?logout=true" >> $GITHUB_ENV
```
nenonja, line 149: ADMIN_URL

.github/workflows/ks-review.yml
```diff
@@ -189,6 +190,7 @@ jobs:
           SERVICE_PORT: ${{ matrix.port }}
           K8S_SECRET_ALLOWED_HOSTS: "*"
           APP_MIGRATE_COMMAND: ${{ matrix.database == true && '/app/.prod/on_deploy.sh' || ''}}
+          LOGOUT_REDIRECT_URL: ${{ env.EMPLOYER_URL }}/login?logout=true
```
nenonja, line 193: This one is still missing from in bf-review and te-review. te-review its `ADMIN_URL`

.github/workflows/bf-review.yml
```diff
@@ -170,6 +171,7 @@ jobs:
           SERVICE_PORT: ${{ matrix.port }}
           K8S_SECRET_ALLOWED_HOSTS: "*"
           APP_MIGRATE_COMMAND: ${{ matrix.database == 'true' && '/app/.prod/on_deploy.sh' || ''}}
+          LOGOUT_REDIRECT_URL: ${{ env.EMPLOYER_URL }}/login?logout=true
```
nenonja: env.APPLICANT_URL

nenonja (approved):
Looks good now

PR #568, opened 2021-12-21
Fix platta browser tests with real integrations

nenonja:
## Description :sparkles:
Browser tests on #yjdh-alerts havent been working for a long time. request hook didnt work on suomi fi profile, now it is fixed.
Also test can manage with draft application if previous run has failed and left a draft behind.

### DIFF
--- a/frontend/kesaseteli/employer/browser-tests/application-page/step2.components.ts
+++ b/frontend/kesaseteli/employer/browser-tests/application-page/step2.components.ts
@@ -30,29 +30,79 @@ export const getStep2Components = (t: TestController) => {
     });
   const withinForm = (): ReturnType<typeof within> => within(formSelector());
 
+  const accordionSelector = (isOpen = true, index = 0) =>
+    screen.findByTestId(`accordion-${index}-${isOpen ? 'open' : 'closed'}`);
+  const withinAccordion = (
+    isOpen = true,
+    index = 0
+  ): ReturnType<typeof within> => within(accordionSelector(isOpen, index));
+
+  const form = async () => {
+    const selectors = {
+      title() {
+        return formSelector();
+      },
+    };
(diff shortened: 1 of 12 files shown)


frontend/kesaseteli/employer/browser-tests/application-page/step2.components.ts
```diff
@@ -30,124 +30,179 @@ export const getStep2Components = (t: TestController) => {
...
+        }
+        /* eslint-enable no-await-in-loop */
+      },
+    };
+    await expectations.isPresent();
+    return {
+      expectations,
+      actions,
+    };
+  };
```
Nizar-Rahme, line 86: Although these functions seem to do the job, they might be error-prone for tests.

Nizar-Rahme (approved):
Looks good!

frontend/kesaseteli/employer/browser-tests/application-page/step2.components.ts
nenonja: It was only way I was able to work with it. There might be draft applications from previous runs which must be removed before starting to fill a new one.. each application has remove button so there might be 0...n remove buttons on page. Each one should be clicked until there is no more remove buttons on page. Also SelectorPromise is not a list, only way to access nth button is to use `removeButtons.nth(0)`. Also every click affects to dom, so you have to re-valuate the selector after each remove... I mean that after `t.click(removeButtons.nth(0))` is done, then `t.click(removeButtons.nth(1))` would not work after it but we have to re-evaluate the `removeButtons` selector in between and slways take the `nth(0)` from the list. 
Anyway it seem to work very well. I tried to cut the previous run in different situations, also manually added 8 applications, and also tried taht there is zero draft applications. In each case it worked correctly.

PR #569, opened 2021-12-22
YJDH-340 | KS-Backend: Add YouthApplication.encrypted_vtj_json

karisal-anders:
## Description :sparkles:

Add YouthApplication.encrypted_vtj_json as EncryptedCharField field
using validate_optional_json validator.

Remove HistoricalModel from YouthApplication as unnecessary.
 - This removes the duplication of encrypted_vtj_json for each change in
   a YouthApplication instance and as such a possible source of database
   growth.

Add validators:
  - validate_json
  - validate_optional_json
    - Otherwise like validate_json but also accepts None and ""

Add tests:
 - backend/kesaseteli/applications/tests/test_application_validation.py:
   - test_validate_youth_application_vtj_json
 - backend/shared/shared/common/tests/test_validators.py:
   - test_validate_json_with_valid_input
   - test_validate_json_with_invalid_input
   - test_validate_optional_json_with_valid_input
   - test_validate_optional_json_with_invalid_input

## Issues :bug:

YJDH-340


karisal-anders:
The browser test for filling the youth application failing has been reported in ticket [YJDH-374](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-374).

backend/kesaseteli/applications/models.py
```diff
@@ -100,6 +104,13 @@ class YouthApplication(HistoricalModel, TimeStampedModel, UUIDModel):
     receipt_confirmed_at = models.DateTimeField(
         null=True, blank=True, verbose_name=_("timestamp of receipt confirmation")
     )
+    encrypted_vtj_json = EncryptedCharField(
+        null=True,
+        blank=True,
+        max_length=1024 * 1024 * 256,
```
haavikko, line 110: Can it really be this large?

haavikko (approved):
Looks good to me! The length limit in the encrypted_vtj_json seems quite large, if you feel it could be shorter then I suggest changing it, but merge at will

backend/kesaseteli/applications/models.py
karisal-anders: > Looks good to me! The length limit in the encrypted_vtj_json seems quite large, if you feel it could be shorter then I suggest changing it, but merge at will

I lowered the max_length of the encrypted_vtj_json from 256MiB to 1MiB.

BTW YouthApplication model is historized using HistoricalModel which uses HistoricalRecords from django-simple-history and looking at its [documentation](https://django-simple-history.readthedocs.io/en/latest/quick_start.html#what-is-django-simple-history-doing-behind-the-scenes) it says—here Choice and Poll are models whose changes are historized:
> Every time a create, update, or delete occurs on Choice or Poll a new row is created in the historical table for that model including all of the fields in the instance of the base model

So, having the encrypted_vtj_json duplicated every time when any of the fields in YouthApplication change will make the database blow up quite easily if the field is large. I won't address this in this PR but I will make a ticket about this.

karisal-anders:
> Looks good to me! The length limit in the encrypted_vtj_json seems quite large, if you feel it could be shorter then I suggest changing it, but merge at will

I lowered the max_length of the encrypted_vtj_json from 256MiB to 1MiB.

BTW YouthApplication model is historized using HistoricalModel which uses HistoricalRecords from django-simple-history and looking at its [documentation](https://django-simple-history.readthedocs.io/en/latest/quick_start.html#what-is-django-simple-history-doing-behind-the-scenes) it says—here Choice and Poll are models whose changes are historized:
> Every time a create, update, or delete occurs on Choice or Poll a new row is created in the historical table for that model including all of the fields in the instance of the base model

So, having the encrypted_vtj_json duplicated every time when any of the fields in YouthApplication change will make the database blow up quite easily if the field is large. I won't address this in this PR but I will make a ticket about this.

backend/kesaseteli/applications/models.py
karisal-anders: Most likely not, see https://github.com/City-of-Helsinki/yjdh/pull/569#issuecomment-1011997948

karisal-anders:
> So, having the encrypted_vtj_json duplicated every time when any of the fields in YouthApplication change will make the database blow up quite easily if the field is large. I won't address this in this PR but I will make a ticket about this.

**I didn't make a ticket** for this but **removed HistoricalModel from YouthApplication** and **force pushed the changes**.

haavikko (approved):
LGTM!

PR #570, opened 2021-12-22
Disable build and review env deploy of empty frontends

mattikl:
The review environment is having issues with excessive load
described in YJDH-370. The temporary solution is to
disable frontends that are not yet under development.


nenonja (commented):
Looks good! Let's just wait until the jobs are done and green before merging, otherwise deploys will hang forever.

nenonja (approved):

PR #574, opened 2021-12-23
Application step 2 fix

melniiv:
## Description :sparkles:
The conditional comission radio broke the flow, so its now removed and will be added when needed


Nizar-Rahme (approved):
Approving, though I am not quite sure why hiding it behind a feature flag won't do the trick.

melniiv:
yes, it needs further investigation

quyenlq (approved):

PR #575, opened 2021-12-23
HL-272: Additional info to handling

melniiv:
## Description :sparkles:
Implements the functionality to the applicant to make additional changes to the application and resubmit it to HANDLING status.

### DIFF
--- a/frontend/benefit/applicant/src/components/applications/forms/application/step5/useApplicationFormStep5.ts
+++ b/frontend/benefit/applicant/src/components/applications/forms/application/step5/useApplicationFormStep5.ts
@@ -15,6 +15,7 @@ import { useContext, useEffect } from 'react';
 import hdsToast from 'shared/components/toast/Toast';
 import { getFullName } from 'shared/utils/application.utils';
 import snakecaseKeys from 'snakecase-keys';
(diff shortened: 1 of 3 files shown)


frontend/benefit/applicant/src/components/applications/forms/application/step5/useApplicationFormStep5.ts
```diff
@@ -96,7 +97,12 @@ const useApplicationFormStep5 = (
 
   const handleSubmit = (): void => {
     const submitFields = isSubmit
-      ? { status: APPLICATION_STATUSES.RECEIVED }
```
nenonja, line 99: It's getting a bit complicated with inner resolving logic, I perhaps would call `resolveStatus` or some other method here, or use `React.useMemo` to resolve the status value, outside the handleSubmit method.

nenonja (approved):
One minor comment but approved also

frontend/benefit/applicant/src/components/applications/forms/application/step5/useApplicationFormStep5.ts
melniiv: its not reusable, only local logic, so separate method is a bit strange at least

PR #583, opened 2022-01-03
YJDH-374 | KS-Backend: Make activation email sending more lenient

karisal-anders:
## Description :sparkles:

**This makes the youth browser tests pass again:**
 - The problem was that send_activation_email in the backend
   failed and made the backend return an internal server error.

### Test posting youth application with email backends 

An activation email is sent when posting a youth application.
Add test_youth_application_post_valid_data_with_email_backends which
tests that using different email backends when posting a youth
application works.
 
### KS-Backend: Send emails to console only if MOCK_FLAG is true:

Using MOCK_FLAG rather than DEBUG for choosing which email backend to
use seems more appropriate here.

## Issues :bug:

YJDH-374

### DIFF
--- a/backend/kesaseteli/applications/models.py
+++ b/backend/kesaseteli/applications/models.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 
 from django.conf import settings
@@ -20,6 +21,8 @@ from applications.enums import (
 )
 from common.utils import validate_finnish_social_security_number
 from companies.models import Company
+
+LOGGER = logging.getLogger(__name__)
 
 
 def validate_school(name) -> None:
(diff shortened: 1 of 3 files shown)


backend/kesaseteli/applications/models.py
```diff
@@ -125,7 +125,7 @@ def send_activation_email(self, request):
             message=self._activation_email_message(request),
             from_email=settings.DEFAULT_FROM_EMAIL,
             recipient_list=[self.email],
-            fail_silently=False,
+            fail_silently=True,
```
haavikko, line 128: I think it would make sense to check the return value when fail_silently is used, and log a message/warning if it is 0.
"The return value will be the number of successfully delivered messages (which can be 0 or 1 since it can only send one message)."

haavikko (approved):
See message about fail_silently, after that merge at will!

backend/kesaseteli/applications/models.py
karisal-anders: Taken into account by logging an error if the return value from send_mail is zero.

PR #589, opened 2022-01-11
HL-385: validate start_date when submitting a draft application

haavikko:
## Description :sparkles:

start_date on a previous year prevented saving a draft application.

## Testing :alembic:
backend/benefit/applications/tests/test_applications_api.py: test_application_date_range_on_submit

### DIFF
--- a/backend/benefit/applications/tests/test_applications_api.py
+++ b/backend/benefit/applications/tests/test_applications_api.py
@@ -690,6 +690,11 @@ def test_application_delete(api_client, application):
         ("2021-02-01", "2021-02-28", 200),  # exactly one month
         ("2021-02-01", "2022-01-31", 200),  # exactly 12 months
         ("2021-02-01", "2022-02-01", 400),  # too long
+        (
+            "2020-02-01",
+            "2020-12-31",
+            200,
+        ),  # past year is allowed, as the application is not submitted
     ],
 )
 def test_application_date_range(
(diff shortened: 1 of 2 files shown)


backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -716,6 +721,40 @@ def test_application_date_range(
         assert response.data["end_date"] == end_date
 
 
+@pytest.mark.parametrize(
```
karisal-anders, line 724: Maybe freezing the time for this test would be good, e.g.
`@freeze_time("2021-01-01")`?

karisal-anders (commented):
Looks ok to me except that `test_application_date_range_on_submit` probably needs time freezing and please could you doublecheck whether `test_application_date_range` needs it? Thank you!

backend/benefit/applications/tests/test_applications_api.py
haavikko: freeze_time is applied to all these tests automatically (from common.tests.conftest)

karisal-anders (approved):

backend/benefit/applications/tests/test_applications_api.py
karisal-anders: Ah, ok so it was `from common.tests.conftest import *` in `backend/benefit/applications/tests/test_applications_api.py` which imported the autouse fixture `setup_test_environment` from `common.tests.conftest` which made the freeze_time be applied to these tests. Alright!

PR #616, opened 2022-01-12
Package bumps 01-12-2022

nenonja:


haavikko (approved):
Looks good to me! Btw, is the process of upgrading dependencies documented / should it? Are you using npm-check-updates or something else?

nenonja:
> Looks good to me! Btw, is the process of upgrading dependencies documented / should it? Are you using npm-check-updates or something else?

I just go through the dependabot pull requests and do the same updates in one pull-request. The problem with dependabot is that the it does not have sufficient rights to start deploying (for some reason) and run browser tests. So only way to test that deploy still works is to manually collect all the changed packages to single pr. This time the deploy didnt work there either because of infra problems? 😞  Still I think the chance for bug is small, if build and all the unit / component tests have passed.

PR #579, opened 2021-12-30
YJDH-362 | Refactor summer voucher applications

karisal-anders:
## Description :sparkles:

**Refactor** summer voucher and summer voucher application related classes **for clarity**:
 1. KS-Backend: Rename Application to EmployerApplication
 2. KS-Backend: Use employer_applications related_name
 3. KS-Backend: Rename SummerVoucher to EmployerSummerVoucher
 4. KS-Backend: Fix comment in get_test_phone_number
 5. KS: Prefix employer related endpoints with employer
 6. Fix styles (i.e. ran black and isort)
 7. KS-Backend: Prefix employer related serializers with Employer
 8. KS-Backend: Add employer to employer related permission names

**Rationale:**

Before in the Kesäseteli (KS) there was only Application which semantically meant EmployerApplication but was not named as such. Then YouthApplication was added and there were Application and YouthApplication. This was less than clear. Clarifying the code by renaming Application to its semantic meaning: EmployerApplication.

There was a similar situation—as in the previous paragraph—with SummerVoucher and YouthSummerVoucher. Clarifying this situation by renaming SummerVoucher to its semantic meaning: EmployerSummerVoucher.

Also doing similar renamings for other classes/endpoints that are employer specific to make that aspect explicit.

## Issues :bug:

YJDH-362

PR #618, opened 2022-01-12
YJDH-404 | KS-Backend: Fix migration conflicts

karisal-anders:
## Description :sparkles:

This migration conflict was caused by merging PR #579 into develop.

Fixes YJDH-404

## Issues :bug:

YJDH-404


haavikko (approved):
LGTM!

karisal-anders:
After discussing and debugging with @nenonja the failing [tests](https://github.com/City-of-Helsinki/yjdh/runs/4799016058?check_suite_focus=true) it seems the browser tests are trying to do real login instead of mock login. This should not happen in review environment, so that is most likely a separate problem already present in `develop`.

Merging as is to fix the migrations in `develop`.

PR #628, opened 2022-01-13
HL-377 | Update azure trigger path

quyenlq:


haavikko (approved):
Looks good to me! The newlines seem to have changed, are they all in the same format now?

PR #630, opened 2022-01-13
HL-374 | Make some calculation and paysubsidy fields nullable

quyenlq:


haavikko (changes requested):
The initial values of some Calculation and PaySubsidy fields should also be null, according to the ticket:
Calculation.start_date
Calculation.end_date
PaySubsidy.start_date
PaySubsidy.end_date
Calculation.state_aid_max_percentage

This probably requires change in Calculation.reset and PaySubsidy.reset_pay_subsidies,
and changing state_aid_max_percentage default to None

haavikko:
Also, just realized in the refinement meeting that HelsinkiBenefitCalculator code seems to rely on these fields being non-null, so that needs to be fixed too.

quyenlq:
> Also, just realized in the refinement meeting that HelsinkiBenefitCalculator code seems to rely on these fields being non-null, so that needs to be fixed too.

I updated the PR, changed the calculator logic so it'll not generate calculation row `if calculation.start_date and calculation.end_date is None`. Also added some exception & fallback data if values are null. Actually, I don't really know the use case of nullable start_date and end_date so let me know if this is ok

haavikko (approved):
Looks good to me! Made some minor changes to tests

haavikko:
Merging on behalf of quyenlq

PR #582, opened 2022-01-03
YJDH-345 | KS-Backend: Use Finnish PostgreSQL in Docker

karisal-anders:
## Description :sparkles:

Based on discussion in https://github.com/docker-library/postgres/issues/725#issuecomment-624141851

### KS-Backend: Use Finnish PostgreSQL in Docker:

backend/docker/kesaseteli.Dockerfile:
 - Add finnish_postgres target based on postgres:12 which sets up the
   PostgreSQL image to use Finnish language and database collation

docker-compose.youth.yml:
 - Use the finnish_postgres target

Using finnish_postgres kesaseteli-db is initialized like this:
 - The database cluster will be initialized with locale "fi_FI.UTF-8".
 - The default database encoding has accordingly been set to "UTF8".
 - The default text search configuration will be set to "finnish".

Listing the databases in the kesaseteli-db using
```docker exec -it kesaseteli-db psql -Ukesaseteli -c"\l"```
now gives (Removed irrelevant information for compact output):

```
                       List of databases
    Name    |   Owner    | Encoding |   Collate   |    Ctype    |
------------+------------+----------+-------------+-------------+
 kesaseteli | kesaseteli | UTF8     | fi_FI.UTF-8 | fi_FI.UTF-8 |
 postgres   | kesaseteli | UTF8     | fi_FI.UTF-8 | fi_FI.UTF-8 |
 template0  | kesaseteli | UTF8     | fi_FI.UTF-8 | fi_FI.UTF-8 |
 template1  | kesaseteli | UTF8     | fi_FI.UTF-8 | fi_FI.UTF-8 |
```

### SchoolListView: Remove unnecessary custom sorter:

Now the underlying PostgreSQL database uses fi_FI.UTF-8 collation so the
sorting works without the custom sorter.

### ks-pytest.yml: Use Finnish PostgreSQL:

Use Finnish PostgreSQL in review environment's (KS) Python tests so
test_schools_list_returns_sorted_collection test from
applications/tests/test_schools_api.py passes in review environment.

## Issues :bug:

YJDH-345

## Testing :alembic:

If you need to you can drop the database volume so it gets recreated with the correct language and collation:
```
docker-compose -f .\docker-compose.youth.yml down
docker volume rm yjdh_pgdata
yarn youth --build
```

PR #631, opened 2022-01-14
YJDH-405 | Fix ks-pytest.yml by not using Finnish Postgres

karisal-anders:
## Description :sparkles:

Revert the changes from PR #582 related to GitHub Actions.

Also:
 - Move Finnish Postgres to own Dockerfile and use it from there
 - Add back PostgreSQL specific custom sorter to SchoolListView because
   otherwise the backend tests won't pass using GitHub Actions.

NOTE:
 - Hotfixing YJDH-405 to get "(KS) Python tests" working again in pull
   requests while looking for a way to use Finnish Postgres in GitHub
   Actions.

## Issues :bug:

YJDH-405

### DIFF
--- a/.github/workflows/ks-pytest.yml
+++ b/.github/workflows/ks-pytest.yml
@@ -20,10 +20,7 @@ jobs:
 
     services:
       postgres:
-        build:
-          context: ../../backend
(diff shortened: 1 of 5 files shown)


haavikko (approved):
LGTM!

PR #633, opened 2022-01-14
HL-379: time range validation errors should be warnings.

haavikko:
## Description :sparkles:

Add new "warnings" element to the API that can be used for other kinds of warnings, too.
Rename past_benefit_info.
Add tests for overlapping previously granted benefit.

### DIFF
--- a/backend/benefit/applications/api/v1/serializers.py
+++ b/backend/benefit/applications/api/v1/serializers.py
@@ -1,5 +1,6 @@
 import re
 from datetime import date, datetime, timedelta
+from typing import Dict, List
 
 import filetype
 from applications.api.v1.status_transition_validator import (
@@ -7,7 +8,7 @@ from applications.api.v1.status_transition_validator import (
     ApplicationBatchStatusValidator,
     HandlerApplicationStatusValidator,
 )
-from applications.benefit_aggregation import get_past_benefit_info
+from applications.benefit_aggregation import get_former_benefit_info
 from applications.enums import (
     ApplicationBatchStatus,
     ApplicationStatus,
@@ -500,7 +501,7 @@ class BaseApplicationSerializer(serializers.ModelSerializer):
             "attachment_requirements",
             "applicant_terms_approval_needed",
             "applicant_terms_in_effect",
-            "past_benefit_info",
+            "former_benefit_info",
             "available_benefit_types",
             "official_company_street_address",
             "official_company_city",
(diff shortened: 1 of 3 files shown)


backend/benefit/applications/tests/test_benefit_aggregation.py
```diff
@@ -33,7 +34,7 @@
             ],
             BenefitType.SALARY_BENEFIT,
             NO_APPRENTICESHIP,
-            400,
+            "Benefit can not be granted before 24-month waiting period expires",
```
karisal-anders, line 37: The source text you are comparing against is a translatable string and here it is not. If someone adds translations for the source string then wouldn't this possibly change to something else?

backend/benefit/applications/api/v1/serializers.py
```diff
@@ -729,6 +733,40 @@ def get_applicant_terms_in_effect(self, obj):
         else:
             return None
 
+    def get_warnings(self, obj):
```
karisal-anders, line 736: How would you feel about adding a return type of `Dict[str, List[str]]` to make the returned data format more explicit?

backend/benefit/applications/api/v1/serializers.py
```diff
@@ -729,6 +733,40 @@ def get_applicant_terms_in_effect(self, obj):
...
+
+        More fields may be added in the future. The format of the data is:
+            {
+                "some_field_name_or_key": [
+                    "warning_string"
+                ],
+                "other_field_name_or_key": [
+                    "warning_string"
+                ],
+            }
```
karisal-anders, line 749: Looking at the code the returned lists as the dictionary values here may be of arbitrary non-empty length but the documentation seems to suggests that they're always of length 1.

haavikko: Hmm I'll make the documentation more clear about that

backend/benefit/applications/tests/test_benefit_aggregation.py
haavikko: I'll check, maybe need to force the locale explictly

backend/benefit/applications/api/v1/serializers.py
haavikko: good idea. I think we could use type annotations everywhere it provides a most benefit.

karisal-anders (approved):
Looks ok to me to merge!

PR #636, opened 2022-01-14
TETP-34: CRUD operations for developing & review

mattikl:
## Description :sparkles:

* startapp postings (this will be used to communicate with LinkedEvents in production)
* create model TetPostingTemp and views for CREATE and READ operations that are enabled when 
`NEXT_PUBLIC_MOCK_FLAG` is set
* fix review environment settings

This setup helps us test the frontend against the backend while developing and in the review environment.


mattikl:
This approach is not needed at the moment.

PR #637, opened 2022-01-14
TETP-23: Job Posting details

mattikl:
## Description :sparkles:

* add `react-hook-form`, use `FormProvider` in the `Editor` component
* add `EditorErrorNotification`
* hook components in `PostingDetails` to form
* use `userQuery` in header (displays the logged in user)
* reorder pages, add page for editing TET job postings, add better login page
* show list of job postings


frontend/tet/admin/Dockerfile
```diff
@@ -26,6 +26,8 @@ COPY --chown=appuser:appuser . .
 FROM appbase as development
 # =============================
 
+ENV NODE_OPTIONS="--max_old_space_size=8192"
```
mattikl, line 29: This was added because Docker Compose was crashing due to V8 heap out of memory (both I and @pekkvil experienced this problem). We should still check that this doesn't cause the container to use more memory in review and other environments.

pekkvil (approved):

PR #638, opened 2022-01-17
HL-365: automatically generate message when application is reopened

haavikko:


haavikko:
Changing to shorter branch name

PR #656, opened 2022-01-18
HL-273: Fix status label

melniiv:
## Description :sparkles:
-Replaced custom status alert label with HDS one
-Fixed the bug with appearing of additional info required status label in the applications list

### DIFF
--- a/frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
+++ b/frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
@@ -1,5 +1,6 @@
 import { APPLICATION_STATUSES } from 'benefit/handler/constants';
 import { ApplicationListItemData } from 'benefit/handler/types/application';
+import { StatusLabel } from 'hds-react';
 import * as React from 'react';
 import LoadingSkeleton from 'react-loading-skeleton';
 import Container from 'shared/components/container/Container';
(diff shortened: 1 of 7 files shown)


frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
```diff
@@ -96,22 +96,22 @@ const ApplicationList: React.FC<ApplicationListProps> = ({
...
-          >
-            {t(
-              `common:applications.list.columns.additionalInformationNeededByVal`,
-              {
-                date: convertToUIDateFormat(additionalInformationNeededBy),
-              }
-            )}
-          </$DataValueAlert>
+          <div>
+            {itemStatus === APPLICATION_STATUSES.INFO_REQUIRED ? (
```
nenonja, line 104: unneeded check? already checked on line 93? Or what's the difference between status and itemStatus

nenonja (approved):
One minor comment, approved anyway because it might be unecessary notification

frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
melniiv: yes, as we have applications with multiple statuses in the list

PR #657, opened 2022-01-18
YJDH-398 | KS-Backend: Add error handling to youth application creation

karisal-anders:
### KS-Backend: Remove unnecessary savings from tests 

Remove unnecessary savings of YouthApplication model instances from
test_youth_applications_api.py tests. This is in preparation for the
YJDH-398 tests where existing saved YouthApplication instances would
make a difference.

Additionally:
 - Make YouthApplicationSerializer return more specific error by
   specifying the social_security_number field as the source of the
   validation error.
 - Update tests to check that returned error is field specific:
   - test_youth_application_post_missing_required_field
   - test_youth_application_post_empty_required_field
 
### KS-Backend: Add error handling to youth application creation

When creating/submitting a new youth application handle errors:
 - Activated youth application with same email or social security number
   already exists
   - Return 400 Bad Request and JSON:
     {"code": "already_assigned", "message": "<human readable error>"}
 - Email is already used by inactive unexpired or active youth
   application
   - Return 400 Bad Request and JSON:
     {"code": "email_in_use", "message": "<human readable error>"}
 - Add test_youth_application_post_error_codes tests for the above

Add YouthApplicationRejectedReason enum:
 - Values:
   - EMAIL_IN_USE
   - ALREADY_ASSIGNED
 - Add json function for outputting value and label of enum as JSON
   - Add test test_youth_application_rejected_reason_json for testing
     the output of this function

Add YouthApplicationQuerySet:
 - Add member functions for ease of use:
   - matches_email_or_social_security_number
   - expired
   - unexpired
   - unexpired_or_active
   - active
 - Add tests for the above member functions to test_querying.py
 - Use YouthApplicationQuerySet as object manager in YouthApplication

Additionally:
 - backend/shared/shared/common/utils.py:
   - Add any_of_q_filter(**kwargs) function
     - Return Q filters combined with | i.e. match any of the given
       keyword arguments. NOTE: Supports "not__" prefix to invert a
       particular Q expression i.e. ~Q().
     - Example: any_of_q_filter(a=1, b__lt=2, not__c__gte=3) returns
       Q(a=1) | Q(b__lt=2) | ~Q(c__gte=3)
   - Add MatchesAnyOfQuerySet class
     - QuerySet which supports OR filtering using matches_any_of member
       function
     - Example if TestClass used TestClassQuerySet(MatchesAnyOfQuerySet)
       based object manager:
       TestClass.objects.matches_any_of(a=1, b__lt=2, not__c__gte=3)
       would match like
       TestClass.objects.filter(Q(a=1) | Q(b__lt=2) | ~Q(c__gte=3))
 - backend/shared/shared/common/tests/test_utils.py:
   - Add test_any_of_q_filter test for testing any_of_q_filter function
 - Test the matches_any_of with YouthApplicationQuerySet

## Issues :bug:

YJDH-398

### DIFF
--- a/backend/kesaseteli/applications/models.py
+++ b/backend/kesaseteli/applications/models.py
@@ -5,10 +5,12 @@ from urllib.parse import quote, urljoin
 from django.conf import settings
 from django.core.mail import send_mail
 from django.db import models, transaction
+from django.db.models import DurationField, ExpressionWrapper, F, Q
 from django.urls import reverse
 from django.utils import timezone, translation
 from django.utils.translation import gettext, gettext_lazy as _
 from encrypted_fields.fields import EncryptedCharField, SearchField
+from shared.common.utils import MatchesAnyOfQuerySet
 from shared.common.validators import (
     validate_name,
     validate_optional_json,
@@ -46,6 +48,69 @@ class School(TimeStampedModel, UUIDModel):
         verbose_name = _("school")
         verbose_name_plural = _("schools")
         ordering = ["name"]
+
+
+class YouthApplicationQuerySet(MatchesAnyOfQuerySet, models.QuerySet):
+    def _annotate_existence_duration(self):
+        """
+        Annotate queryset with existence_duration with value timezone.now() - created_at
+
+        :return: The source queryset with annotated existence_duration DurationField
+                 with value timezone.now() - created_at.
+        """
+        return self.annotate(
+            existence_duration=ExpressionWrapper(
+                timezone.now() - F("created_at"), output_field=DurationField()
+            )
+        )
+
+    def _active_q_filter(self) -> Q:
+        """
+        Return Q filter for active youth applications
+        """
+        return Q(receipt_confirmed_at__isnull=False)
+
+    def _unexpired_q_filter(self) -> Q:
+        """
+        Return Q filter for unexpired youth applications.
+
+        NOTE: Needs existence_duration field, so use with _annotate_existence_duration.
+        """
+        return Q(existence_duration__lt=YouthApplication.expiration_duration())
+
+    def matches_email_or_social_security_number(self, email, social_security_number):
+        """
+        Return youth applications that match given email or social security number
+        """
+        return self.matches_any_of(
+            email=email, social_security_number=social_security_number
+        )
+
+    def expired(self):
+        """
+        Return youth applications that are expired
+        """
+        return self._annotate_existence_duration().filter(~self._unexpired_q_filter())
+
+    def unexpired(self):
+        """
+        Return youth applications that are unexpired
+        """
+        return self._annotate_existence_duration().filter(self._unexpired_q_filter())
+
+    def unexpired_or_active(self):
+        """
+        Return youth applications that are unexpired or active
+        """
+        return self._annotate_existence_duration().filter(
+            self._unexpired_q_filter() | self._active_q_filter()
+        )
+
+    def active(self):
+        """
+        Return active youth applications
+        """
+        return self.filter(self._active_q_filter())
 
 
 class YouthApplication(TimeStampedModel, UUIDModel):
@@ -97,6 +162,7 @@ class YouthApplication(TimeStampedModel, UUIDModel):
         verbose_name=_("vtj json"),
         validators=[validate_optional_json],
     )
+    objects = YouthApplicationQuerySet.as_manager()
 
     def _localized_frontend_page_url(self, page_name):
         return urljoin(
@@ -115,6 +181,12 @@ class YouthApplication(TimeStampedModel, UUIDModel):
     def _activation_link(self, request):
         return request.build_absolute_uri(
             reverse("v1:youthapplication-activate", kwargs={"pk": self.id})
+        )
+
+    @staticmethod
+    def expiration_duration() -> timedelta:
+        return timedelta(
+            seconds=settings.NEXT_PUBLIC_ACTIVATION_LINK_EXPIRATION_SECONDS
         )
 
     @property
@@ -172,6 +244,35 @@ class YouthApplication(TimeStampedModel, UUIDModel):
             self.receipt_confirmed_at = timezone.now()
             self.save()
         return self.is_active
+
+    @classmethod
+    def is_email_or_social_security_number_active(
+        cls, email, social_security_number
+    ) -> bool:
+        """
+        Is there an active youth application created that uses the same email or social
+        security number as this youth application?
+
+        :return: True if this youth application's email or social security number are
+                 used by at least one active youth application, otherwise False.
+        """
+        return (
+            cls.objects.matches_email_or_social_security_number(
+                email, social_security_number
+            )
+            .active()
+            .exists()
+        )
+
+    @classmethod
+    def is_email_used(cls, email) -> bool:
+        """
+        Is this youth application's email used by unexpired or active youth application?
+
+        :return: True if this youth application's email is used by at least one
+                 unexpired or active youth application, otherwise False.
+        """
+        return cls.objects.filter(email=email).unexpired_or_active().exists()
 
     @property
     def name(self):
--- a/backend/shared/shared/common/utils.py
+++ b/backend/shared/shared/common/utils.py
@@ -1,0 +1,67 @@
+import operator
+from functools import reduce
+
+from django.db.models import Q, QuerySet
+
+_ALWAYS_FALSE_Q_FILTER = Q(pk=None)  # A hack but works as primary keys can't be null
+
+
+def any_of_q_filter(**kwargs):
+    """
+    Return Q filters combined with | i.e. match any of the given keyword arguments.
+    Return an always false Q filter if no keyword arguments are given.
+
+    NOTE: Supports "not__" prefix to invert a particular Q expression i.e. ~Q().
+
+    Example:
+        any_of_q_filter(a=1, b__lt=2, not__c__gte=3)
+        returns
+            Q(a=1) | Q(b__lt=2) | ~Q(c__gte=3)
+
+    :warning: Does not support field with name "not"
+    :warning: Does not support multiple "not__" prefixes for a single field
+    """
+    if not kwargs:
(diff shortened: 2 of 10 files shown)


karisal-anders:
Tested and the youth `browser-test:ci` passes locally with Docker environment with:
 - `yarn youth --build`
 - `cd ./frontend/kesaseteli/youth/`
 - `yarn browser-test:ci`
```
frontend/kesaseteli/youth> yarn browser-test:ci
yarn run v1.22.17
$ testcafe 'chrome:headless --disable-gpu --window-size="1249,720" --ignore-certificate-errors-spki-list="8sg/cl7YabrOFqSqH+Bu0e+P27Av33gWgi8Lq28DW1I=,gJt+wt/T3afCRkxtMMSjXcl/99sgzWc2kk1c1PC9tG0=,zrQI2/1q8i2SRPmMZ1sMntIkG+lMW0legPFokDo3nrY="' --screenshots path=report,fullPage=true --video report --reporter spec,custom,html:report/index.html testcafe.dotenv.js browser-tests/
-
 Running tests in:
 - Chrome 97.0.4692.71 / Windows 10

 Frontpage
 √ can send application and return to front page

 Frontpage
 √ can change to languages


 2 passed (20s)
Done in 37.13s.
```

karisal-anders:
The tests failing has been reported to DevOps as ticket [DEVOPS-374](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-374).

karisal-anders:
The tests failing has been reported to DevOps as ticket [DEVOPS-368](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-368).

backend/kesaseteli/applications/models.py
```diff
@@ -48,6 +50,69 @@ class Meta:
...
+        """
+        return Q(receipt_confirmed_at__isnull=False)
+
+    def _unexpired_q_filter(self) -> Q:
+        """
+        Return Q filter for unexpired youth applications.
+
+        NOTE: Needs existence_duration field, so use with _annotate_existence_duration.
+        """
+        return Q(existence_duration__lt=YouthApplication.expiration_duration())
```
quyenlq, line 79: Do you have any mechanism to auto clean expired YouthApplication? How many YouthApplication are we expected to have?
I think the way you query unexpired application is quite expensive and might cause performance issues when the number of applications grows up. Backend will always need to recalculate the existence duration of **every** Application in **every request** from the client. I believe an expired application most likely will never become unexpired again so it's not a good idea to try to re-calculate the existence duration of every application.
So, unless the system requires an application to be unavailable as soon as it expired, I'd make a separated BooleanField, to store the expired status of the application, then use a cronjob that runs every minute (or 5-15 minutes depends on the need) to query every unexpired application and update status if it expires

backend/shared/shared/common/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        (
+            ~Q(**{key[len("not__") :]: value})
+            if key.startswith("not__")
+            else Q(**{key: value})
+            for key, value in kwargs.items()
+        ),
+    )
+
+
+class MatchesAnyOfMixin(QuerySet):
```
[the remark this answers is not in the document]
karisal-anders: Taken into account by renaming MatchAnyOfMixin to MatchAnyOfQuerySet.

karisal-anders:
Force pushed:
 - Rebased on develop
 - Updated commit message related to MatchesAnyOfMixin → MatchesAnyOfQuerySet change
 - Renamed MatchesAnyOfMixin to MatchesAnyOfQuerySet and updated documentation—see the diff below
```diff
diff --git a/backend/kesaseteli/applications/models.py b/backend/kesaseteli/applications/models.py
index 7aaaf977..01b05bad 100644
--- a/backend/kesaseteli/applications/models.py
+++ b/backend/kesaseteli/applications/models.py
@@ -10,7 +10,7 @@ from django.urls import reverse
 from django.utils import timezone, translation
 from django.utils.translation import gettext, gettext_lazy as _
 from encrypted_fields.fields import EncryptedCharField, SearchField
-from shared.common.utils import MatchesAnyOfMixin
+from shared.common.utils import MatchesAnyOfQuerySet
 from shared.common.validators import (
     validate_name,
     validate_optional_json,
@@ -50,7 +50,7 @@ class School(TimeStampedModel, UUIDModel):
         ordering = ["name"]
 
 
-class YouthApplicationQuerySet(MatchesAnyOfMixin, models.QuerySet):
+class YouthApplicationQuerySet(MatchesAnyOfQuerySet, models.QuerySet):
     def _annotate_existence_duration(self):
         """
         Annotate queryset with existence_duration with value timezone.now() - created_at
diff --git a/backend/shared/shared/common/utils.py b/backend/shared/shared/common/utils.py
index f430f142..2a794287 100644
--- a/backend/shared/shared/common/utils.py
+++ b/backend/shared/shared/common/utils.py
@@ -34,18 +34,18 @@ def any_of_q_filter(**kwargs):
     )
 
 
-class MatchesAnyOfMixin(QuerySet):
+class MatchesAnyOfQuerySet(QuerySet):
     """
-    Mixin for adding matches_any_of member function to a custom QuerySet class.
+    QuerySet which supports OR filtering using matches_any_of member function.
+    See matches_any_of function for details.
 
     NOTE: Put left in class declaration to have consistent method resolution order i.e.
-          Test(MatchesAnyOfMixin, QuerySet) instead of Test(QuerySet, MatchesAnyOfMixin)
+          Test(MatchesAnyOfQuerySet, QuerySet) works
+          but Test(QuerySet, MatchesAnyOfQuerySet) does not
 
-    Example if TestClass used TestClassQuerySet(MatchesAnyOfMixin) based object manager:
+    Example if TestClass used TestClassQuerySet(MatchesAnyOfQuerySet) as object manager:
         TestClass.objects.matches_any_of(a=1, b__lt=2, not__c__gte=3) would match like
             TestClass.objects.filter(Q(a=1) | Q(b__lt=2) | ~Q(c__gte=3))
-
-    :warning: Does not support field with name "not"
     """
 
     def matches_any_of(self, **kwargs):
```

karisal-anders:
There was a warning when running the youth backend's tests:
```
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[False-False-False-True-None]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[False-False-True-True-None]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[False-True-False-True-None]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[False-True-True-True-already_assigned]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[True-False-False-True-None]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[True-False-True-True-already_assigned]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[True-True-False-True-None]
applications/tests/test_youth_applications_api.py::test_youth_application_post_error_codes[True-True-True-True-already_assigned]
  /usr/local/lib/python3.8/site-packages/django/db/models/fields/__init__.py:1416: RuntimeWarning: DateTimeField YouthApplication.created_at received a naive datetime (2021-01-01 00:00:00) while time zone support is active.
    warnings.warn("DateTimeField %s received a naive datetime (%s)"
```

Removed the warning by force pushing the following changes:
```diff
diff --git a/backend/kesaseteli/applications/tests/test_youth_applications_api.py b/backend/kesaseteli/applications/tests/test_youth_applications_api.py
index d2e3203c..fe1f3838 100644
--- a/backend/kesaseteli/applications/tests/test_youth_applications_api.py
+++ b/backend/kesaseteli/applications/tests/test_youth_applications_api.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from typing import Optional
 
 import factory.random
@@ -12,6 +13,7 @@ from rest_framework.reverse import reverse
 
 from applications.api.v1.serializers import YouthApplicationSerializer
 from applications.enums import get_supported_languages, YouthApplicationRejectedReason
+from applications.models import YouthApplication
 from common.tests.factories import (
     ActiveYouthApplicationFactory,
     InactiveYouthApplicationFactory,
@@ -401,7 +403,9 @@ def test_youth_application_post_error_codes(
     existing_app.receipt_confirmed_at = now if is_existing_active else None
     if is_existing_expired:
         # Make the saved youth application expired
-        existing_app.created_at = timezone.datetime(2021, 1, 1)
+        existing_app.created_at = (
+            now - YouthApplication.expiration_duration() - timedelta(hours=1)
+        )
     else:
         existing_app.created_at = now
     existing_app.save()
```

backend/kesaseteli/applications/models.py
karisal-anders: No, there's no auto clean mechanism available at the moment.

Based on the previous years' data from a product owner I'd say the YouthApplication objects per year would probably realistically be in range 5k–10k. This is for one application per one person situation. As it is possible to submit multiple applications per person then this might make the resulting object count higher, e.g. double it to 20k.

It is true that the performance may become an issue with higher amount of YouthApplication objects as the `existence_duration` annotated by `YouthApplicationQuerySet._annotate_existence_duration` is not indexed and can't be indexed by a functional index because it would not be immutable.

We have the expiration time configurable by a shared setting between the backend and frontend to keep it dynamic. With this conversation it's evident this has a downside related to query performance. If we could and would hardcode the value then we could create a functional index on `YouthApplication.created_at + <hardcoded expiration time>` and then query against that using `timezone.now()` to filter out the expired YouthApplication objects. That query should hopefully be fast by using the functional index.

The marking with a cronjob sounds like a viable solution too.

I propose in the lean spirit to split the possible query performance issue fixing to another ticket and not include its implementation in this pull request. How would you feel about this?

quyenlq: Totally agree, thanks for the clarification!

quyenlq (approved):

backend/kesaseteli/applications/models.py
karisal-anders: You're welcome!

PR #661, opened 2022-01-19
HL-401: fix error that prevented applicant from sending messages

haavikko:
* The allowed statuses are listed in APPLICANT_MESSAGE_ALLOWED_STATUSES
* Also add validation that the handler must send a message first,
  before the applicant is able to do so.


quyenlq (approved):
LGTM!

PR #662, opened 2022-01-19
HL-283: Applicant messanging

melniiv:
## Description :sparkles:
- Messenger added to the applicant portal
- polling every 30 seconds refetching messages in applicant and handler side
- added close drawer button functionality
- added empty messages to the messages list (not to the notes, missing design)
- fixed small visual bug in handlers review page with attachments alignments


frontend/benefit/handler/src/components/messenger/useMessenger.ts
```diff
@@ -35,6 +37,14 @@ const useMessenger = (): ExtendedComponentProps => {
     MESSAGE_URLS.NOTES
   );
 
+  const doPolling = React.useCallback((): void => {
```
nenonja, line 40: I dont think you need polling, this is what react-query does automatically for you. I havent used refetch myself but I think you can just set  `refetchInterval` in the useQuery options inside useMessagesQuery: https://react-query.tanstack.com/reference/useQuery

nenonja (changes requested):
One requested change on polling...
Also I would suggest instead of copy-pasting the messenger I would just move those common components now to benefit/shared package and when they start to differ, then parameterize the props them accordingly. But you can go there the way you want it, I understand that the requirements

frontend/benefit/handler/src/components/messenger/useMessenger.ts
melniiv: cool, i ll test!

frontend/benefit/applicant/src/components/header/Header.tsx
```diff
@@ -22,27 +22,16 @@ const Header: React.FC = () => {
   const logoutQuery = useLogoutQuery();
 
   const isLoading = userQuery.isLoading || logoutQuery.isLoading;
-  const isLoginPage = asPath?.startsWith('/login');
+  const isLoginPage = asPath?.startsWith(ROUTES.LOGIN);
+  const isApplicationPage = asPath?.startsWith(ROUTES.APPLICATION_FORM);
 
-  const [isDrawerOpen, setIsDrawerOpen] = React.useState(false);
+  const [isMessagesDrawerVisible, setIsMessagesDrawerVisible] =
```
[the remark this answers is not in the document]
melniiv: fixed

frontend/benefit/handler/src/components/messenger/useMessenger.ts
melniiv: fixed

nenonja (approved):
Looks good now, but I was not able to test it. Also the copy-paste code with messenger stays there but it's perhaps ok because requirements might change.

frontend/shared/src/components/benefit/tabs/Tabs.sc.ts
```diff
@@ -1,13 +1,17 @@
 import { TabList, TabPanel } from 'hds-react';
```
nenonja, line 1: the file path is > `frontend/shared/src/components/benefit/tabs/Tabs.sc.ts` Should the contents of the `benefit` folder be in `benefit-shared` instead?

PR #665, opened 2022-01-20
Pipeline yaml files for youth ui main branch

Vitals9367:


mikkovihonen:
This will cause deployments of all UIs regardless of changed UI. This is accepted by the dev team as long as it doesn't make deployments hang (long deployment time is not an issue as PLATTA is not used for DEV style use).

PR #674, opened 2022-01-21
Fix incorrect database url

nenonja:
## Description :sparkles:
Database name doesnt match to the name path in the database url


Hi-Fi (approved):
I think stops are already OK.

PR #680, opened 2022-01-24
HL-372: Fix messenger

melniiv:
## Description :sparkles:
-Adds line breaks to the messages
-shows message icon for applicant only in additional info required application status
-shows count of unread messages for each application list item in the list


nenonja (approved):
One nitpick comment, but otherwise looks good and approved 👍

nenonja (approved):
Just wait for jobs to finish and check that everything related is green

PR #699, opened 2022-01-26
HL-414: Application freezing on submit fix

melniiv:
## Description :sparkles:
-Fix submitting of the application
-Fix notification for submitted application
-Fix resending application

### DIFF
--- a/frontend/benefit/applicant/src/components/header/useHeader.ts
+++ b/frontend/benefit/applicant/src/components/header/useHeader.ts
@@ -1,10 +1,11 @@
-import AppContext from 'benefit/applicant/context/AppContext';
+import { APPLICATION_STATUSES } from 'benefit/applicant/constants';
+import useApplicationQuery from 'benefit/applicant/hooks/useApplicationQuery';
 import useLocale from 'benefit/applicant/hooks/useLocale';
 import { useTranslation } from 'benefit/applicant/i18n';
 import { getLanguageOptions } from 'benefit/applicant/utils/common';
 import { useRouter } from 'next/router';
 import { TFunction } from 'next-i18next';
-import React from 'react';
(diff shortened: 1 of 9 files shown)


frontend/benefit/applicant/src/components/header/useHeader.ts
```diff
@@ -24,14 +25,27 @@ const useHeader = (): ExtendedComponentProps => {
...
+  const [hasMessenger, setHasMessenger] = useState<boolean>(false);
   const { pathname, asPath, query } = router;
-  const { hasMessenger } = React.useContext(AppContext);
 
   const languageOptions = React.useMemo(
     (): OptionType<string>[] => getLanguageOptions(t, 'supportedLanguages'),
     [t]
   );
 
+  const { data: application } = useApplicationQuery(id);
```
nenonja, line 37: Generally this is better than using the context 👍

frontend/benefit/applicant/src/context/AppContext.tsx
```diff
@@ -4,15 +4,11 @@ import React from 'react';
 import { SubmittedApplication } from '../types/application';
 
 export type AppContextType = {
```
nenonja, line 6: Usually react-query caches the latest submitted application, so I think only the id is needed to save to the context and then use `useApplicationQuery(id)` to get the cached value. 
Also Context has the problem that it vanishes if page is reloaded (also reload happens now when changing language). query url parameter or localstorage could be more persistent.

nenonja: This is anyway more like a comment, not request for a change :)

nenonja (approved):
Some general comments left, no change requests though so appeoved

PR #703, opened 2022-01-27
HL-400: handler UI configuration for the KuVa review environment

haavikko:
## Description :sparkles:

Run the handler UI in the KuVa review environment

### DIFF
--- a/.github/workflows/bf-review.yml
+++ b/.github/workflows/bf-review.yml
@@ -37,6 +37,7 @@ env:
   K8S_PROBE_PERIOD: 20
   NEXT_PUBLIC_BACKEND_URL: https://helsinkibenefit-bf-bknd-${{ github.event.pull_request.number }}.${{ secrets.BASE_DOMAIN_STAGING }}
   NEXT_PUBLIC_MOCK_FLAG: 1
+
 jobs:
   build:
     # No building for dependabot PRs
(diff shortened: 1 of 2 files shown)


.github/workflows/bf-review.yml
```diff
@@ -45,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        service: ["bf-bknd", "bf-appl"]
+        service: ["bf-bknd", "bf-appl", "bf-hand"]
```
nenonja, line 49: the name must be `bf-hdlr` because that is reserved namespace name, otherwise it does not work

.github/workflows/bf-review.yml
```diff
@@ -75,7 +80,7 @@ jobs:
...
```
nenonja, line 83: the name must be `bf-hdlr` because that is the reserved namespace name, otherwise it does not work

Hi-Fi: Same applies access to Vault that is based on the same

.github/workflows/bf-review.yml
haavikko: OK thanks, fixed!

haavikko:
Tested with the successful deployment:
First, login to the backend django admin at https://helsinkibenefit-bf-bknd-703.test.kuva.hel.ninja/admin/ (username:admin password:admin2)
In future deployments, the password should also be "admin".
After that, open the handler UI at https://helsinkibenefit-bf-hdlr-703.test.kuva.hel.ninja/ -> the authentication with backend APIs should work

nenonja (approved):
Looks good now and seem to deploy also.

PR #707, opened 2022-01-28
Use Finnish locale in postgres in local development

haavikko:
## Description :sparkles:

Use Finnish locale for the Postgres instance used for local development

### DIFF
--- a/docker-compose.benefit.yml
+++ b/docker-compose.benefit.yml
@@ -1,12 +1,16 @@
 version: "3.8"
 services:
   postgres:
-    image: postgres:12
+    build:
+      context: ./backend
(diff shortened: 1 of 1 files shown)


karisal-anders (commented):
Commit message could be made specific to benefit, e.g. "BF: " or "BF-Backend: " prefix would solve that.

haavikko:
> Commit message could be made specific to benefit, e.g. "BF: " or "BF-Backend: " prefix would solve that.

good idea. Maybe we should use this practice in the future, when the context is not clear otherwise, for example if Jira issue ID isn't clearly visible in the case

karisal-anders (approved):
LGTM!

PR #711, opened 2022-01-28
Package version bumps 28.1.2022

nenonja:


frontend/benefit/applicant/src/components/applications/forms/application/step2/ApplicationFormStep2.tsx
```diff
@@ -244,7 +244,6 @@ const ApplicationFormStep2: React.FC<DynamicFormStepComponentProps> = ({
                 }
                 options={subsidyOptions}
                 id={fields.paySubsidyPercent.name}
-                name={fields.paySubsidyPercent.name}
```
nenonja, line 247: This was the problem here: https://github.com/City-of-Helsinki/yjdh/runs/4986525433?check_suite_focus=true
It seems that `name` was not necessary

nenonja:
@melniiv fixed the `closeButtonLabelText`which was missing from benefit. now this can be reviewed

melniiv (approved):

PR #712, opened 2022-01-28
Fix azure pipeline triggers

nenonja:
## Description :sparkles:
- fixed paths to include to trigger specific pipelines. Especially shared code is forgotten that when it changes, most of the services should be built
- added pipelines for kesäseteli handler ui


mikkovihonen (approved):
Should work as advertised.

PR #722, opened 2022-01-31
YJDH-459 | KS-Backend: Compile messages in Docker image build

mikkovihonen:
### KS-Backend: Compile messages in Docker image build

Compile messages in Docker image build in kesaseteli.Dockerfile to
remove the need for writeable localization directories in the Docker
container. This should fix the error in Platta environment about not
being able to write the generated django.mo files.

Also override the entrypoints in the docker-compose files using the
Kesäseteli backend because although "django-admin compilemessages" is
now run as a part of Docker image build, the resulting files are
overridden by the volume mounts in the docker-compose files and thus
have to regenerated.

## Issues :bug:

YJDH-459

### DIFF
--- a/backend/docker/kesaseteli.Dockerfile
+++ b/backend/docker/kesaseteli.Dockerfile
@@ -41,6 +41,11 @@ ENV DEV_SERVER=1
 COPY --chown=appuser:appuser /kesaseteli/ /app/
 
 USER appuser
+
+# Compile messages as a part of Docker image build so it doesn't have to be done during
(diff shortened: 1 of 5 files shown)


nenonja (commented):
It seems that deploying backend timeouts: https://github.com/City-of-Helsinki/yjdh/runs/5018490108?check_suite_focus=true
Perhaps because of the change? rerunning didnt help.

karisal-anders (approved):
LGTM!

nenonja (approved):

PR #723, opened 2022-02-01
TETP-17 Employer can delete posting and base dialog component

pekkvil:
…ete posting mutation


mattikl (changes requested):
The shared `Modal` component is also used by https://github.com/City-of-Helsinki/yjdh/blob/develop/frontend/benefit/handler/src/components/applicationReview/actions/handlingApplicationActions/HandlingApplicationActions.tsx#L70 so prop `cancelButtonLabel` needs to be added.

Otherwise looks good, so good to merge after adding this, fixing the merge conflict and seeing that all tests pass.

frontend/shared/src/components/modal/Modal.tsx
```diff
@@ -65,7 +67,7 @@ const Modal: React.FC<ModalProps> = ({
               onClick={handleToggle}
               data-testid="cancel"
             >
-              {t('common:applications.actions.close')}
+              {cancelButtonLabel}
```
nenonja, line 70: It's not absolutely necessary to parameterize every text in shared components, if just use different texts in different application. So if you use shared Modal component, just bring translation with using the same key in your service.

Pros: less properties for shared component
Cons: have to use exactly the same key.

frontend/shared/src/components/toast/show-success-toast.ts
```diff
@@ -0,0 +1,11 @@
+import Toast from 'shared/components/toast/Toast';
```
nenonja, line 1: Another option would be to rename current `show-error-toast` to `error-toast` and add `type` as required parameter, so we could use same component for `type: 'error'` and `type: 'succes'`. Of course it needs some refactoring which can be daunting job...

frontend/tet/admin/src/pages/_app.tsx
```diff
@@ -11,11 +11,19 @@ import React from 'react';
...
+import ConfirmDialog from 'tet/admin/components/base/ConfirmDialog';
 
 const App: React.FC<AppProps> = (appProps) => (
   <BackendAPIProvider baseURL={getBackendDomain()}>
     <QueryClientProvider client={createQueryClient()}>
-      <BaseApp header={<Header />} footer={<Footer />} {...appProps} />
+      <DialogContextProvider>
+        <BaseApp header={<Header />} footer={<Footer />} {...appProps} />
+        <Portal>
+          <ConfirmDialog></ConfirmDialog>
```
nenonja, line 24: `<ConfirmDialog />` ?

frontend/tet/admin/src/pages/_document.tsx
```diff
@@ -0,0 +1,16 @@
+import Document, { Html, Head, Main, NextScript } from 'next/document';
```
nenonja, line 1: There is `ServerDocment` in shared code which perhaps could be used?... just the portal part should rendered only optionally, if some new property (ie `portal:boolean`) is given.

frontend/tet/admin/src/store/DialogContext.tsx
```diff
@@ -0,0 +1,71 @@
...
+type DialogActionTypes = ShowConfirmAction | HideConfirmAction;
+
+const initialState: DialogState = {
+  show: false,
+  header: '',
+  content: '',
+};
+
+const reducer = (state: DialogState = initialState, action: DialogActionTypes) => {
+  const content = action?.payload?.content ? action.payload.content : '';
```
nenonja, line 46: `const content = action?.payload?.content ?? ''`

nenonja (commented):
Nice job with the modal! I probably also need confirmation dialog soon in kesaseteli handler-ui so it's nice addition. Some small somments added

mattikl (approved):

frontend/tet/admin/src/pages/_document.tsx
pekkvil: Ok, thanks! I missed the shared option. I'll check if that could be used.

frontend/shared/src/components/toast/show-success-toast.ts
pekkvil: I also thought about using multi purpose component but chose the easy way instead and used the same convention that was used of the show-success

pekkvil:
Failing back end checks in review enviroment but no back end changes in pr

PR #726, opened 2022-02-02
HL-421: reduce review environment build failures by limiting concurrency of jobs

haavikko:
## Description :sparkles:
Juho Saarinen commented in Slack: "Basically lock should be namespace + service"

NOTE: If this seems to work, I'll add the same config for other YJDH subprojects

### DIFF
--- a/.github/workflows/bf-review.yml
+++ b/.github/workflows/bf-review.yml
@@ -60,6 +60,9 @@ jobs:
             context: ./frontend
             dockerfile: ./frontend/benefit/handler/Dockerfile
             port: 3100
+    concurrency:
+      group: ${{ github.event.pull_request.number }}-${{ matrix.service }}
+      cancel-in-progress: false
     runs-on: ubuntu-latest
     name: Build
     steps:
@@ -97,6 +100,9 @@ jobs:
             dockerfile: ./frontend/benefit/handler/Dockerfile
             database: false
             port: 3100
+    concurrency:
+      group: ${{ github.event.pull_request.number }}-${{ matrix.service }}
+      cancel-in-progress: false
     runs-on: ubuntu-latest
     needs: build
     name: Review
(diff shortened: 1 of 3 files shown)


.github/workflows/bf-review.yml
```diff
@@ -55,6 +55,9 @@ jobs:
             context: ./frontend
             dockerfile: ./frontend/benefit/applicant/Dockerfile
             port: 3000
+    concurrency:
+      group: ${{ env.K8S_NAMESPACE }}-${{ matrix.service }}
+      cancel-in-progress: false
```
Hi-Fi, line 60: I think there's no need to block build concurrency, as each build created different image anyway.

.github/workflows/bf-review.yml
```diff
@@ -87,6 +90,9 @@ jobs:
             dockerfile: ./frontend/benefit/applicant/Dockerfile
             database: false
             port: 3000
+    concurrency:
+      group: ${{ env.K8S_NAMESPACE }}-${{ matrix.service }}
```
Hi-Fi, line 94: Namespace is not available here, but only during the execution from `andersinno/kolga-setup-action@v2`.
```suggestion
      group: ${{ github.event.pull_request.number }}-${{ matrix.service }}
```

haavikko: OK thanks!

haavikko:
After the change suggested by @Hi-Fi , the build got further, but stopped with "timed out waiting for the condition". Retrying. Could be other builds that don't have the concurrency config are causing the error.

Hi-Fi:
> After the change suggested by @Hi-Fi , the build got further, but stopped with "timed out waiting for the condition". Retrying. Could be other builds that don't have the concurrency config are causing the error.

Other namespaces shouldn't directly matter, but without events it's hard to say if more time would help, or where the issue is

haavikko:
Still getting the "timed out waiting for condition" error. Don't know if it's due the new concurrency config or something else.

Hi-Fi:
> Still getting the "timed out waiting for condition" error. Don't know if it's due the new concurrency config or something else.

This fix shouldn't matter to that, but prevent the stage where Helm is stuck on e.g. `pending-upgrade` stage.

Timeout issue can be seen now from events:
```
0/9 nodes are available: 1 node(s) had taint {node.kubernetes.io/disk-pressure: }, that the pod didn't tolerate, 4 Insufficient memory, 6 Insufficient cpu.  
``` 

That would be fixed eventually by putting requests in the level application really needs. Currently (when idleing) apps require only about 1m CPU, and now those reserve 50m per pod.

haavikko:
@Hi-Fi The builds seem to pass now, I did the same changes to all subprojects, if you think this is OK I'll merge it and we'll see if helps with Helm stuckage.

Hi-Fi (approved):

PR #728, opened 2022-02-02
HL-430: apprenticeship_program only needs to be selected if pay_subsidy_granted is selected

haavikko:
## Description :sparkles:

Change validation of apprenticeship_program for submitted applications:
If pay_subsidy_granted is selected, then the applicant needs to also select if
it's an apprenticeship_program or not

## Testing :alembic:
backend/benefit/applications/tests/test_applications_api.py:test_apprenticeship_program_validation_on_submit


karisal-anders (approved):
For example `HL-430: Require apprenticeship_program only with pay_subsidy_granted` title would be [shorter](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting). Looks ok for me to merge, pending tests passing.

PR #743, opened 2022-02-03
YJDH-461 | KS-Backend: Prevent activating multiple youth applications

karisal-anders:
## Description :sparkles:

Prevent a person with the same social security number from activating
more than one youth application.

Add test test_youth_applications_dual_activate_unexpired_inactive.

## Issues :bug:

YJDH-461


haavikko (approved):
LGTM!

PR #747, opened 2022-02-04
YJDH-446 | KS-Backend: Set handler permission to mock flag state

karisal-anders:
## Description :sparkles:

Before everyone was a handler.
Now everyone is a handler if the mock flag is set, otherwise no one is.

Parametrize tests with/without mock flag and with different clients:
 - test_youth_applications_process_valid_pk
   - Should return 403 Forbidden for all without mock flag
   - Should return 501 Not implemented for all with mock flag
 - test_youth_applications_detail_valid_pk
   - Should return 403 Forbidden for all without mock flag
   - Should return 200 OK for all with mock flag

Rename invalid_pk->unused_pk & parametrize tests with/without mock flag:
 - test_youth_applications_process_unused_pk
   - Use random UUID rather than "invalid value" string as primary key
 - test_youth_applications_detail_unused_pk
   - Use random UUID rather than "invalid value" string as primary key

Test only with mock flag:
 - test_youth_applications_detail_response_field
 - test_youth_applications_detail_encrypted_vtj_json

Simplify YouthApplicationViewSet's process and activate functions by
making use of GenericAPIView.get_object().

## Issues :bug:

YJDH-446

### DIFF
--- a/backend/kesaseteli/applications/api/v1/views.py
+++ b/backend/kesaseteli/applications/api/v1/views.py
@@ -68,10 +68,7 @@ class YouthApplicationViewSet(AuditLoggingModelViewSet):
 
     @action(methods=["get"], detail=True)
     def process(self, request, pk=None) -> HttpResponse:
-        try:
-            YouthApplication.objects.get(pk=pk)
-        except (exceptions.ValidationError, YouthApplication.DoesNotExist):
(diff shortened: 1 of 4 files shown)


haavikko (approved):
LGTM!

PR #752, opened 2022-02-07
HL-450: add duration_in_months_rounded to the application API

haavikko:
## Description :sparkles:

The field is needed for display in the UI

## Testing :alembic:
backend/benefit/applications/tests/test_applications_api.py:test_application_single_read_as_applicant


karisal-anders (approved):
Is the missing start_date and end_date case tested? If not then adding such a test case would be good but I leave it up to you whether you want to add that.

haavikko:
> 

good point, added tests

PR #753, opened 2022-02-07
HL-442: application can be returned from ACCEPTED and REJECTED statuses back to HANDLING

haavikko:
## Description :sparkles:

The application handlers need to be able to revert their decision

## Testing :alembic:
backend/benefit/applications/tests/test_applications_api.py:test_application_status_change_as_handler

### DIFF
--- a/backend/benefit/applications/enums.py
+++ b/backend/benefit/applications/enums.py
@@ -18,14 +18,25 @@ class ApplicationStatus(models.TextChoices):
         if not user.is_authenticated:
             return False
         elif user.is_handler():
-            return cls.is_handler_editable_status(status)
+            return cls.is_handler_editable_status(status, None)
         else:
             return cls.is_applicant_editable_status(status)
 
     @classmethod
-    def is_handler_editable_status(cls, status):
+    def is_handler_editable_status(cls, status, new_status=None):
         if status not in cls.values:
             raise ValueError(_("Invalid application status"))
+        if new_status is not None and new_status not in cls.values:
+            raise ValueError(_("Invalid application status change"))
+
+        # Application can be transitioned back from these non-editable statuses, so make an exception
+        # to make the application editable if status transition is done at the same time
+        if (status, new_status) in [
+            (ApplicationStatus.ACCEPTED, ApplicationStatus.HANDLING),
+            (ApplicationStatus.REJECTED, ApplicationStatus.HANDLING),
+        ]:
+            return True
+
         # drafts may be edited by the handler when entering data from a paper application
         return status in (
             cls.DRAFT,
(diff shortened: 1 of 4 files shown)


backend/benefit/applications/enums.py
```diff
@@ -18,14 +18,25 @@ def is_editable_status(cls, user, status):
...
         else:
             return cls.is_applicant_editable_status(status)
 
     @classmethod
-    def is_handler_editable_status(cls, status):
+    def is_handler_editable_status(cls, status, new_status=None):
         if status not in cls.values:
             raise ValueError(_("Invalid application status"))
+        if new_status is not None and new_status not in cls.values:
+            raise ValueError(_("Invalid application status change"))
```
karisal-anders, line 30: For uniformity perhaps having the "is not None" handling would be good to have in both if statements or not in either. To me it looks it's not needed at all because cls.values does not contain None and thus "if new status is not None and new_status not in cls.values" is equal to "new_status not in cls.values".

karisal-anders: Hmm... come to think of Zen of Python:
> Explicit is better than implicit

So having `if new_status is not None` combined with the default value assignment `new_status=None` in the function declaration makes it *explicit that the default value is not accepted*. Because of this argument I accept this code as it is.

backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -890,9 +890,12 @@ def test_application_status_change_as_applicant(
...
         (ApplicationStatus.HANDLING, ApplicationStatus.REJECTED, 200),
         (ApplicationStatus.HANDLING, ApplicationStatus.CANCELLED, 200),
+        (ApplicationStatus.ACCEPTED, ApplicationStatus.HANDLING, 200),
+        (ApplicationStatus.REJECTED, ApplicationStatus.HANDLING, 200),
         (ApplicationStatus.RECEIVED, ApplicationStatus.DRAFT, 400),
         (ApplicationStatus.ACCEPTED, ApplicationStatus.RECEIVED, 400),
         (ApplicationStatus.CANCELLED, ApplicationStatus.ACCEPTED, 400),
+        (ApplicationStatus.CANCELLED, ApplicationStatus.HANDLING, 400),
         (ApplicationStatus.REJECTED, ApplicationStatus.DRAFT, 400),
     ],
```
karisal-anders, line 900: How about adding the rest of the cases going back to HANDLING state:

```python
(ApplicationStatus.DRAFT, ApplicationStatus.HANDLING, <status code>),
(ApplicationStatus.RECEIVED, ApplicationStatus.HANDLING, <status code>),
(ApplicationStatus.HANDLING, ApplicationStatus.HANDLING, <status code>),
```

karisal-anders (approved):
LGTM! If you want you can add a few more test cases but you can merge as is also, I leave that up to you.

backend/benefit/applications/enums.py
haavikko: new_status can be None, in the case that status is not being changed.
status on the other hand can not be None
Added comment to clarify

backend/benefit/applications/tests/test_applications_api.py
haavikko: The allowed state transitions are defined in the StatusTransitionValidator, I'm only testing a subset of state transitions to ensure the state machine logic works

PR #758, opened 2022-02-07
Pipeline yaml files

Vitals9367:
Added pipeline config files. Please check if the triggers are the way you need them


nenonja:
You could add these extra excludes because changing only test files should not trigger deploy? 
https://github.com/City-of-Helsinki/yjdh/blob/d671c3143f1ce4a2810729c6e2177f03077ded78/frontend/benefit/handler/azure-pipelines-master.yml#L15-L16

Vitals9367:
Added extra triggers

nenonja (approved):

PR #764, opened 2022-02-08
YJDH-380 | KS-Backend: If activation email not sent return error 500

karisal-anders:
## Description :sparkles:

If activation email is not sent when posting a new youth application
then return HTTP status 500 i.e. internal server error.

Document some of the email sending functions in YouthApplication.

Add tests for testing that the activation email sending failure
returns HTTP status 500.

Add test for testing that the activation link's path, e.g.
/v1/youthapplications/936bf988-f3c2-4279-b3d9-e7729d94c02c/activate/
is found in the activation email body.

Add testing that posting a valid new youth application creates a youth
application.

Add testing that activation email sending failure does not create a
youth application.

## Issues :bug:

YJDH-380


haavikko (approved):
LGTM!

PR #767, opened 2022-02-09
Hl 403/fix calculating duration in months

ngocdh236:
## Description :sparkles:
Copy the method days360 from backend and fix diffMonths method to follow Excel Days360 function


melniiv (approved):
Great!

PR #772, opened 2022-02-09
YJDH-473 | KS-Backend: Try different collations in SchoolListView

karisal-anders:
## Description :sparkles:

Previously SchoolListView used "fi-FI-x-icu" collation for sorting the
school list. This failed in the Platta production environment.

Now SchooListView tries out the following collations in order and uses
the first one that works:
 - fi_FI.UTF8
   - PostgreSQL UTF-8 version
 - Finnish_Swedish_CI_AS_UTF8
   - Azure's UTF-8 version
 - fi-FI-x-icu
   - PostgreSQL fallback
 - None
   - No collation override

It was not tested beforehand whether any of the above collations work in
the Platta production environment. If none of the explicit collations
work then the "No collation override" version is used which should
always work. The worst case scenario is that no collation override is
used, the underlying database has not been configured to use
Finnish UTF-8 collation and thus the school list is sorted incorrectly.

## Issues :bug:

YJDH-473

### DIFF
--- a/backend/kesaseteli/applications/api/v1/views.py
+++ b/backend/kesaseteli/applications/api/v1/views.py
@@ -1,9 +1,11 @@
 import logging
+from functools import cached_property
 
 from django.conf import settings
 from django.core import exceptions
 from django.db import transaction
-from django.db.models import Func
+from django.db.models import F, Func
+from django.db.utils import ProgrammingError
 from django.http import FileResponse, HttpResponse, HttpResponseRedirect, JsonResponse
 from django.utils import translation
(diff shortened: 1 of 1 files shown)


haavikko (approved):
I assume this worked in local environment?
Full speed ahead then!

karisal-anders:
> I assume this worked in local environment? Full speed ahead then!

Yes, there's a `test_schools_list_returns_sorted_collection` test in backend/kesaseteli/applications/tests/test_schools_api.py which test the school list returns a sorted collection and that test passed.

PR #791, opened 2022-02-14
TETP-73: Show TET Youth data from LE

mattikl:
## Description :sparkles:

This PR brings TET Youth back to the review environment. It implements showing data from LinkedEvents test environment in order for us to verify how the data is set.

Note that this is a very quick PoC implementation to enhance our testing capabilities and to start reasoning about TET Youth data logic.


frontend/shared/src/components/hocs/initLocale.tsx
```diff
@@ -10,7 +10,9 @@ const initLocale =
     const { locale } = router;
     const { i18n } = useTranslation();
     React.useEffect(() => {
-      void i18n.changeLanguage(locale ?? DEFAULT_LANGUAGE);
+      if (i18n?.changeLanguage) {
```
mattikl, line 13: When navigating to dynamic route `postings/[id]` (for example https://te-yout-791.test.kuva.hel.ninja/posting/tet%3Aaf7scp4qfy), `i18n.changeLanguage` is undefined on the first render and this caused an exception.

@nenonja Any thoughts?

mattikl: Hmm, that route doesn't work at all in Review, so I need to look at this more deeply.

mattikl: This change is now reverted and static route like https://te-yout-791.test.kuva.hel.ninja/posting?id=tet%3Aaf7sd2c5oe is used in order for us to merge to develop and continue the implementation of TET Youth. The dynamic route is kept so that further investigate the problem, because the dynamic route is more ideal for us as it gives us the possibility to prerender pages and results in more elegant URLs.

pekkvil (approved):

PR #794, opened 2022-02-14
Updated triggers

Vitals9367:
Commented out - frontend/* trigger, it seems to trigger all pipelines quite often, causes a big CPU usage spike. Is it needed?


frontend/tet/admin/azure-pipelines-devtest.yml
```diff
@@ -7,7 +7,7 @@ trigger:
   paths:
     include:
     - frontend/tet/admin/**
-    - frontend/*
+#    - frontend/*
```
nenonja, line 10: I think it should trigger all services when
- `frontend/*` (something under frontend folder changes (ike package.json)
- `frontend/shared/*` ( any common code changes)

Vitals9367: Alright was concerned about Node CPU spikes when all pipelines start running at the same time, but if its needed lets leave it like it is then.

PR #796, opened 2022-02-14
HL-435,429,439: Accept, reject and cancel handler applications

melniiv:
## Description :sparkles:
- Handler can complete the application by accepting, rejecting and cancelling
- Handler can close the application and be redirected to the main page
- Notification views are implemented when actions are completed

Accept and cancel:
image

Cancel dialog:
image

Success notification view:
image


ngocdh236:
For cancelling application, is there a reason why we don't have the same pattern as when accept / reject application: we setHandledApplication in context with status cancel and log message? Then we can reuse the onDone method, I think it will be more consistent also..

melniiv:
yes. the problem is with Dialog component, which does not allow to use the Context. HDS should fix the component first. The initial idea was to reuse the onDone, but its a workaround now.

ngocdh236 (approved):

PR #798, opened 2022-02-15
YJDH-435 | KS-Backend: Implement HandlerPermission class

karisal-anders:
## Description :sparkles:

User has permission to act as a handler for youth applications / youth
summer vouchers if NEXT_PUBLIC_MOCK_FLAG setting is set or user is an
active staff or superuser user. This depends on the user's staff status
being equal to its controller status which should have been updated by
HelsinkiAdfsAuthCodeBackend.authenticate function.

Change to shared code:
 - HelsinkiAdfsAuthCodeBackend.authenticate:
   - Don't just set user.is_staff if the user belongs to any of the
     ADFS controller groups but additionally unset user.is_staff if they
     don't belong to any of them anymore. This makes user.is_staff equal
     to the user's controller status after ADFS authentication.

## Issues :bug:

YJDH-435


haavikko (approved):
LGTM!

PR #802, opened 2022-02-16
YJDH-417: application can be opened on handler-ui

nenonja:
## Description :sparkles:
- refactoring shared header-component: (use useLocale-hook inside the component intead of passing the locale object)
- show youth application data on handler-ui
- component / browser-tests
- no action buttons yet


frontend/benefit/applicant/src/components/header/Header.tsx
```diff
@@ -14,7 +14,7 @@ import { $CustomMessagesActions } from './Header.sc';
 import { useHeader } from './useHeader';
 
 const Header: React.FC = () => {
-  const { t, locale, languageOptions, hasMessenger, handleLanguageChange } =
```
nenonja, line 17: useLocale is used inside `shared/Header` component so no need to pass it anymore.

frontend/kesaseteli/handler/src/pages/_app.tsx
```diff
@@ -14,7 +13,7 @@ import BaseApp from 'shared/components/app/BaseApp';
 const App: React.FC<AppProps> = (appProps: AppProps) => (
   <BackendAPIProvider baseURL={getBackendDomain()}>
     <QueryClientProvider client={createQueryClient()}>
-      <BaseApp header={<Header />} footer={<Footer />} {...appProps} />
+      <BaseApp header={<Header />} {...appProps} />
```
nenonja, line 16: no footer on handler-ui

frontend/shared/src/components/header/Header.tsx
```diff
@@ -17,13 +17,12 @@ export type HeaderProps = {
   titleUrl?: string;
   skipToContentLabel?: string;
   menuToggleAriaLabel?: string;
-  locale: string;
-  languages: OptionType<string>[];
+  languages?: OptionType<string>[];
```
nenonja, line 20: some services (as handler) does not support translations so let's make languages optional

frontend/kesaseteli/employer/src/pages/thankyou.tsx
```diff
@@ -27,14 +28,14 @@ const ThankYouPage: NextPage = () => {
...
-  if (!applicationId) {
+  if (!isRouterLoading && !applicationId) {
     void goToPage('/');
     return null;
   }
 
   if (applicationQuery.isSuccess) {
     const application = applicationQuery.data;
-    if (application.status === 'draft') {
+    if (applicationId && application.status === 'draft') {
```
melniiv, line 38: still could be moved to const enum

melniiv (approved):
Looks good!

frontend/kesaseteli/employer/src/pages/thankyou.tsx
nenonja: Perhaps it's just me but I dont like enumerations but prefer the enumerated type instead. Because I think it's more readable and tslint will accept only valid values anyway, also IDE supports auto-suggestions
Screenshot 2022-02-21 at 10 40 13 AM
Screenshot 2022-02-21 at 10 40 25 AM
Screenshot 2022-02-21 at 10 41 42 AM

nenonja: There is still situations where enums are not redundant: https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant for example iterating all the possible values, but I havent seen the use case yet in the codebase.

PR #803, opened 2022-02-16
TETP-32 employer can preview posting

pekkvil:


frontend/tet/shared/src/util/getFormattedDate.ts
```diff
@@ -0,0 +1,10 @@
+const getFormattedDate = (date: string): string => {
```
nenonja, line 1: I think that calling `convertToUIDateFormat(string) || ''` method from `date.utils.ts` (in shared code) does effectively the same. It's better to collect all date utils to the same shared file, as they are quite same accross all the projects.

pekkvil: True, thanks!

frontend/tet/admin/src/components/editor/SelectionGroup.tsx
```diff
@@ -10,17 +10,18 @@ type Props = {
...
   const { control, setValue, getValues, clearErrors } = useFormContext<TetPosting>();
-  const checkboxChangeHandler = (optionId: string) => {
+  const checkboxChangeHandler = (option: OptionType) => {
     const values = getValues(fieldId);
     if (Array.isArray(values)) {
-      let list = [...values];
-      const index = list.indexOf(optionId);
+      let list: OptionType[] = [...values];
+      const index = list.findIndex((item) => item.value === option.value);
       if (index === -1) {
```
nenonja, line 23: Isn't this same as
```
if (list.every((item) => item.value !== option.value)) {
   list = list.concat(option);
}
```
?

frontend/tet/admin/src/components/editor/SelectionGroup.tsx
```diff
@@ -10,17 +10,18 @@ type Props = {
...
-const SelectionGroup: React.FC<Props> = ({ fieldId, label, options, required }) => {
+const SelectionGroup: React.FC<Props> = ({ fieldId, label, options, required, rules }) => {
   const { control, setValue, getValues, clearErrors } = useFormContext<TetPosting>();
-  const checkboxChangeHandler = (optionId: string) => {
+  const checkboxChangeHandler = (option: OptionType) => {
     const values = getValues(fieldId);
     if (Array.isArray(values)) {
-      let list = [...values];
-      const index = list.indexOf(optionId);
+      let list: OptionType[] = [...values];
```
nenonja, line 21: `const list`

frontend/tet/admin/src/components/editor/companyInfo/CompanyInfo.tsx
```diff
@@ -25,15 +24,19 @@ const CompanyInfo: React.FC = () => {
...
     ? keywordsResults.data.map(
         (keyword) =>
           ({
+            name: keyword.name?.fi,
             label: `${keyword.name.fi}, ${keyword.street_address?.fi ? keyword.street_address.fi : ''}, ${
               keyword.postal_code ? keyword.postal_code : ''
             }`,
             value: keyword['@id'],
-          } as OptionType),
+            street_address: keyword.street_address?.fi ? keyword.street_address.fi : '',
```
nenonja, line 36: `street_address: keyword.street_address?.fi ?? '',` is shorter

frontend/tet/admin/src/components/editor/companyInfo/CompanyInfo.tsx
```diff
@@ -25,15 +24,19 @@ const CompanyInfo: React.FC = () => {
 
   const keywordsResults = useQuery(['keywords', addressSearch], () => getAddressList(addressSearch));
 
-  const keywords: OptionType[] = keywordsResults.data
+  const keywords: LocationType[] = keywordsResults.data
```
nenonja, line 27: map methods can be heavy to calculate every render so I would suggest to wrap it into `React.useMemo` -hook

nenonja, line 27: map methods can be heavy to calculate every render so I would suggest to wrap it into `React.useMemo` -hook

frontend/tet/admin/src/components/editor/EditById.tsx
```diff
@@ -1,25 +1,47 @@
...
+    );
+  }
+  return (
+    <>
+      <Container>
+        <BackButton />
+        <$HeadingContainer>
+          <$Heading>{title}</$Heading>
+        </$HeadingContainer>
+        <Editor initialValue={isInitialRender ? data : tetPosting} />
```
nenonja, line 41: I think you could use `initialData` or `placeholderData` of [react-query](https://react-query.tanstack.com/guides/initial-query-data)

frontend/tet/admin/src/store/PreviewContext.tsx
```diff
@@ -0,0 +1,114 @@
...
+  contact_last_name: '',
+  date_published: null,
+  contact_language: '',
+  contact_phone: '',
+  keywords: [],
+  keywords_working_methods: [],
+  keywords_attributes: [],
+};
+
+export const PreviewContext = React.createContext<PreviewContextObj>({
```
nenonja, line 60: Perhaps I'm not just big fan of React.context so I'm thinking do we absolutely need this context? What happens if thepage is refreshed? Does the data just get lost or is it retained from backend + react-query? If it's derived from react-query data, then you might not need it but you could use select function to derive value from useQuery.
Some ideas: https://tkdodo.eu/blog/react-query-as-a-state-manager
I have actually used just query params or `useLocalstorage` to save user settings or that part of application state which is not synced to backend. If losign data on page refresh is not a problem then useContext is probably just fine :)

frontend/tet/admin/src/pages/copystatic.tsx
```diff
@@ -0,0 +1,35 @@
...
+    delete postingData.id;
+    return <EditById title={t('common:editor.copyTitle')} data={postingData} />;
+  } else {
+    return <PageNotFound />;
+  }
+};
+
+export const getStaticProps: GetStaticProps = getServerSideTranslations('common');
+
+export default CopyStaticPage;
```
mattikl, line 35: Should be exported `withAuth(CopyStaticPage)` so that unauthorized users are redirected to login page.

mattikl (approved):

PR #804, opened 2022-02-16
HL-449: hide draft applications from handler API

haavikko:
Also
* Deny handler's access to applicant applications API
* Fix tests where handler was accessing draft applications
  or using the applicant API
* Force the rest framework to use English error messages in tests


ngocdh236 (approved):

ngocdh236 (commented):
Oh sorry I missed one file, is the `BFIsAuthenticated` supposed to be changed in `/backend/benefit/users/api/v1/views.py` as well?

PR #814, opened 2022-02-21
YJDH-253: Add static forbidden page for handler application

nenonja:
## Description :sparkles:
Add static forbidden page for kesäseteli handler application

### DIFF
--- a/frontend/kesaseteli/handler/public/locales/fi/common.json
+++ b/frontend/kesaseteli/handler/public/locales/fi/common.json
@@ -14,6 +14,10 @@
     "pageNotFoundLabel": "Sivua ei löytynyt",
     "pageNotFoundContent": "Voi ei, jokin meni nyt valitettavasti vikaan. Kokeile mennä <lnk>tästä</lnk> etusivun kautta takaisin hakemukseen."
   },
+  "403Page": {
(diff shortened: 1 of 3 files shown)


frontend/kesaseteli/handler/public/locales/fi/common.json
```diff
@@ -14,6 +14,10 @@
     "pageNotFoundLabel": "Sivua ei löytynyt",
     "pageNotFoundContent": "Voi ei, jokin meni nyt valitettavasti vikaan. Kokeile mennä <lnk>tästä</lnk> etusivun kautta takaisin hakemukseen."
   },
+  "403Page": {
+    "forbiddenPageLabel": "Sinulla ei ole oikeuksia",
+    "forbiddenPageContent": "Oikeutesi eivät riitä avaamaan hakemusta. Ota yhteyttä <a>helpdeskiin</a>."
+  },
   "error": {
```
ngocdh236, line 21: Hi, should this also be in en and sv translations..?

nenonja: Handler supports only finnish language at least for now. It might change in the future, that'a why finnish texts are not hard coded to components and we use translation library already.

ngocdh236 (approved):

PR #821, opened 2022-02-23
Hl 470/calculate apprenticeship

ngocdh236:
## Description :sparkles:
Implement UI and API integration for adding, removing apprenticeship calculation

## Screenshots :camera_flash:
Screenshot 2022-02-23 at 10 05 23
Screenshot 2022-02-23 at 10 07 21

### DIFF
--- a/frontend/benefit/handler/public/locales/sv/common.json
+++ b/frontend/benefit/handler/public/locales/sv/common.json
@@ -230,6 +230,10 @@
       "workTimePercent": {
         "label": "Työaika %",
         "placeholder": "Työaika %"
+      },
+      "monthlyAmount": {
+        "label": "Koulutuskorvaus €/kk",
+        "placeholder": "Koulutuskorvaus €/kk"
       }
     },
     "notifications": {
@@ -269,7 +273,11 @@
       "tableRowHeader": "Helsinki-lisä / kk (enintään {{amount}} €)",
       "tableRowValue": "{{amount}} €",
       "tableTotalHeader": "Helsinki-lisä yhteensä",
-      "manualInput": "Manuaalinen syöttö"
+      "manualInput": "Manuaalinen laskelma",
+      "apprenticeshipCompensation": "Oppisopimuksesta maksettava korvaus",
+      "apprenticeshipPeriod": "Koulutuskorvaus myönnetty ajalle",
+      "add": "Liisä",
+      "remove": "Poista"
     }
   },
   "organizationTypes": {
--- a/frontend/benefit/handler/public/locales/sv/common.json
+++ b/frontend/benefit/handler/public/locales/sv/common.json
@@ -276,6 +276,7 @@
       "manualInput": "Manuaalinen laskelma",
       "apprenticeshipCompensation": "Oppisopimuksesta maksettava korvaus",
       "apprenticeshipPeriod": "Koulutuskorvaus myönnetty ajalle",
+      "apprenticeshipPeriodWithDiff": "Koulutuskorvaus myönnetty ajalle ({{period}} kk)",
       "add": "Liisä",
       "remove": "Poista"
     }
--- a/frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
+++ b/frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
@@ -10,7 +10,15 @@ import {
   PaySubsidy,
   SalaryBenefitCalculatorViewProps,
 } from 'benefit/handler/types/application';
-import { Button, DateInput, Select, TextArea, TextInput } from 'hds-react';
+import {
+  Button,
+  DateInput,
+  IconMinusCircle,
+  IconPlusCircle,
+  Select,
+  TextArea,
+  TextInput,
+} from 'hds-react';
 import * as React from 'react';
 import { $ViewField } from 'shared/components/benefit/summaryView/SummaryView.sc';
 import DateInputWithSeparator from 'shared/components/forms/fields/dateInputWithSeparator/DateInputWithSeparator';
@@ -48,6 +56,11 @@ const SalaryBenefitCalculatorView: React.FC<
     getPaySubsidyPercentageSelectValue,
     isManualCalculator,
     changeCalculatorMode,
+    newTrainingCompensation,
+    setNewTrainingCompensation,
+    addNewTrainingCompensation,
+    removeTrainingCompensation,
+    isDisabledAddTrainingCompensationButton,
   } = useSalaryBenefitCalculatorData(data);
   const {
     t,
@@ -327,6 +340,7 @@ const SalaryBenefitCalculatorView: React.FC<
                       );
                     }}
                     value={convertToUIDateFormat(item.endDate)}
+                    style={{ paddingRight: `${theme.spacing.s}` }}
                   />
                 </$GridCell>
               </>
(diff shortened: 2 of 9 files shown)


frontend/benefit/handler/src/hooks/useHandlerReviewActions.ts
```diff
@@ -108,6 +109,16 @@ const useHandlerReviewActions = (
         }))
       : undefined;
 
+    const trainingCompensations = values.trainingCompensations
```
nenonja, line 112: `values.trainingCompensations?.map` is shorter. Usually also derived values like mapping could be wrapped with `const trainingCompensations = React.useMemo(() => ...values.trainingCompensations?.map(...),[values.trainingCompensations]);` to avoid re-rendering.

frontend/shared/src/utils/date.utils.ts
```diff
@@ -174,7 +174,7 @@ export const diffMonths = (
 export const getCorrectEndDate = (
   startDate: string,
   endDate: string
-): string | undefined | Date => {
+): string => {
   if ((parseDate(startDate) ?? 0) > (parseDate(endDate) ?? 0)) return startDate;
```
nenonja, line 178: screams for unit tests

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
```diff
@@ -372,10 +386,181 @@ const SalaryBenefitCalculatorView: React.FC<
           invalid={!!getErrorMessage(fields.endDate.name)}
           aria-invalid={!!getErrorMessage(fields.endDate.name)}
           errorText={getErrorMessage(fields.endDate.name)}
-          style={{ paddingRight: `${theme.spacing.xs}` }}
+          style={{ paddingRight: `${theme.spacing.s}` }}
         />
       </$GridCell>
 
+      {!isManualCalculator && (
```
nenonja, line 393: I would suggest to put it into another component / refactor somehow. This is superlong component already.

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
```diff
@@ -372,10 +386,181 @@ const SalaryBenefitCalculatorView: React.FC<
...
+        <>
+          <$GridCell $colStart={1} $colSpan={11}>
+            <$CalculatorText
+              style={{ marginTop: theme.spacing.m, fontWeight: 'bold' }}
+            >
+              {t(`${translationsBase}.apprenticeshipCompensation`)}
+            </$CalculatorText>
+          </$GridCell>
+
+          {formik.values.trainingCompensations &&
```
nenonja, line 403: `formik.values.trainingCompensations?.length > 0 && ...`

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
```diff
@@ -372,10 +386,181 @@ const SalaryBenefitCalculatorView: React.FC<
...
-          style={{ paddingRight: `${theme.spacing.xs}` }}
+          style={{ paddingRight: `${theme.spacing.s}` }}
         />
       </$GridCell>
 
+      {!isManualCalculator && (
+        <>
+          <$GridCell $colStart={1} $colSpan={11}>
+            <$CalculatorText
+              style={{ marginTop: theme.spacing.m, fontWeight: 'bold' }}
```
nenonja, line 397: $CalculatorText is styled component so I would prefer to put styles inside the component parameterized like 
`<<$CalculatorText bold />` or prefer `css` atribute instead of `style .

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
```diff
@@ -372,10 +386,181 @@ const SalaryBenefitCalculatorView: React.FC<
...
+                ),
+              })}
+            </$CalculatorText>
+          </$GridCell>
+          <$GridCell $colStart={1} $colSpan={1}>
+            <TextInput
+              id={fields.monthlyAmount.name}
+              name={fields.monthlyAmount.name}
+              onChange={(e) =>
+                setNewTrainingCompensation({
```
nenonja, line 495: setNewTrainingCompensation((prevValue) => ({...prevValue, monthlyAmount: e.target.value}));` is more safe. Perhaps I would consider also using useReducer here as the state is complex object:
https://reactjs.org/docs/hooks-reference.html#:~:text=useReducer%20is%20usually%20preferable%20to,dispatch%20down%20instead%20of%20callbacks.

frontend/benefit/handler/public/locales/sv/common.json
```diff
@@ -269,7 +273,12 @@
       "tableRowHeader": "Helsinki-lisä / kk (enintään {{amount}} €)",
       "tableRowValue": "{{amount}} €",
       "tableTotalHeader": "Helsinki-lisä yhteensä",
-      "manualInput": "Manuaalinen syöttö"
+      "manualInput": "Manuaalinen laskelma",
+      "apprenticeshipCompensation": "Oppisopimuksesta maksettava korvaus",
+      "apprenticeshipPeriod": "Koulutuskorvaus myönnetty ajalle",
+      "apprenticeshipPeriodWithDiff": "Koulutuskorvaus myönnetty ajalle ({{period}} kk)",
+      "add": "Liisä",
```
nenonja, line 280: `Lisää`

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
```diff
@@ -372,10 +386,181 @@ const SalaryBenefitCalculatorView: React.FC<
...
+
+                <$GridCell $colStart={3} $colSpan={3}>
+                  <$CalculatorText>
+                    {t(`${translationsBase}.apprenticeshipPeriod`)}
+                  </$CalculatorText>
+                </$GridCell>
+              </>
+            )}
+
+          {formik.values.trainingCompensations?.map((item) => (
```
nenonja, line 420: using `key` somewhere inside the map is adviced to make better performance: https://reactjs.org/docs/lists-and-keys.html#keys

nenonja (commented):
Some technical comments only, I have no idea about the business case, so someone else can review this from the business point of view..

frontend/benefit/handler/src/hooks/useHandlerReviewActions.ts
ngocdh236: I have changed to `values.trainingCompensations?.length > 0`, this is because I don't need to use values from `trainingCompensations`, just need to check if they are available

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
ngocdh236: I will get error `Object is possibly 'undefined'.` (trainingCompensations?: TrainingCompensation[] | undefined) if used like this..

frontend/shared/src/utils/date.utils.ts
ngocdh236: I have now added unit test

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
ngocdh236: This is now fixed

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
ngocdh236: This is now fixed

frontend/benefit/handler/public/locales/sv/common.json
ngocdh236: This is now fixed

frontend/benefit/handler/src/hooks/useHandlerReviewActions.ts
nenonja: Not sure what you mean... the quesition mark would make it anyway shorter whatever is the map function:
Screenshot 2022-02-28 at 12 35 11 PM
useMemo might help to avoid unnecessary re-renders.

frontend/shared/src/utils/__tests__/date.utils.test.ts
```diff
@@ -39,4 +45,11 @@ describe('dates', () => {
       expect(diffMonthsResult).toBeLessThan(0);
     });
   });
+
+  describe('getCorrectEndDate', () => {
+    it('should be expected result', () => {
+      expect(getCorrectEndDate('31.12.2022', '1.1.2021')).toBe('31.12.2022');
```
nenonja, line 51: for 100% path coverage you could also add two more checks with invalid string values.

nenonja: Third case would be that both valeus are invalid, it seem to return the first value even if it's invalid. Perhaps it could return undefined or null instead. That's a minor thing anyway.

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
ngocdh236: This is now fixed but not yet using useReducer, imo useReducer is not really necessary here because we only have 1 action type which is to update the newTraining.. object, in addition we use TypeScript so there will be more typing to add

frontend/shared/src/utils/__tests__/date.utils.test.ts
ngocdh236: This is now fixed

frontend/benefit/handler/src/components/applicationReview/salaryBenefitCalculatorView/SalaryBenefitCalculatorView.tsx
ngocdh236: I moved the manual calculator UI to a separate component

nenonja (commented):
There is some linter errors in checks. Otherwise looks good now!

PR #824, opened 2022-02-24
Fix applicant cannot choose 1.1 as start date

ngocdh236:
## Description :sparkles:
Fix applicant cannot choose 1.1 as start date

### DIFF
--- a/frontend/benefit/applicant/src/components/applications/forms/application/step2/utils/validation.ts
+++ b/frontend/benefit/applicant/src/components/applications/forms/application/step2/utils/validation.ts
@@ -16,7 +16,10 @@ import startOfYear from 'date-fns/startOfYear';
 import { FinnishSSN } from 'finnish-ssn';
 import { TFunction } from 'next-i18next';
 import { NAMES_REGEX, PHONE_NUMBER_REGEX } from 'shared/constants';
-import { convertToUIDateFormat, parseDate } from 'shared/utils/date.utils';
+import {
+  convertToUIDateFormat,
+  validateDateIsFromCurrentYearOnwards,
+} from 'shared/utils/date.utils';
 import { getNumberValue } from 'shared/utils/string.utils';
 import * as Yup from 'yup';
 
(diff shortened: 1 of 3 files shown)


frontend/benefit/applicant/src/components/applications/forms/application/step2/utils/validation.ts
```diff
@@ -76,7 +75,7 @@ export const getValidationSchema = (t: TFunction): Yup.SchemaOf<Step2> =>
         }),
         test: (value = '') => {
           const date = parseDate(value);
-          return date ? isAfter(date, startOfYear(new Date())) : false;
```
nenonja, line 79: Looks okay, but this would be a great moment to add unit tests for the validation too. Not necessary though. It seems that the test method should be first extracted here to be able to test it?

ngocdh236: I have moved to util and written unit test for this now

frontend/shared/src/utils/__tests__/date.utils.test.ts
```diff
@@ -39,4 +45,19 @@ describe('dates', () => {
       expect(diffMonthsResult).toBeLessThan(0);
     });
   });
+
+  describe('validateDateIsFromCurrentYearOnwards', () => {
+    it('should return false', () => {
+      const validateDateIsFromCurrentYearOnwardsResult =
+        validateDateIsFromCurrentYearOnwards('31.12.2021');
```
nenonja, line 52: `31.12.${new Date().getFullYear()-1}` could more of an edge value in the future too :)
Also If you want to get path coverage 100% then there could be one more test with invalid string value which should return false.

nenonja (approved):
One minor comment but approved

frontend/shared/src/utils/__tests__/date.utils.test.ts
ngocdh236: Thank you, I have updated the test, if it's ok then I will merge

PR #832, opened 2022-02-28
YJDH-434: all but nextjs 12 update :)

nenonja:
## Description :sparkles:
It seems that after painstakingly work to make the all the services to deploy with nextjs 12, it turns out that styledComponents `css` prop does not work. After researching a bit, the current nextjs 12 version transforms only the `core` features of the styledComponents but not all: 
https://nextjs.org/blog/next-12-1#improved-swc-support 
https://github.com/vercel/next.js/issues/30802

It was a bit unfortunate to find it out so late, but there is still very good refactorings in this pr:
- Use only one dockerfile from the root folder which has parameterizable folder structure (with docker ARGS)
- Only one yarn.lock file in the root
- tsconfig, eslint and nextjs configs have base configurations on the root folder, which are overridden by each service only by configurations that alterns from default. This means a lot less copy-paste.
- projects can no longer import (accidentally) code from other projects except shared.
- benefit/applicant and benefit/handler has mixed up react-i18next and next-i18next packages. It's now fixed.
- react-query client was create inside app in some services which is can cause suprising errors. This is now fixed.

PS: Nextjs 12 update is now forked to the other pr: https://github.com/City-of-Helsinki/yjdh/pull/878


.github/workflows/bf-review.yml
```diff
@@ -77,6 +81,10 @@ jobs:
         env:
           DOCKER_BUILD_ARG_NEXT_PUBLIC_BACKEND_URL: ${{ env.NEXT_PUBLIC_BACKEND_URL }}
           DOCKER_BUILD_ARG_NEXT_PUBLIC_MOCK_FLAG: ${{ env.NEXT_PUBLIC_MOCK_FLAG }}
+          DOCKER_BUILD_ARG_PORT: ${{ matrix.port }}
+          DOCKER_BUILD_ARG_PROJECT: ${{ matrix.project }}
+          DOCKER_BUILD_ARG_SERVICE: ${{ matrix.service }}
+          DOCKER_BUILD_PATH: ${{ matrix.port }}
```
Hi-Fi, line 87: Port?

.github/workflows/bf-review.yml
```diff
@@ -77,6 +81,10 @@ jobs:
         env:
           DOCKER_BUILD_ARG_NEXT_PUBLIC_BACKEND_URL: ${{ env.NEXT_PUBLIC_BACKEND_URL }}
           DOCKER_BUILD_ARG_NEXT_PUBLIC_MOCK_FLAG: ${{ env.NEXT_PUBLIC_MOCK_FLAG }}
+          DOCKER_BUILD_ARG_PORT: ${{ matrix.port }}
+          DOCKER_BUILD_ARG_PROJECT: ${{ matrix.project }}
+          DOCKER_BUILD_ARG_SERVICE: ${{ matrix.service }}
```
Hi-Fi, line 86:
```suggestion
          DOCKER_BUILD_ARG_SERVICE: ${{ matrix.service }}
          DOCKER_BUILD_ARG_FOLDER: ${{ matrix.folder }}          
```

Hi-Fi (changes requested):

.github/workflows/bf-review.yml
nenonja: fixed

.github/workflows/bf-review.yml
nenonja: fixed

nenonja:
This is now otherwise ready for review except TET-projects need to address incorrecrt imports and ts errors. @pekkvil @mattikl, can you check it?

frontend/shared/src/components/table/Table.tsx
```diff
@@ -198,8 +198,7 @@ TableProps<D>): React.ReactElement => {
     const headers = [
       ...(canSelectRows ? [selectorCol] : []),
       ...(canSelectOneRow ? [radioSelectorCol] : []),
-      // fix to a hard ts problem: https://github.com/TanStack/react-table/discussions/2664
-      ...columns.map((col) => ({ ...col, accessor: 'id' as keyof D })),
+      ...columns,
```
nenonja, line 201: This was a fix to this build problem: https://github.com/City-of-Helsinki/yjdh/runs/5620854899?check_suite_focus=true
Not sure how to fix the typesceipt error otherwise. Any ideas?
https://github.com/TanStack/react-table/discussions/2664

ngocdh236: I will try to resolve this and get back to you

mattikl (approved):
Reviewed TET-specific changes. Looks good. PLATTA build configs need change after this is merged.

PR #839, opened 2022-03-01
TETP-99: Add AuthProvider and withAuth to pages

mattikl:


frontend/tet/admin/src/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import useUserQuery from 'kesaseteli/employer/hooks/backend/useUserQuery';
```
mikkovihonen, line 1: Is it good practice to import code straight from other subprojects in the monorepo? Should this be refactored as a common component instead?

mattikl: No, this should be imported from `tet/admin/hooks/backend/useUserQuery`. Kesaseteli's `userQuery` doesn't even work if TET Admin if `NEXT_PUBLIC_MOCK_FLAG` is not set.

mikkovihonen (approved):

PR #847, opened 2022-03-03
YJDH-500/YJDH-492: Confirm dialog before accept/reject application + browser tests for accept/reject

nenonja:
## Description :sparkles:
- move confirm dialog from tet/admin project to shared code
- confirm dialog is shown when handler tries to accept or reject an application
- updated component tests
- browser tests to accept and reject application.


mattikl (approved):
TET Admin tested locally, confirming both publishing and deleting works. Good refactoring to make code cleaner as well.

pekkvil (approved):

PR #860, opened 2022-03-08
Backends: Escape string literals in tests' regular expressions

karisal-anders:
## Description :sparkles:

GitHub Code Scanning / CodeQL found "Incomplete regular expression for
hostnames" problems in code in develop to main branch merging on
7th of March, 2022, for an example:
 - "This regular expression has an unescaped '.' before 'microsoft.com',
   so it might match more hosts than expected."

Fix uses of re.compile in tests with string literals by escaping the
input i.e. by changing re.compile(text) to re.compile(re.escape(text)).

## Issues :bug:

`-`


haavikko (approved):
LGTM!

PR #864, opened 2022-03-09
TETP-127: Enforce access control

mattikl:


backend/tet/events/services.py
```diff
@@ -50,6 +50,7 @@ def _is_city_employee(self, user):
 
     def _get_event_and_raise_for_unauthorized(self, user, event_id):
         event = self.client.get_event(event_id)
+        LOGGER.warning(event)
```
karisal-anders, line 60: The logging is to Sentry, right? What does the event here contain? Might it contain sensitive/personally identifiable information which falls under GDPR?

karisal-anders (commented):
Combining formatting only changes and code changes makes code review harder, e.g. ["fix tests"](https://github.com/City-of-Helsinki/yjdh/pull/864/commits/a1e13f042871c1c7af933eaf9407dbcfb59c5fcd) commit contains a mix of formatting changes and code changes. Now one has to be extra careful to spot which changes are formatting only and which contain code changes. It would be nice to have formatting only changes as their own commits—those commits could actually also be removed from git blame by using .git-ignore-blame-revs if you don't squash them, but that's up to you.

backend/tet/events/services.py
```diff
@@ -42,6 +42,7 @@ def _is_city_employee(self, user):
     def _get_event_and_raise_for_unauthorized(self, user, event_id):
         event = self.client.get_event(event_id)
         if self._is_city_employee(user) and not user.is_superuser:
+            LOGGER.warning(event["custom_data"])
```
karisal-anders: May this contain data which falls under GDPR?

backend/tet/events/tests/test_services.py
```diff
@@ -0,0 +1,98 @@
...
+    matcher = re.compile('.*')
+    # We just care that we get a valid event response back, so ADD_EVENT_RESPONSE is okay to use here
+    requests_mock.put(matcher, json=ADD_EVENT_RESPONSE)
+
+    requests_mock.get('http://localhost/event/tet:af7w4jmjla/', json=EVENT_RESPONSE_TESTUSER)
+    requests_mock.get('http://localhost/event/tet:af7wto6nze/', json=EVENT_RESPONSE_OTHERUSER)
+    requests_mock.get('http://localhost/event/tet:af7wtenvii/', json=EVENT_RESPONSE_NO_USER)
+    requests_mock.get('http://localhost/event/tet:af7wjlxpyy/', json=EVENT_RESPONSE_NO_CUSTOM_DATA)
+
+    # LOGGER.warning(EVENT_RESPONSE_NO_CUSTOM_DATA)
```
karisal-anders: Leaving commented out code is IMHO not good practice—if you need the code then it should not be commented out, if you do not need the code then it should be removed.

backend/tet/events/tests/test_services.py
```diff
@@ -0,0 +1,98 @@
...
+
+    with pytest.raises(PermissionDenied):
+        client.update_tet_event('tet:af7wto6nze', ADD_EVENT_PAYLOAD, user)
+
+    with pytest.raises(PermissionDenied):
+        client.update_tet_event('tet:af7wtenvii', ADD_EVENT_PAYLOAD, user)
+
+    # TODO why doesn't this raise?
+    # with pytest.raises(PermissionDenied):
+    #     client.update_tet_event('tet:af7wjlxpyy', ADD_EVENT_PAYLOAD, user)
```
karisal-anders: Hmm... this is an interesting case. A test case which does not pass and you have left it in comments.

In this case I would actually say this is ok as this shows the code which you expect to pass but it doesn't.

Other way to do this might be to add this specific test case which fails as a separate parametrization of a test or a separate test altogether and leave it in uncommented and let it fail. Then it would be explicit that we expect this to pass but it does not. Then one might add a comment into the test so anyone who sees the test failing would understand that this is currently normal behavior although not intended.

One downside of adding a failing test case would be that it would complicate code review because one would have to make a habit of mentioning that this specific test case is ok to fail. That might be circumvented at least partially by putting all such TODO/Failing tests into their own file or prefix/suffix the test cases with a string that would make it clear that these test cases are currently failing and it is ok until they are fixed.

I don't see a clear way to handle these kind of cases so for me leaving this as commented out code in this case is fine.

backend/tet/events/transformations.py
```diff
@@ -25,11 +25,23 @@ def _new_from(obj, keys):
...
 
 
+def _shorten_description(descobj):
+    shortened_obj = {}
+    for lang in descobj.keys():
+        desc = descobj[lang]
+        if len(desc) <= 125:
+            shortened_obj[lang] = desc
+        else:
+            shortened_obj[lang] = f"{desc[:125]}\u2026"
```
karisal-anders, line 35: It would be beneficial for understandability to add something to tell what the \u2026 character is, e.g.:

import unicodedata
...
shortened_obj[lang] = f"{desc[:125]}{unicodedata.lookup('Horizontal ellipsis')}"
or
ellipsis = "\u2026"
shortened_obj[lang] = f"{desc[:125]}{ellipsis}"
or
ellipsis = unicodedata.lookup('Horizontal ellipsis')
shortened_obj[lang] = f"{desc[:125]}{ellipsis}"
or
shortened_obj[lang] = f"{desc[:125]}\u2026"  # \u2026 = …

backend/tet/events/tests/test_services.py
karisal-anders: Actually after consulting a colleague about this I now know that there's [@pytest.mark.xfail](https://docs.pytest.org/en/6.2.x/skipping.html#xfail-mark-test-functions-as-expected-to-fail) which could be used to mark a test case which is expected to fail. So you could uncomment the failing test code and put it into its own test, mark that with @pytest.mark.xfail and leave a comment explaining the current situation and what's to be done.

karisal-anders: [Reason ](https://docs.pytest.org/en/6.2.x/skipping.html#reason-parameter) / [raises](https://docs.pytest.org/en/6.2.x/skipping.html#raises-parameter) parameters to @pytest.mark.xfail could be used to add more clarity.

mattikl: Thanks for the information! This was actually something I never intended to make it to the final state of the PR (for some reason git didn't show that "your branch is one commit ahead", so I had one commit I didn't know I hadn't pushed, sorry for that). But good to learn all this.

backend/tet/events/services.py
```diff
@@ -57,29 +58,30 @@ def _get_event_and_raise_for_unauthorized(self, user, event_id):
...
+            _raise_permission_denied()
 
         return event
 
     def _filter_events_for_user(self, all_events, user):
         """Filters events that `user` is authorized to access"""
         if self._is_city_employee(user):
             if not settings.NEXT_PUBLIC_MOCK_FLAG:
-                if user.is_superuser:
-                    return all_events
```
karisal-anders: Was it your intention to remove the superuser having access to all events here?

mattikl: Yes. It still needs to be specified further whether superuser should have that right.

backend/tet/tet/mock_views.py
```diff
@@ -0,0 +1,41 @@
...
+    @method_decorator(ensure_csrf_cookie)
+    def get(self, request, *args, **kwargs):
+        if not request.user.is_authenticated:
+            user = StaffUserFactory()
+            auth.login(
+                request,
+                user,
+                backend="django.contrib.auth.backends.ModelBackend",
+            )
+        return HttpResponseRedirect(settings.LOGIN_REDIRECT_URL or "/")
```
karisal-anders, line 41: To let you know if you find it useful there's `redirect_to_login_using_request` in backend/shared/shared/common/utils.py which has been taken from https://github.com/django/django/blob/3.2.4/django/contrib/auth/decorators.py#L22-L33

karisal-anders (approved):

PR #876, opened 2022-03-10
HL-478: organization type determination

haavikko:
## Description :sparkles: 

Store organization type code in addition to the Finnish
type name and use that to determine organization type.

## Testing :alembic:
backend/benefit/companies/tests/test_api.py


backend/benefit/applications/tests/conftest.py
```diff
@@ -114,6 +115,7 @@ def association_application(mock_get_organisation_roles_and_create_company):
     application.company = mock_get_organisation_roles_and_create_company
     application.save()
     application.company.company_form = "association"  # TODO: fix with actual value
```
karisal-anders, line 117: Is the TODO on this line still valid?
If it is do you now know the actual value that would fix this?

backend/benefit/companies/tests/test_api.py
```diff
@@ -129,6 +132,8 @@ def test_get_company_from_yrtti(
     )
     company_data = CompanySerializer(company).data
     assert response.data == company_data
+    assert response.data["company_form_code"] == 29
```
karisal-anders, line 135: `assert response.data["company_form_code"] == ASSOCIATION_FORM_CODE` would be more readable.

backend/benefit/helsinkibenefit/settings.py
```diff
@@ -41,6 +41,10 @@
     CSRF_TRUSTED_ORIGINS=(list, []),
     YTJ_BASE_URL=(str, "http://avoindata.prh.fi/opendata/tr/v1"),
     YTJ_TIMEOUT=(int, 30),
+    ASSOCIATION_FORM_CODES=(
+        str,
+        "18,21,25,29,31,32,35,38,39,4,44,45,46,47,50,58,6,71,9,90,94999",
+    ),
```
karisal-anders, line 47: Adding e.g. a comment about what these numbers are or at least where they come from or perhaps a link to documentation from which they can be looked up from would add transparency to this magic number data.

backend/benefit/helsinkibenefit/settings.py
```diff
@@ -90,7 +94,7 @@
     ENABLE_SEND_AUDIT_LOG=(bool, False),
     WKHTMLTOPDF_BIN=(str, "/usr/bin/wkhtmltopdf"),
     DISABLE_AUTHENTICATION=(bool, False),
-    DUMMY_COMPANY_FORM=(str, "OY"),
+    DUMMY_COMPANY_FORM_CODE=(int, 16),
```
karisal-anders, line 97: Importing and using COMPANY_FORM_CODE here instead of 16 would add readability.

backend/shared/shared/yrtti/yrtti_client.py
```diff
@@ -7,6 +7,13 @@
 
 
 class YRTTIClient:
+
+    # The YRTTI API only returns data for associations, so the response does not contain
+    # a field for company form.
+    # Use the YTJ "yritysmuoto" code for associations when creating the Company objects
+    ASSOCIATION_FORM_CODE = 29
```
karisal-anders, line 14: You have now two `ASSOCIATION_FORM_CODE = 29` definitions, one in YRTTIClient and one in factories. Would it be prudent to combine them into one or do you see some reason why not to do that?

karisal-anders (commented):
On the whole looks quite ok though adding an enumeration for the company form codes and using its values instead of the magic numbers and some documentation/link to documentation about their values would add readability, understandability and maintainability. We could have a talk about this in Slack.

karisal-anders (approved):
LGTM!

PR #878, opened 2022-03-10
(Not To be Merged Yet): Yjdh-434 Nextjs 12 update (second try)

nenonja:
## Description :sparkles:
This is cherry-picked nextjs 12 changes from previous pr: https://github.com/City-of-Helsinki/yjdh/pull/832
As mentioned there, nextjs 12 still does not support css-prop of the styled components so we cannot take it into use yet. 
Styles are broken for example here: https://helsinkibenefit-bf-hdlr-878.test.kuva.hel.ninja/reports and basically here and there in every project.
Lets wait for the next release and hope they have fixed it. Issue can be followed here: https://github.com/vercel/next.js/issues/30802

PS: There is a good example of nextjs 12 monorepo: https://github.com/belgattitude/nextjs-monorepo-example
Perhaps it has some ideas that could be implemented sooner or later ( like upgrading to yarn 3)


sirtawast:
Just listing here, might be helpful:
https://www.reddit.com/r/nextjs/comments/ssb3pi/how_do_i_use_styledcomponents_in_nextjs_12/

PR #884, opened 2022-03-11
HL-478: remove application from batch when moving back to handling

haavikko:
## Description :sparkles:
Application.batch field needs to be set to null when moving back to handling

## Testing :alembic:
backend/benefit/applications/tests/test_applications_api.py:test_application_with_batch_back_to_handling

### DIFF
--- a/backend/benefit/applications/tests/test_applications_api.py
+++ b/backend/benefit/applications/tests/test_applications_api.py
@@ -26,7 +26,7 @@ from applications.enums import (
 )
 from applications.models import Application, ApplicationLogEntry, Attachment, Employee
 from applications.tests.conftest import *  # noqa
-from applications.tests.factories import ApplicationFactory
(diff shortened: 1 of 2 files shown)


backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -1153,6 +1153,28 @@ def test_application_accept(
     assert handling_application.calculation.granted_as_de_minimis_aid is True
 
 
+def test_application_with_batch_back_to_handling(
+    request,
+    handler_api_client,
+    decided_application,
+):
+    """
+    granted_as_de_minimis_aid is set at the same time the application is accepted.
```
karisal-anders, line 1162: Is this comment up to date? There is nothing in this test which directly mentions granted_as_de_minimis_aid other than this comment.

haavikko: good catch, fixed

karisal-anders (approved):
LGTM!

PR #886, opened 2022-03-11
TET-134 Accept terms before posting creation and add final translations

pekkvil:


mattikl (approved):
Let's merge and see if we spot anything more that needs to be changed.

PR #889, opened 2022-03-14
YJDH-503 | KS-Backend: Set ENABLE_ADMIN default to False

karisal-anders:
## Description :sparkles:

Turn off Django admin interface by default by setting ENABLE_ADMIN's
default value to False. This makes the default functionality safer.

Fix test_excel_view_get_with_unauthenticated_user test by making it
handle both cases of ENABLE_ADMIN on and off. NOTE: Currently it only
runs one of the cases depending on which value ENABLE_ADMIN is set in
the settings because it was not straightforward to run both cases.

Fix "LookupError: No installed app with label 'admin'." exceptions when
trying to register models with the Django admin site in Kesäseteli and
shared backends by not doing the registering if django.contrib.admin
application is not installed.

## Issues :bug:

YJDH-503


haavikko (approved):
LGTM!

PR #904, opened 2022-03-15
Allow title click handler on base header

pekkvil:


frontend/tet/admin/src/components/header/Header.tsx
```diff
@@ -40,13 +40,18 @@ const Header: React.FC = () => {
...
+  };
+
   return (
     <BaseHeader
       title={t('common:appName')}
       skipToContentLabel={t('common:header.linkSkipToContent')}
       menuToggleAriaLabel={t('common:header.menuToggleAriaLabel')}
       languages={languageOptions}
       onLanguageChange={handleLanguageChange}
+      onTitleClick={titleClickHandler}
```
nenonja, line 54: You can actually use a hook from shared code here: `onTitleClick={useGoToFrontPage()}` and it should do the same.

mattikl (approved):

PR #906, opened 2022-03-15
YJDH-498/YJDH-511: ks-youth: Additional info form

nenonja:
## Description :sparkles:
Loading and showing additional info form for youth-ui.Browser tests are broken here because also broken in develop

## Screenshots :camera_flash:
Screenshot 2022-03-21 at 8 25 42 AM
Screenshot 2022-03-21 at 8 25 57 AM


frontend/kesaseteli/youth/src/pages/additional_info.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <Head>
+        <title>{`${t(`common:appName`)} - ${t(
+          'common:additionalInfo.title'
+        )}`}</title>
+      </Head>
+      <FormProvider {...methods}>
+        {isSuccess &&
+          (() => {
+            switch (applicationStatusQuery.data.status) {
+              case 'additional_information_requested':
```
ngocdh236, line 45: Can these statuses be moved to an enum..?

nenonja: I think it's matter of taste,  `applicationStatusQuery.data.status` is not string type but literal type: https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

I use IntelliJ and at least it can autocomplete / suggest the literal type same way as enum:
Screenshot 2022-03-29 at 9 41 06 AM
Also shows error when typo:
Screenshot 2022-03-29 at 9 46 48 AM

frontend/kesaseteli/youth/src/components/forms/AdditionalInfoForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const register = useRegisterInput<AdditionalInfoFormData>('additionalInfo');
+  const reasons: AdditionalInfoReasonOption[] = React.useMemo(
+    () =>
+      ADDITIONAL_INFO_REASON_TYPE.map((type) => ({
+        name: type,
+        label: t(`common:additionalInfo.reasons.${type}`),
+      })),
+    [t]
+  );
```
ngocdh236, line 33: A discussion: I came across some articles about overusing useMemo, React doc also mentions that useMemo helps to avoid expensive calculations. This is a simple one, so using useMemo might not optimize performance but can use more memory. 1 reference: https://medium.com/swlh/should-you-use-usememo-in-react-a-benchmarked-analysis-159faf6609b7, but there are more you can find

nenonja: You're probably right. I've read about the useMemo/useCallback that it isn't always good to use them but I still had. In this pr was at least one place that I had to use it to make the code work. Anyway it seems that useMemo is good with extensive lists or other extensive calculation, while with useCallback this [article](https://www.alexeyshmalko.com/20210128041144/) says:

> Two legible use cases for useCallback are:

- the child component uses React.memo or shouldComponentUpdate (→ [React.useCallback breaks encapsulation](https://www.alexeyshmalko.com/20210128042138/))
- the callback is used as a dependency in useEffect

nenonja: I still feel a little lazy to touch the code now to remove these useMemos or callbacks, but it's perhaps something to do in the future when refactoring. There was one place where it was needed or rendering didnt work correctly so removing all now from this pr probably would get that problem back for now

ngocdh236: yeah just want to have a discussion with you about this, thank you for the insight!

ngocdh236 (approved):

PR #925, opened 2022-03-20
HL-514: change CSRF_COOKIE_NAME to avoid conflict with Tunnistamo

haavikko:
## Description :sparkles:
[mikkovihonen](https://app.slack.com/team/U017Z6SEH70) [Mar 17th at 3:11 PM]
@here Does anyone have any clue why csrftoken would be set two times, first for .[hel.fi](http://hel.fi/) domain and the latter for .[arodevtest.hel.fi](http://arodevtest.hel.fi/)? We noticed that two csrftokens on a call will cause an error which will disappear once the “wrong” csrftoken expires. I witnessed this yesterday in Kesäseteli and today in TET project.


karisal-anders (approved):
LGTM!

PR #930, opened 2022-03-21
HL-247: localized errors for IBAN fields

haavikko:
## Description :sparkles:
The djangol
## Testing :alembic:
backend/benefit/applications/tests/test_applications_api.py:test_application_post_invalid_data


backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -352,10 +362,14 @@ def test_application_post_invalid_data(api_client, application):
     data["status"] = "foo"  # invalid value
     data["bases"] = ["something_completely_different"]  # invalid value
     data["applicant_language"] = None  # non-null required
+    data["company_bank_account_number"] = "FI91 4008 0282 0002 02"  # invalid number
 
     data[
         "company_contact_person_phone_number"
     ] = "+359505658789"  # Invalid country code
+
+    # with translation.override(language):
```
karisal-anders, line 371: Was this commented out code left here by accident?

backend/benefit/common/localized_iban_field.py
```diff
@@ -0,0 +1,29 @@
...
+            super(LocalizedIBANValidator, self).__call__(value)
+        except ValidationError as e:
+            raise ValidationError(_("Invalid IBAN")) from e
+
+
+class LocalizedIBANField(IBANField):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # switch the validator
+        self.validators[:] = [
```
karisal-anders, line 24: What is the rationale for using [:] after the self.validators? What does it do here?

ngocdh236 (dismissed):

backend/benefit/applications/tests/test_applications_api.py
haavikko: Good catch, fixed. The language is determined by accept-language header during request processing so activating translation before making the request didn't work.

backend/benefit/common/localized_iban_field.py
haavikko: Added comment.
self.validators is a read-only property so it can't be assigned to. The property value is a list that is mutable so we can replace the contents of the list with the slice assignment. Also, this is a good way of avoiding list modification while iterating over it.

karisal-anders (approved):

PR #931, opened 2022-03-22
HL-501: ahjo csv export formatting changes

haavikko:
## Description :sparkles:
* Add columns for "Siirrettävä ahjo-rivi" for easier processing
* Finnish texts in colunns "Haettava lisä" and "Työnantajan tyyppi"
* move "Hyväksymisen/hylkäyksen/peruutuksen syy" to a more logical place

## Testing :alembic:
backend/benefit/applications/tests/test_applications_csv_report.py

### DIFF
--- a/backend/benefit/applications/services/applications_csv_report.py
+++ b/backend/benefit/applications/services/applications_csv_report.py
@@ -1,3 +1,4 @@
+from applications.enums import BenefitType
 from applications.services.csv_export_base import (
     CsvColumn,
     CsvExportBase,
@@ -5,7 +6,6 @@ from applications.services.csv_export_base import (
     nested_queryset_attr,
 )
 from django.utils import translation
-from django.utils.translation import gettext
 
 
 def CsvDefaultColumn(*args, **kwargs):
@@ -34,6 +34,13 @@ def get_export_notes(application):
     return ", ".join(notes)
 
 
+def format_datetime(value):
+    if value:
+        return value.isoformat()
+    else:
+        return ""
+
+
 def format_bool(value):
     if value is None:
         return "ei valintaa"
@@ -43,6 +50,21 @@ def format_bool(value):
         return "ei"
     else:
         raise ValueError(f"Invalid value: {value}")
+
+
+def current_ahjo_row_field_getter(field_name):
+    def getter(item):
+        if rows := list(item.ahjo_rows):
+            if item.application_row_idx - 1 < len(rows):
+                # application_row_idx is 1-based
+                return getattr(rows[item.application_row_idx - 1], field_name)
+        return ""
+
+    return getter
+
+
+def get_benefit_type_label(benefit_type):
+    return str(BenefitType(benefit_type).label)
 
 
 class ApplicationsCsvService(CsvExportBase):
@@ -64,7 +86,7 @@ class ApplicationsCsvService(CsvExportBase):
             CsvColumn("Hakemusnumero", "application_number"),
             CsvColumn("Hakemusrivi", "application_row_idx"),
             CsvColumn("Hakemuksen tila", "status"),
-            CsvDefaultColumn("Haettava lisä", "benefit_type", gettext),
+            CsvDefaultColumn("Haettava lisä", "benefit_type", get_benefit_type_label),
             CsvDefaultColumn("Haettu alkupäivä", "start_date"),
             CsvDefaultColumn("Haettu päättymispäivä", "end_date"),
             CsvColumn("Työnantajan tyyppi", get_organization_type),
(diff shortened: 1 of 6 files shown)


backend/benefit/applications/services/applications_csv_report.py
```diff
@@ -45,6 +45,20 @@ def format_bool(value):
         raise ValueError(f"Invalid value: {value}")
 
 
+def get_current_ahjo_row_field(field_name):
```
karisal-anders, line 48: The function name is misleading as it does not get the current ahjo row field but a getter function to get the current ahjo row field. How about `current_ahjo_row_field_getter(field_name)`, `ahjo_row_field_getter(field_name)`  or similar?

backend/benefit/applications/services/applications_csv_report.py
```diff
@@ -193,13 +204,44 @@ def CSV_COLUMNS(self):
...
+            ),
             CsvDefaultColumn("Päättäjän nimike", "batch.decision_maker_title"),
             CsvDefaultColumn("Päättäjän nimi", "batch.decision_maker_name"),
             CsvDefaultColumn("Päätöspykälä", "batch.section_of_the_law"),
             CsvDefaultColumn("Päätöspäivä", "batch.decision_date"),
             CsvDefaultColumn("Asiantarkastajan nimi", "batch.expert_inspector_name"),
             CsvDefaultColumn("Asiantarkastajan email", "batch.expert_inspector_email"),
+            # In case there are multiple rows per application, always have the nth ahjo row
+            # in the same column.
+            # The row data here comes from calculation.ahjo_rows[application_row_idx]
```
karisal-anders, line 218: Is it so that application_row_idx starts from 1 and not from 0? Looking at the getter function that would make sense. The comment here is slightly misleading or at least a bit ambiguous because it's referring to dereferencing an array with the ahjo_rows[application_row_idx] instead of ahjo_rows[application_row_idx - 1]. Maybe a comment here and/or in the getter function about the application_row_idx starting from 1 rather than from 0 and fixing the used index here would do the trick?

haavikko: fixed comment

backend/benefit/setup.cfg
```diff
@@ -7,6 +7,7 @@ ignore = E309
 exclude = migrations,snapshots
 max-line-length = 120
 max-complexity = 10
+ignore = E203, W503
```
karisal-anders, line 10: Having a comment explaining why this is done could be clarifying like e.g. in https://github.com/City-of-Helsinki/yjdh/blob/develop/backend/shared/setup.cfg#L8-L11
```python
# E203 and W503 are disabled and W504 enabled for compatibility with Black.
# For details see "Why are Flake8's E203 and W503 violated?" in Black FAQ at
# https://black.readthedocs.io/en/stable/faq.html
```

backend/shared/precommit-setup.cfg
```diff
@@ -7,6 +7,7 @@ ignore = E309
 exclude = migrations,snapshots
 max-line-length = 120
 max-complexity = 10
+ignore = E203, W503
```
karisal-anders, line 10: Having a comment explaining why this is done could be clarifying like e.g. in https://github.com/City-of-Helsinki/yjdh/blob/develop/backend/shared/setup.cfg#L8-L11
```python
# E203 and W503 are disabled and W504 enabled for compatibility with Black.
# For details see "Why are Flake8's E203 and W503 violated?" in Black FAQ at
# https://black.readthedocs.io/en/stable/faq.html
```

backend/benefit/applications/tests/conftest.py
```diff
@@ -76,14 +77,21 @@ def talpa_service_with_one_application(talpa_service):
...
     )
 
 
 @pytest.fixture
-def applications_csv_service_with_one_application():
-    return ApplicationsCsvService([DecidedApplicationFactory()])
+def applications_csv_service_with_one_application(applications_csv_service):
+    applications_csv_service.applications[1].delete()
+    applications_csv_service.applications = applications_csv_service.applications[:1]
+    return applications_csv_service
```
karisal-anders, line 94: What is the benefit here from using applications_csv_service as a base?
Would you feel this is better or worse:
```python
@pytest.fixture
def applications_csv_service_with_one_application():
    application1 = DecidedApplicationFactory(application_number=100001)
    return ApplicationsCsvService(Application.objects.filter(pk=application1.pk))

karisal-anders (approved):
Left a few comments which I leave you to consider whether to still do something about them. This is ok to merge as is or with something still done related to the comments.

backend/shared/precommit-setup.cfg
haavikko: good idea, added

backend/benefit/applications/tests/conftest.py
haavikko: hmm my idea was to do the creation of applications at one point... on the other hand, your suggestion makes applications_csv_service_with_one_application more self-contained. Changed

PR #935, opened 2022-03-22
Make browser tests great again ✅ 

nenonja:
## Description :sparkles:
It works! All browser tests are back to [green](https://github.com/City-of-Helsinki/yjdh/actions/runs/2096142717)
Only these platta tests are red but can be ignored for now: https://github.com/City-of-Helsinki/yjdh/runs/5829049685?check_suite_focus=true

It seems that review environment went considerably slower two or three weeks ago which break the test.
This is resolved by set testcafe flag: `retryTestPages` to true and also make the default timeouts to 60 seconds +
Also there is general improvements:
 - logging all browser console messages
 - retry all tests three times if they fail (testcafe's quarantine mode)
 - better request logger (doesnt log everything but only relevant information)


mattikl (approved):
Nice improvements! LGTM

PR #953, opened 2022-03-25
YJDH-504 | KS-Backend: Wrap additional_info endpoint with @transaction.atomic

karisal-anders:
## Description :sparkles:

Wrap additional_info endpoint with @transaction.atomic

## Issues :bug:

YJDH-504


haavikko (approved):
LGTM!

PR #959, opened 2022-03-25
TET Add correct keyword list in youth event filter

pekkvil:

### DIFF
--- a/frontend/tet/shared/src/types/linkedevents.d.ts
+++ b/frontend/tet/shared/src/types/linkedevents.d.ts
@@ -18,8 +18,8 @@ export type LocalizedObject = {
 };
 
 export type IdObject = {
-  id: string;
-  '@id'?: string;
+  id?: string;
+  '@id': string;
   name?: LocalizedObject;
 };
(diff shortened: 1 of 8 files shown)


frontend/tet/shared/src/types/linkedevents.d.ts
```diff
@@ -18,6 +18,7 @@ export type LocalizedObject = {
 };
 
 export type IdObject = {
+  id: string;
```
mattikl, line 21: Would it be simpler to always use `@id` as the option value? Now they both required properties and there are compilation errors due to that.

mattikl (changes requested):

frontend/tet/shared/src/types/linkedevents.d.ts
pekkvil: `id` could be set optional. To my understanding 'id' value from the requested keywords for the list is required (instead of `@id`) since linkedEvents API requires `yso:p4354` format for the event event filtering with keywords

PR #962, opened 2022-03-28
YJDH-499: Show additional info application on handler-ui

nenonja:
## Description :sparkles:
If youth has sent additional info application, let's show the information on the handler-ui along with the original application.
Also section added for vtj-raw data. It shows only fixture data for now.

+ changed shared code confirm dialog to have submitButtonVariant as prop (so we can red alert color as seen below).

## Screenshots :camera_flash:
Screenshot 2022-03-28 at 8 14 52 AM


frontend/kesaseteli/handler/src/components/form/HandlerForm.tsx
```diff
@@ -74,12 +78,44 @@ const HandlerForm: React.FC<Props> = ({ application }) => {
...
+          <pre>
+            {JSON.stringify(
+              {
+                name: 'Anneli kulmala',
+                hetu: '010105-T123',
+                address: '12345 Vantaa',
+              },
+              null,
+              2
+            )}
```
ngocdh236, line 115: Is this data here on purpose..?

nenonja: Yes, there is mock data for the demo purpose until we get the real data out.

ngocdh236 (approved):

PR #963, opened 2022-03-28
Fix audit errors and bump npm package versions 28-03-22

nenonja:


ngocdh236 (approved):
Tested benefit applications and they worked

PR #965, opened 2022-03-28
HL-484: set default sorting of applications in the handler UI

haavikko:
Also:
* move perform_update to BaseApplicationViewSet in order to avoid
  issues with stale submitted_at values
* remove dead code from ApplicationManager

## Description :sparkles:

The default ordering in the handling views:
* In the "received" table, ordering should be by the send time, most recent first
* In the "handling" table, ordering should be by the calculation modification time, most recent first
* In the archive page, ordering should be by handled_at, most recent first.

All these goals are achieved by ordering by first handled_at, then
modified_at.
* in the "received" and "handling" table, no application has handled_at set yet,
so applications will compare as equals
* For received applications, the send time is the same as application modification time

### DIFF
--- a/backend/benefit/applications/api/v1/views.py
+++ b/backend/benefit/applications/api/v1/views.py
@@ -126,6 +126,14 @@ class BaseApplicationViewSet(AuditLoggingModelViewSet):
         "de_minimis_aid_set",
     ]
 
+    def perform_update(self, serializer):
+        super().perform_update(serializer)
+        # In case new AuditLogEntry objects were created during the
+        # processing of the update, then the annotation value for handled_at
+        # in the serializer.instance might have become stale.
+        # Update the object.
+        serializer.instance = self.get_queryset().get(pk=serializer.instance.pk)
+
     @action(methods=["get"], detail=False, url_path="simplified_list")
     def simplified_application_list(self, request):
         """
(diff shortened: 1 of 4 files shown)


backend/benefit/applications/api/v1/views.py
```diff
@@ -313,7 +332,7 @@ def get_queryset(self):
             .prefetch_related(
                 "pay_subsidies", "training_compensations", "calculation__rows"
             )
-        )
+        ).order_by("-handled_at", "-calculation__modified_at")
```
karisal-anders, line 335: The used order_by parameters here and the comment above describing the needed order are out of sync or if they are in sync then please could you explain in more detail how they reflect the same reality. Thank you!

karisal-anders (commented):
Please have a look a the comment I added, otherwise looks ok.

backend/benefit/applications/api/v1/views.py
haavikko: The latest version was not pushed to the server, sorry!

backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -2035,3 +2048,52 @@ def test_application_status_last_changed_at(api_client, handling_application):
...
+                    ApplicationStatus.CANCELLED,
+                ]
+            )
+            .order_by("-created_at")
+            .first()
+        ):
+            handled_at_key = log_entry.created_at.timestamp()
+
+        return (
+            -handled_at_key if handled_at_key else float("-inf"),
```
karisal-anders, line 2087: Why the float("-inf")? I tried this:
```python
>>> a=[(1, 2), (2, -54), (3, 0), (1, 0), (1, 4), (1, -5), (None, 2), (None, -3), (None, 4), (None, 0)]
>>> a
[(1, 2), (2, -54), (3, 0), (1, 0), (1, 4), (1, -5), (None, 2), (None, -3), (None, 4), (None, 0)]
>>> sorted(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

So sorting None and non-None types doesn't work normally. Ok.

karisal-anders (approved):

PR #968, opened 2022-03-29
KS-Backend: Use transactions in test_models.py tests

karisal-anders:
## Description :sparkles:

Enable use of transactions and reset sequences in all tests in
test_models.py. Previously the "with transaction.atomic()" clauses in
these tests did nothing besides having the code wrapped in a with clause
because the related default parameters to pytest.mark.django_db are:
 - transaction=False
   - "With transaction=False (the default when not specified),
     transaction operations are noops during the test."
 - reset_sequences=False
   - "The reset_sequences argument will ask to reset auto increment
     sequence values (e.g. primary keys) before running the test.
     Defaults to False. Must be used together with transaction=True to
     have an effect."

## Issues :bug:

`–`

### DIFF
--- a/backend/kesaseteli/applications/tests/test_models.py
+++ b/backend/kesaseteli/applications/tests/test_models.py
@@ -6,7 +6,7 @@ from applications.models import YouthSummerVoucher
 from common.tests.factories import AcceptedYouthApplicationFactory
 
 
-@pytest.mark.django_db
+@pytest.mark.django_db(transaction=True, reset_sequences=True)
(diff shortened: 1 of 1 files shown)


haavikko (commented):
LGTM! My only question is, does this increase the run time of the tests in any significant way?

karisal-anders:
> LGTM! My only question is, does this increase the run time of the tests in any significant way?

Total running time changed by ~1,8s locally on my machine, so I'd say no.

haavikko (approved):
LGTM!

PR #974, opened 2022-03-30
HL-501: Fix start_date, end_date in new Ahjo "current row" columns

haavikko:
Also fix translation for association_has_business_activities

## Description :sparkles:

In some cases, the start_date and end_date were missing in Ahjo rows

## Testing :alembic:

backend/benefit/applications/services/applications_csv_report.py
backend/benefit/applications/tests/test_applications_csv_report.py

### DIFF
--- a/backend/benefit/applications/services/applications_csv_report.py
+++ b/backend/benefit/applications/services/applications_csv_report.py
@@ -135,7 +135,7 @@ class ApplicationsCsvService(CsvExportBase):
                 str,
             ),
             CsvColumn(
-                "Yhdistys jolla yritystoimintaa?",
+                "Yhdistys jolla taloudellista toimintaa?",
                 "association_has_business_activities",
                 format_bool,
             ),
--- a/backend/benefit/calculator/rules.py
+++ b/backend/benefit/calculator/rules.py
@@ -97,6 +97,8 @@ class HelsinkiBenefitCalculator:
                     training_compensation,
                 )
             )
+        assert ranges[0].start_date == self.calculation.start_date
+        assert ranges[-1].end_date == self.calculation.end_date
         return ranges
 
     def get_amount(self, row_type, default=None):
@@ -172,7 +174,11 @@ class DummyBenefitCalculator(HelsinkiBenefitCalculator):
 
 class ManualOverrideCalculator(HelsinkiBenefitCalculator):
     def create_rows(self):
-        self._create_row(ManualOverrideTotalRow)
+        self._create_row(
+            ManualOverrideTotalRow,
+            start_date=self.calculation.start_date,
+            end_date=self.calculation.end_date,
+        )
 
 
 class SalaryBenefitCalculator2021(HelsinkiBenefitCalculator):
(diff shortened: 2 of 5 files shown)


backend/benefit/applications/services/applications_csv_report.py
```diff
@@ -135,7 +135,7 @@ def CSV_COLUMNS(self):
                 str,
             ),
             CsvColumn(
-                "Yhdistys jolla yritystoimintaa?",
+                "Yhdistys jolla taloudellista toimintaa??",
```
karisal-anders, line 138: Is the double question mark at the end intentional?

backend/benefit/calculator/rules.py
```diff
@@ -282,9 +286,17 @@ def create_rows(self):
                 end_date=date_ranges[-1].end_date,
                 prefix_text="Koko ajalta",
             )
-            self._create_row(SalaryBenefitSumSubTotalsRow)
+            self._create_row(
+                SalaryBenefitSumSubTotalsRow,
+                start_date=date_ranges[0].start_date,
+                end_date=date_ranges[-1].end_date,
```
karisal-anders, line 292: As the date_ranges are created with get_sub_total_ranges they seem to be sorted according to the start_date values in ascending order. That makes date_ranges[0].start_date to be the minimum start_date of the date_ranges. But is date_ranges[-1].end_date always the maximum end_date of the date_ranges? Can the date ranges overlap each other?

backend/benefit/calculator/rules.py
```diff
@@ -282,9 +286,17 @@ def create_rows(self):
...
+            )
         else:
-            self._create_row(SalaryBenefitTotalRow)
+            self._create_row(
+                SalaryBenefitTotalRow,
+                start_date=date_ranges[0].start_date,
+                end_date=date_ranges[-1].end_date,
```
karisal-anders, line 298: Same question as in https://github.com/City-of-Helsinki/yjdh/pull/974/commits/cebb439f8dcf54e040740d53a9f442ef1c78580d#r838207292

karisal-anders (commented):
Left a few comments, on the whole looks quite good to me. Please take a look at the comments and let's sync after that.

backend/benefit/calculator/rules.py
haavikko: Actually the first start_date in the date_ranges must be equal to calculation.start_date and the last end_date must be equal to calculation.end_date, so I was able to simplify this. Added assertions about these invariants to get_sub_total_ranges.

haavikko: start_date=self.calculation.start_date,
                end_date=self.calculation.end_date,

backend/benefit/applications/services/applications_csv_report.py
haavikko: fixed

backend/benefit/calculator/rules.py
haavikko: fixed

karisal-anders (approved):

PR #996, opened 2022-03-30
Increase V8 max heap size in development

mattikl:
## Description :sparkles:

When running TET project locally Docker Compose has been crashing often due to `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`.


nenonja (approved):
browser tests fail but it seems to be unrelated to code changes

PR #1008, opened 2022-04-01
HL-387: send email notification to applicant about new messages

haavikko:
## Description :sparkles:

Email sending is implemented in the same way as in Kesäseteli

Report email sending errors to Sentry so that the errors can be reported further to the handlers. Error while sending email should not block sending messages.

## Testing :alembic:

backend/benefit/messages/tests/test_api.py

### DIFF
--- a/backend/benefit/messages/automatic_messages.py
+++ b/backend/benefit/messages/automatic_messages.py
@@ -1,7 +1,15 @@
+import logging
+from smtplib import SMTPException
+
+from django.conf import settings
+from django.core.mail import send_mail
 from django.utils import translation
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
 from messages.models import Message, MessageType
+from sentry_sdk import capture_message
+
+LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


backend/benefit/messages/automatic_messages.py
```diff
@@ -29,3 +37,58 @@ def send_application_reopened_message(
...
+        try:
+            return send_mail(
+                subject=_message_notification_email_subject(),
+                message=_message_notification_email_body(application),
+                from_email=settings.DEFAULT_FROM_EMAIL,
+                recipient_list=[application.company_contact_person_email],
+                fail_silently=False,
+            )
+        except SMTPException as ex:
+            capture_exception(ex)
```
karisal-anders, line 93: Is it possible that GDPR data is sent to Sentry with this?

karisal-anders: If the email in recipient list can be included in a SMTPException then it is relevant to consider whether application.company_contact_person_email is GDPR data or not?

karisal-anders (approved):

karisal-anders (approved):

PR #1011, opened 2022-04-01
Fix the error message for phone max length validation

ngocdh236:


frontend/benefit/applicant/public/locales/en/common.json
```diff
@@ -428,6 +428,9 @@
         "max": "Maximum value is {{max}}",
         "min": "Value must be at least {{min}}"
       },
+      "phoneNumber": {
+        "max": "Maximum lenght is {{max}}"
```
melniiv, line 432: typo

melniiv (approved):

PR #1024, opened 2022-04-04
Upgrade React from 17 to 18

nenonja:
## Description :sparkles:
Also ignored failing RequestRetry package from audit and upgraded some other packages.


frontend/tet/admin/audit-ci.json
```diff
@@ -0,0 +1,4 @@
+{
```
mattikl, line 1: Why is this under `tet/admin`?

Have we thought about adding `--skip-dev` option to `yarn audit`? That would skip `devDependencies` and thus reduce maintainers work of fixing exposures that are not vulnerabilities in production code.

nenonja: Yes the file should be in the root folder, I'll fix it. I think I was searching to use skip-dev type of option but it didnt exist yet? At least if I try to run now `yarn audit --skip-dev` it still checks the dev deps. That's why I started to use `audit-ci` package instead of built-in audit. It supports that you camn whitelist some packages (for example requestretry below)

mattikl (approved):
Manually tested TET Admin and Youth.

PR #1038, opened 2022-04-07
TETP-56: Add browser tests

mattikl:


frontend/shared/src/__tests__/utils/i18n/create-translations.ts
```diff
@@ -0,0 +1,18 @@
+// cannot import because build doesn't understand '@frontend/shared/src/i18n/i18n'
+// while testcase doesn't understand 'shared/i18n/i18n'
+// import { DEFAULT_LANGUAGE, Language } from '@frontend/shared/src/i18n/i18n';
```
nenonja, line 3: No need to do this, `@frontend/shared/src/` should work for both the build and testcafe run. I think that the problem is that if this file is in the shared project so it doesnt have refenrence to itself with `@frontend/shared`. So just use relative path `import { DEFAULT_LANGUAGE, Language } from '../../../i18n/i18n'`

nenonja, line 3: No need to do this, `@frontend/shared/src/` should work for both the build and testcafe run. I think that the problem is that if this file is in the shared project so it doesnt have refenrence to itself with `@frontend/shared`. So just use relative path `import { DEFAULT_LANGUAGE, Language } from '../../../i18n/i18n'`

frontend/tet/admin/browser-tests/actions/employer-header.actions.ts
```diff
@@ -0,0 +1,18 @@
+import { doLogin, SuomiFiData } from '@frontend/shared/browser-tests/actions/login-action';
+import { getHeaderComponents } from '@frontend/shared/browser-tests/components/header.components';
+import { DEFAULT_LANGUAGE } from '@frontend/shared/src/i18n/i18n';
+import User from '@frontend/shared/src/types/user';
+import TestController from 'testcafe';
+import getEmployerTranslationsApi from '../../src/__tests__/utils/i18n/get-employer-translations-api';
+
+export const doEmployerLogin = async (
```
nenonja, line 8: could be named as `doTetAdminLogin`

frontend/tet/admin/browser-tests/actions/employer-header.actions.ts
```diff
@@ -0,0 +1,18 @@
...
+  t: TestController,
+  lang = DEFAULT_LANGUAGE,
+  cachedUser?: User,
+): Promise<SuomiFiData> => {
+  const { translations } = getEmployerTranslationsApi();
```
nenonja, line 13: could be `getTetAdminTranslationsApi()`

frontend/tet/youth/browser-tests/pages/frontpage.ts
```diff
@@ -0,0 +1,13 @@
+import { t, Selector } from 'testcafe';
+
+class Frontpage {
+  quickSearchInput = Selector('#searchText');
```
nenonja, line 4: using class is nice and perhaps better than my function-like components-apis. However I would prefer to use [testing-library](https://testing-library.com/) -style of selectors instead of css-selectors, something like this could be done: https://dev-tester.com/fix-your-flaky-end-to-end-tests-with-testing-library/
One benefit is that it enforces better accessibiity etc, if you cannot find a good selector, then probably the element is not very accessible either :)
Anyway we can keep these selectors for now but consider to refactor later?

frontend/tet/youth/browser-tests/search.testcafe.ts
```diff
@@ -0,0 +1,61 @@
...
+  .afterEach(async () =>
+    // eslint-disable-next-line no-console
+    console.log(filterLoggedRequests(requestLogger)),
+  );
+
+test('user can search and navigate', async (t) => {
+  const currentYear: number = new Date().getFullYear();
+  const startDate = `31.12.${currentYear}`;
+  const endDate = `31.12.${currentYear + 1}`;
+  const workMethodText = 'varjosta';
```
nenonja, line 28: Itäs a bit hard to understand where these values are coming from by reading just the testing code :)
I know it's hard to write a good test against the linked events though... perhaps ultimately best test would first create the searchable item first (to linked events?) and then search it, perhaps it's something to think about later.

nenonja (changes requested):
Some comments...

All tests does not pass yet:
https://github.com/City-of-Helsinki/yjdh/runs/6091440160?check_suite_focus=true

Form summary page you can download videos, it's easier to debug what went wrong by wathcing the recorded videos. Also browsing logged results carefully might reveal the issue.

.github/workflows/te-review.yml
```diff
@@ -234,79 +234,88 @@ jobs:
...
+        id: testcafe
+        run: yarn --cwd ${{matrix.dir}} browser-test:ci -q attemptLimit=3,successThreshold=1
+        env:
+          GITHUB_WORKFLOW_NAME: ${{ github.workflow }}
+          GITHUB_WORKFLOW_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          FRONTEND_URL: ${{ env.ENVIRONMENT_URL }}
+          EMPLOYER_URL: ${{ env.EMPLOYER_URL }}
+          YOUTH_URL: ${{ env.YOUTH_URL }}
+          ADMIN_URL: ${{ env.ADMIN_URL }}
+          NEXT_PUBLIC_BACKEND_URL: ${{ env.NEXT_PUBLIC_BACKEND_URL }}
```
nenonja, line 289: NEXT_PUBLIC_BACKEND_URL is not set and thus the backend responds Internal server error (500), because testcafe uses NEXT_PUBLIC_BACKEND_URL as referrer value for the http requests: https://stackoverflow.com/questions/61116376/set-referrer-for-firefox-on-test-cafe

nenonja, line 289: NEXT_PUBLIC_BACKEND_URL is not set and thus the backend responds Internal server error (500), because testcafe uses NEXT_PUBLIC_BACKEND_URL as referrer value for the http requests: https://stackoverflow.com/questions/61116376/set-referrer-for-firefox-on-test-cafe

frontend/tet/youth/browser-tests/pages/frontpage.ts
nenonja: Actually no need to do anything for this pr. I'm working on Helsinki lisä browser test and started to create an abstract page component class for these page model classes which sets translations and selectors.
- Here is the [root class](https://github.com/City-of-Helsinki/yjdh/blob/992e7682d0195bf931283ee63fe3c499af8f44e2/frontend/shared/browser-tests/page-models/PageComponent.ts) 
- And here is the [base class](https://github.com/City-of-Helsinki/yjdh/blob/992e7682d0195bf931283ee63fe3c499af8f44e2/frontend/benefit/applicant/browser-tests/page-model/ApplicantPageComponent.ts) per project that set translations.
- And finally a [sub class](https://github.com/City-of-Helsinki/yjdh/blob/992e7682d0195bf931283ee63fe3c499af8f44e2/frontend/benefit/applicant/browser-tests/page-model/TermsOfService.ts) for page component which has concrete selectors and actions.

Perhaps if this is a good idea, we can adapt it gradually with other projects?

nenonja: Now the [page component](https://github.com/City-of-Helsinki/yjdh/blob/develop/frontend/shared/browser-tests/page-models/PageComponent.ts) root class and others are merged to develop and can be used also in tet-browser test, if needed. But perhaps it's better to start use it later and try to get this pr in first :)

frontend/tet/youth/browser-tests/search.testcafe.ts
mattikl: It's a good question what the testing strategy should be like. Currently we aim to to test the details in unit and component tests, and the browser test is more of a proof of concept that all of the services work together. I'll add a comment clarifying this.

There is no direct need to test Linked Events here. However, there are no strong guarantees that Linked Events won't introduce changes in the future that are backwards incompatible from TET's point of view. Having an end-to-end test that highlights the corner cases how TET is using LE might be the best way to detect if a change like this has happened.

mattikl:
Created issue https://helsinkisolutionoffice.atlassian.net/browse/TETP-213 about refactorings discussed here.

nenonja (approved):
tests are passing!

PR #1039, opened 2022-04-08
YJDH-526: Show "send it anyway" link to allow user to submit the application without vtj-validation

nenonja:
## Description :sparkles:

When vtj-validation fails and user's application does not match to it, an error summary is displayed where user is asked to check the data. The form also shows a link button which allows user to send the application anyway and bypass vtj validation if (s)he think thats application data is correct.

Also done:
- refactored youth form to smaller react components
- component tests
- improve usage of translations in jest/browser tests 

## Screenshots :camera_flash:

Screenshot 2022-04-09 at 9 51 55 AM


mattikl (approved):
The component tests seem to cover the new feature well. The improvements to browser tests look good but unrelated to the new feature.

PR #1042, opened 2022-04-12
TETP-199 show single location on map

pekkvil:


mattikl (approved):
All the errors seem to be resolved. All the changes inside TET project except adding devDependencies to frontend.

PR #1055, opened 2022-04-20
YJDH-527: Show applicant's vtj-data on handler UI

nenonja:
## Description :sparkles:
Shows Vtj (Väestötietojärjestelmä) data for the application. There is also error indicators when there is some notable exceptions in the data:
Screenshot 2022-04-21 at 4 19 19 PM

Raw data is also available under accordion:
Screenshot 2022-04-21 at 4 19 06 PM


pekkvil (approved):
I didn't really find anything for change recommendations or even typos. Good to have date.utils also tested :). Using translationApi for the queries in tests seems also a good idea that should be added to tet-project's tests. Prevents writing tests again in a case of translation changes.

nenonja:
> I didn't really find anything for change recommendations or even typos. Good to have date.utils also tested :). Using translationApi for the queries in tests seems also a good idea that should be added to tet-project's tests. Prevents writing tests again in a case of translation changes.

🙌

PR #1057, opened 2022-04-22
TETP-146 youth can filter with attributes

pekkvil:


nenonja (commented):
Looks like I did review yesterday but forgot to submit it :)
Anyway some minor comments about the code only, I didnt think about the business case.

mattikl (approved):
Tested in review.

PR #1088, opened 2022-04-28
YJDH-539: Update hl consent pdfs with new links

nenonja:
## Description :sparkles:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-539


melniiv (approved):
Correct

PR #1089, opened 2022-04-28
Update npm packages 28.4.2022

nenonja:


nenonja:
employer browser test seems to be unstable, but going to be fixed in other pr https://github.com/City-of-Helsinki/yjdh/pull/1093

mattikl (approved):
The updates are minor and patch in dependencies, and all the major version updates are in devDependencies.

PR #1094, opened 2022-05-02
YJDH-532 | KS-Backend: Disregard rejected youth applications

karisal-anders:
## Description :sparkles:

Disregard rejected youth applications when checking whether a social
security number or email is used already. Update related test.

## Issues :bug:

YJDH-532


karisal-anders:
@mattikl KS browser tests won't pass, they are known to be broken at the moment (Talked with @nenonja about their state just now).

mattikl (approved):
Looks clear!

PR #1105, opened 2022-05-03
YJDH-543 | KS-Backend: Hide VTJ data from youth / applicant

karisal-anders:
## Description :sparkles:

The handler is allowed to see the VTJ data retrieved from Finnish
Population Information System i.e. VTJ i.e. Väestötietojärjestelmä
but the youth / applicant is not. Previously both could see the VTJ
data, now it is not shown to the youth / applicant.

## Issues :bug:

YJDH-543


mattikl (approved):
Looks good to me!

PR #1113, opened 2022-05-04
TETP-208: Add support for mocked suomi.fi authentication

mattikl:
## Description :sparkles:

The purpose of this PR is that login from path `/oidc/authenticate` logs in a "company user" (suomi.fi puolesta asiointi) instead of city of Helsinki user (login via /oauth2/authenticate). We can currently test when the logins are mocked (localhost or review environment). Setup for testing suomi.fi login is still underway by Helsinki profiili.

### DIFF
--- a/backend/tet/events/api/v1/permissions.py
+++ b/backend/tet/events/api/v1/permissions.py
@@ -18,7 +18,7 @@ class TetAPIPermission(BasePermission):
          the postings of their company.
       3. Users logged in via AD NOT belonging to a group listed in `settings.ADFS_CONTROLLER_GROUP_UUIDS`.
          This could be anyone with a Microsoft account. The ADFS setup might prevent these logins,
-         but we cannot be sure about this, so we need to check that not anyone is allowed to creating
+         but we cannot be sure about this, so we need to check that they aren't allowed to create
          a TET posting as an employee of the city of Helsinki.
 
          This permission prevents case 3 from accessing the service.
--- a/backend/tet/events/api/v1/permissions.py
+++ b/backend/tet/events/api/v1/permissions.py
@@ -1,12 +1,41 @@
+from django.conf import settings
+from django.http import HttpRequest
 from rest_framework.permissions import BasePermission
 
 
-class TetAdminPermission(BasePermission):
+class TetAPIPermission(BasePermission):
     """
-    Permission check for managing city TET postings.
+    Permission to access the Linked Events API as authorized.
+
+    There are three kinds of possible logins:
+
+      1. Users logged in via AD belonging to a group listed in `settings.ADFS_CONTROLLER_GROUP_UUIDS`.
+         This group should contain all City of Helsinki users needing to use this servicing.
+         These users are logged in as staff users and should be allowed access to their own
+         TET postings.
+      2. Users logged in via suomi.fi (puolesta asiointi)
+         The service works if we can find a business_id for the user and the user is allowed access to
+         the postings of their company.
+      3. Users logged in via AD NOT belonging to a group listed in `settings.ADFS_CONTROLLER_GROUP_UUIDS`.
+         This could be anyone with a Microsoft account. The ADFS setup might prevent these logins,
+         but we cannot be sure about this, so we need to check that not anyone is allowed to creating
+         a TET posting as an employee of the city of Helsinki.
+
+         This permission prevents case 3 from accessing the service.
+
+         For cases 1 and 2, `ServiceClient` is responsible to further check the authorization
+         based on user actions.
     """
 
-    def has_permission(self, request, view):
+    def has_permission(self, request: HttpRequest, view):
+        # Allow all authenticated access when logins are mocked
+        if settings.NEXT_PUBLIC_MOCK_FLAG:
+            return bool(request.user)
+
         return bool(
-            request.user and (request.user.is_staff or request.user.is_superuser)
+            request.user
+            and (
+                (request.user.is_staff or request.user.is_superuser)
+                or (request.session.get("eauth_id_token") is not None)
+            )
         )
--- a/backend/tet/events/services.py
+++ b/backend/tet/events/services.py
@@ -2,6 +2,7 @@ import logging
 from datetime import date
 
 from django.conf import settings
+from django.http import HttpRequest
 from rest_framework.exceptions import PermissionDenied
 
 from events.linkedevents import LinkedEventsClient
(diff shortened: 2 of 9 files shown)


backend/tet/events/api/v1/permissions.py
```diff
@@ -1,12 +1,41 @@
...
+         This group should contain all City of Helsinki users needing to use this servicing.
+         These users are logged in as staff users and should be allowed access to their own
+         TET postings.
+      2. Users logged in via suomi.fi (puolesta asiointi)
+         The service works if we can find a business_id for the user and the user is allowed access to
+         the postings of their company.
+      3. Users logged in via AD NOT belonging to a group listed in `settings.ADFS_CONTROLLER_GROUP_UUIDS`.
+         This could be anyone with a Microsoft account. The ADFS setup might prevent these logins,
+         but we cannot be sure about this, so we need to check that not anyone is allowed to creating
+         a TET posting as an employee of the city of Helsinki.
```
karisal-anders, line 22: "so we need to check that not anyone is allowed to creating"

Do you mean this?
"so we need to check that they are not allowed to create"

backend/tet/events/services.py
```diff
@@ -35,6 +37,20 @@ def _user_matches(event, user):
...
+    return event["provider"][PROVIDER_BUSINESS_ID_FIELD] == business_id
+
+
+def _business_id_is_set(event):
+    return (
+        event["provider"] is not None
+        and PROVIDER_BUSINESS_ID_FIELD in event["provider"]
+    )
+
+
```
karisal-anders, line 53: Possible alternative by using `_business_id_is_set`:
```python
def _business_id_is_set(event):
    return (
        event["provider"] is not None
        and PROVIDER_BUSINESS_ID_FIELD in event["provider"]
    )

def _business_id_matches(event, business_id):
    return (
        _business_id_is_set(event)
        and event["provider"][PROVIDER_BUSINESS_ID_FIELD] == business_id
    )
```

backend/tet/events/services.py
```diff
@@ -43,11 +59,7 @@ class ServiceClient:
...
         self.client = LinkedEventsClient()
 
-    # All users are city employees (AD login) in MVP phase
-    def _is_city_employee(self, user):
-        return True
-
-    def _get_event_and_raise_for_unauthorized(self, user, event_id):
+    def _get_event_and_raise_for_unauthorized(self, request: HttpRequest, event_id):
         """Raises PermissionDenied unless `user` is authorized to access event `event_id`
         and also returns the event.
```
karisal-anders, line 64: user → request.user?

backend/tet/events/services.py
```diff
@@ -56,51 +68,60 @@ def _get_event_and_raise_for_unauthorized(self, user, event_id):
...
     def _get_publisher(self, user):
         # In MVP phase just returning ahjo:00001 (Helsingin kaupunki) for all
         # We can also get industry (toimiala) from ADFS Graph API (needs translation to ahjo scheme)
         # For company users it is still undecided how we store the company business id (Y-tunnus) in an event
         return "ahjo:00001"
 
-    def list_job_postings_for_user(self, user):
+    def list_job_postings_for_user(self, request: HttpRequest):
         # Currently this fetches all events under the TET data source, but it's more efficient if we can
         # filter by industry (toimiala) or company business id (Y-tunnus)
```
karisal-anders, line 110: Is this comment still up to date with the code changes in this function?

backend/tet/events/tests/test_services.py
```diff
@@ -20,6 +20,25 @@
...
+        user.username = username
+
+    factory = RequestFactory()
+    request = factory.get("/")
+    request.user = user
+
+    return request
+
+
 # We expect to have a user that is properly authenticated and is authorized to access Linked Events services.
```
karisal-anders, line 52: Are these comments still up to date with the code changes made to the tests?

backend/tet/events/utils.py
```diff
@@ -0,0 +1,48 @@
...
+def get_business_id(request: HttpRequest):
+    if settings.NEXT_PUBLIC_MOCK_FLAG:
+        return "123456-7"
+
+    try:
+        organization_roles = get_organization_roles(request)
+    except RequestException:
+        raise PermissionDenied(
+            detail="Unable to fetch organization roles from eauthorizations API"
+        )
```
karisal-anders, line 29: This handles all RequestException exceptions as PermissionDenied here.
More finegrained error handling could be beneficial but I leave it up to you ponder and decide.

backend/tet/events/tests/data/linked_events_responses.py
```diff
@@ -551,7 +551,8 @@
             "search_vector_sv": "",
             "replaced_by": null,
             "provider": {
-                "fi": "Päiväkoti"
+                "fi": "Test company",
+                "sv": "654321-5"
```
karisal-anders, line 555: "fi" contains company name and "sv" contains business ID, is this correct?

backend/tet/events/tests/test_services.py
```diff
@@ -11,40 +11,51 @@
...
     EVENT_RESPONSE_OTHERUSER,
+    EVENT_RESPONSE_TEST_COMPANY,
     EVENT_RESPONSE_TESTUSER_EMAIL,
     EVENT_RESPONSE_TESTUSER_OID,
     SAMPLE_EVENTS,
 )
 
 LOGGER = logging.getLogger(__name__)
+TEST_COMPANY_BUSINESS_ID = (
+    "654321-5"  # Needs to match linked_events_responses.EVENT_RESPONSE_TEST_COMPANY
```
karisal-anders, line 23: Maybe use this value in EVENT_RESPONSE_TEST_COMPANY or get this value from EVENT_RESPONSE_TEST_COMPANY so that it doesn't have to be duplicated?

karisal-anders (commented):
Left some comments, please have a look at them.

backend/tet/events/tests/data/linked_events_responses.py
mattikl: This is indeed a bit misleading. Normally we use Linked Event's `custom_data` field for attributes that the Linked Events data model doesn't support. But the problem is that `custom_data` is not searchable, and we need to have these searchable. This is why we use the `provider` object to store this data.

The code uses constants `PROVIDER_NAME_FIELD` and `PROVIDER_BUSINESS_ID_FIELD` to refer to these field. I'll update the test code to also them.

backend/tet/events/utils.py
mattikl: Good question how the error should be handled. User needs to be linked to a company and this error means we cannot find that company, which makes the TET service unusable. So maybe a 5xx error would be more suitable?

karisal-anders (approved):

PR #1114, opened 2022-05-04
TETP Fix error toast before submit

pekkvil:


mattikl (approved):
This caused `<EditorErrorNotification>` flashing after submitting a valid form, because `isValid` was still false. We tested that using `noTetErrors` is sufficient to find the errors.

You could still delete unused variables `isValid` and `submitCount`.

PR #1117, opened 2022-05-05
TETP-98 show confirmation when leaving form

pekkvil:


nenonja (approved):
Looks good! Only thing is that you could move the hook to the shared code as this is a feature that has been asked for other projects too but not yet been prioritized :)

pekkvil:
Moved hook to shared and added option for the own translated message

PR #1118, opened 2022-05-05
YJDH-538 | BF-Backend: Export CSVs using StreamingHttpResponse

karisal-anders:
### BF-Backend: Export CSVs using StreamingHttpResponse

Rationale:
 - Rationale for exporting CSVs using StreamingHttpResponse is to try to
   make CSV exports not return Gateway Timeout i.e. HTTP status code 504
   so easily. The slow CSV export in question is
   HandlerApplicationViewSet.export_csv which seems to be making a lot
   of queries to the database (I tried one export_csv call in a test and
   it made over 200 queries to the database). Because optimizing the
   database query wasn't straightforward other ways to make the CSV
   export pragmatically usable albeit slow were undertaken, and voilà:
   this commit.

Make all CSV exports use StreamingHttpResponse:
 - ApplicationBatchViewSet.talpa_export_batch
 - HandlerApplicationViewSet.export_csv
 - HandlerApplicationViewSet.export_new_accepted_applications
 - HandlerApplicationViewSet.export_new_rejected_applications

To accomplish streaming of CSV data change CSV data generation to use
Python generators by using the yield keyword.

Tests:
 - Fix related tests
 - Add new tests:
   - test_talpa_csv_cell_list_lines_generator
   - test_talpa_csv_string_lines_generator
   - test_applications_csv_cell_list_lines_generator
   - test_applications_csv_string_lines_generator

Add type hints.

Refs YJDH-538

### BF-Backend: Fix bf-py-coding-style.yml & run isort

The GitHub workflow bf-py-coding-style.yml was problematic before
because the isort check for some unknown reason worked differently with
the precommit-setup.cfg than with the benefit backend's own setup.cfg.

Fix the GitHub workflow bf-py-coding-style.yml's "Formatting", "Linting"
and "Import sorting" by separating them into (BF) and (Shared) portions
like in other backends' *-py-coding-style.yml files.

Run "isort ." to fix import sortings.

Refs YJDH-538 (Noticed import sortings were incorrect)

## Issues :bug:

YJDH-538

### DIFF
--- a/.github/workflows/bf-py-coding-style.yml
+++ b/.github/workflows/bf-py-coding-style.yml
@@ -40,11 +40,20 @@ jobs:
       - name: Install dependencies
         run: cd backend/benefit && pip install -r requirements.txt -r requirements-dev.txt
 
-      - name: Formatting
+      - name: (BF) Formatting
         run: cd backend/benefit && black --check .
 
-      - name: Linting
-        run: cd backend/benefit && flake8 --config ../shared/precommit-setup.cfg
+      - name: (BF) Linting
+        run: cd backend/benefit && flake8
 
-      - name: Import sorting
-        run: cd backend/benefit && isort -c --settings-path ../shared/precommit-setup.cfg .
+      - name: (BF) Import sorting
+        run: cd backend/benefit && isort -c .
+
+      - name: (Shared) Formatting
+        run: cd backend/shared && black --check .
+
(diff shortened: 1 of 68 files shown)


mattikl (approved):
This change seems to keep the logic intact while introducing the change that CSVs are streamed row by row via HTTP instead of first loading the rows and then sending the chunk.

PR #1120, opened 2022-05-06
YJDH-532 | KS-Backend: Ignore rejected youth applications in activation

karisal-anders:
## Description :sparkles:

PR #1094 was incomplete, it only disregarded rejected youth applications
in youth application creation. This commit adds disregarding of
rejected youth applications in activation. This should now allow a
person to submit a new youth application and have it accepted even if
they have a previously rejected youth application.

Add test cases for activation where a rejected youth application for the
same person using the same email already exists. All rejected youth
applications should be ignored in activation.

Refs YJDH-532 (Disregarding rejected youth applications)
Refs PR #1094 (Disregarding rejected youth applications on creation)

## Issues :bug:

YJDH-532
PR #1094


mattikl (approved):
Reviewed together on a Slack huddle.

PR #1131, opened 2022-05-06
TETP-207 Fix required validation for the location field

pekkvil:
…d location

### DIFF
--- a/frontend/tet/admin/src/components/editor/companyInfo/CompanyInfo.tsx
+++ b/frontend/tet/admin/src/components/editor/companyInfo/CompanyInfo.tsx
@@ -12,6 +12,7 @@ import { LocationType, OptionType } from 'tet-shared/types/classification';
 import ComboboxSingleSelect from 'tet/admin/components/editor/ComboboxSingleSelect';
 import TetPosting from 'tet-shared/types/tetposting';
 import useEventPostingTransformation from 'tet-shared/hooks/backend/useEventPostingTransformation';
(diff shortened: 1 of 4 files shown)


frontend/tet/admin/src/components/editor/companyInfo/CompanyInfo.tsx
```diff
@@ -78,7 +84,7 @@ const CompanyInfo: React.FC = () => {
               options={keywords}
               optionLabelField={'label'}
               filter={addressFilterHandler}
-              validation={{ required: { value: true, message: t('common:editor.posting.validation.required') } }}
```
nenonja, line 81: Hmm, I think that the react-hook-form basic validation should work? Maybe the object was in wrong format?
I have had it like this for example: ` validation={{ required: true, maxLength: 64 }}`

nenonja (commented):
One comment

frontend/tet/admin/src/components/editor/companyInfo/CompanyInfo.tsx
pekkvil: Problem was that posting's location key's value is set with initial posting data from PreviewContext.tsx. Since location key is not optional, whole Location type object is set (with empty strings and arrays). This seen as a value in validation, since it is not just 'null' and error for the field is not shown, if form is saved without changes on this field.

mattikl (approved):
Tested locally. The address field now shows error messages if the value is not an address.

PR #1145, opened 2022-05-13
TETP-182: Show correct organization

mattikl:
## Description :sparkles:

Sets the organization for events when adding or editing. Displays organization in both UIs.


mattikl:
Tested extensively on localhost with `NEXT_PUBLIC_MOCK_FLAG` both on and off (I have no test account to test the ADFS flow with the mock flag off, so this needs to be tested on PLATTA TEST).

PR #1093, opened 2022-04-29
YJDH-545: KS: enable Vtj mock integration + new browser tests

nenonja:
## Description :sparkles:
- Enable vtj mock integration
- bunch of new browser tests testing vtj mock integration
- refactoring browser tests components to page model classes

PR #1146, opened 2022-05-13
YJDH-535: Show backend validation error

nenonja:
## Description :sparkles:
In production sometime we got 400 error when user sends the application. Because of gdpr we cannot see why is it for but it's probably unknown validation error. Now user is asked to fix the form fields that are invalid in the backend.


nenonja:
Failing browser tests will be fixed in https://github.com/City-of-Helsinki/yjdh/pull/1093

pekkvil (approved):

PR #1152, opened 2022-05-23
(wip): Yjdh-553 bank account input for employer

nenonja:
## Description :sparkles:
Add bank account input field for kesäseteli employer. it supports all iban values and implemented with react-input-mask (same package as helsinkilisä uses).
One issue with the `react-input-mask` is that it's not seem to be possible to write react-testing-library test for it: https://github.com/sanniassin/react-input-mask/issues/174
So component tests are missing. Browser test will still cover the happy case and fills the input with random iban value.


mattikl (approved):
Looks clear & clean and improvements to local dev setup look good too.

PR #1153, opened 2022-05-23
YJDH-555: Remove invoicer data from employer form

nenonja:
## Description :sparkles:
these invoicer fields are removed:
Screenshot 2022-05-23 at 3 15 40 PM

+ bonus: also fixed occasionally failing gdpr-test


nenonja:
Browser test fails because it already expects bank account number which is implemented in other pr: https://github.com/City-of-Helsinki/yjdh/pull/1152
Let's wait to get it into develop first and the. rebase + test before removing wip status.

pekkvil (approved):
Class model for the fakeObjects seems like a good way to when number of fake datas starts rising. I was wondering if there should be a factory of it's own for the more common shared objects (like user) since employments and attachment are for the kesäseteli only? No need for now (since it seems like user object is the only shared one), but just wondering :)

nenonja:
> Class model for the fakeObjects seems like a good way to when number of fake datas starts rising. I was wondering if there should be a factory of it's own for the more common shared objects (like user) since employments and attachment are for the kesäseteli only? No need for now (since it seems like user object is the only shared one), but just wondering :)

Thanks! There is actually multiple fake-objects.ts files per project but now I only converted the shared one to the class model. The main reason was the gdpr-test that started to fail occasionally. Apparently it masks ssns and sometimes generated uuids included ssn. The only way to fix the test failing randomly was to make the fake-objects to have setting to generate either uuids or counter ids. And that type of setting is easier to implement in class model. Perhaps later the other fake-object files could be refactored to be same type of class model.

PR #1160, opened 2022-05-27
DRAFT: YJDH-574 | Add Suomi.fi integration

karisal-anders:
## Description :sparkles:

Add Suomi.fi integration and make it configurable whether to use Helsinki profile (Helsinki-profiili in Finnish) or Suomi.fi to login.

## Issues :bug:

YJDH-574 (Taking djangosaml2 into use)
YJDH-578 (Making Suomi.fi / Helsinki profile login configurable)


mattikl (approved):
Reviewed quickly due to need to deploy fast. Builds in review environment are failing, but this looks like an issue with the review environment itself (we can open a new PR after this is merged to see if the issues persist).

Focused that this change doesn't break other apps and that no secrets are listed in code. Both seem fine and we can fix possible other issues in the coming week.

PR #1173, opened 2022-05-31
TETP-209 return to list view with same params

pekkvil:


mattikl (approved):
Fixes the return button that was broken when navigating to the page from an external link. Tested locally.

PR #1183, opened 2022-06-01
Update npm deps 06-01-22 + refactor helsinki-lisä types

nenonja:
## Description :sparkles:
One thing leads to another. This was supposed to be only dependency update pr but helsinki-lisä typescript start nagging after updates. I was not able to resolve it instantly so I ended up refactoring the types more clear, there was a lot of copy-paste types in benefit-applicant and benefit-handler projects, so those types are now unified and moved to benefit-shared code.


frontend/tet/admin/package.json
```diff
@@ -32,7 +32,7 @@
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
     "react-hook-form": "^7.30.0",
-    "react-leaflet": "^3.2.5",
+    "react-leaflet": "^4.0.0",
```
pekkvil, line 35: Was there a dependency issue that required 4.0.0? I'm a bit worried about react-leaflet-marketcluster, since it hasn't really been supported lately. There wasn't really any good alternatives that would do the same (good clustering with markers) so had go with that. But luckily i has worked well lately. I tested with react-leaflet 4.0 and it seems to still work :)

nenonja: I can revert it, I already did from another package but didnt see the other one: https://github.com/City-of-Helsinki/yjdh/pull/1183/commits/2f014d302d1ce8866fe89ff93b435328fd1e2876
in tet/youth it caused a test to fail.

pekkvil: Ok. Yeah, it might be safe to just keep react-leaflet on 3.x-version. Hope it doesn't clash with other updates in the future

PR #1184, opened 2022-06-01
YJDH-592 | Dev env improvements

charn:
This PR moves the logic in the overridden docker entrypoint in docker-compose configs to `docker-entrypoint.sh` scripts where it can be controlled using a environment variable. Also debugging output is increased by printing deprecation warnings when developing (i.e. when running Django's runservers).

A new `COMPILE_TRANSLATIONS` environment variable is added, which can be used to enable compiling translations on backend container startup (benefit, kesaseteli projects)


mattikl (approved):
Looks good.

PR #1186, opened 2022-06-02
YJDH-581 YJDH-582 | Send copy of the youth summer voucher to handlers

charn:
Send a hidden copy (bcc) of the youth summer voucher to the handler email. This will enable the handlers to deal with possible problems the users are having e.g. handlers are able to re-send a lost voucher to a user.

### DIFF
--- a/backend/kesaseteli/applications/models.py
+++ b/backend/kesaseteli/applications/models.py
@@ -874,13 +874,23 @@ class YouthSummerVoucher(HistoricalModel, TimeStampedModel, UUIDModel):
             content_id="helsinki_logo",
         )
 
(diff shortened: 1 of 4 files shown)


mattikl (approved):
Looks clear and the added functionality is tested. As a side note, the code is calling `send_youth_summer_voucher_email(language=youth_application.language)` and this call is not covered by the test. It's pretty clear by reading the code that this call still works, just a note that there's no regression test for that particular call.

charn:
I added an additional test for the call with no optional arguments i.e. for the call `send_youth_summer_voucher_email(language=youth_application.language)`

PR #1192, opened 2022-06-06
YJDH-593 | Don't check VTJ on additional info check if DISABLE_VTJ set

charn:
Reactivating an application redirected to additional info page instead of "already activated" page.


mattikl (approved):
Looks clear!

nenonja (commented):
Quick work! 👍

PR #1182, opened 2022-06-01
YJDH-585: ks-youth translations for additional information application

nenonja:

PR #1208, opened 2022-06-08
YJDH-595 | Additional DISABLE_VTJ fixes

charn:
When `DISABLE_VTJ` is set the logic now also checks if additional information is required for the application. `YouthApplication.need_additional_info` will require additional info when an unlisted school is selected if VTJ is disabled.


nenonja (approved):
Looks good! Only thing that I just realized is that the additional info page is actually missing in production. But I think that we only need this pr in https://github.com/City-of-Helsinki/yjdh/pull/1182 and can deploy youth-ui to production as well.
We can put this in

PR #1215, opened 2022-06-14
YJDH-596: buttons to generate pdf reports for ahjo

nenonja:
## Description :sparkles:
Buttons to download pdf reports zipped + rest backend implementation
More info (in finnish): https://helsinkisolutionoffice.atlassian.net/browse/YJDH-596

Screenshot 2022-06-23 at 3 32 01 PM

## Issues :bug:

YJDH-596

### DIFF
--- a/backend/benefit/applications/services/ahjo_integration.py
+++ b/backend/benefit/applications/services/ahjo_integration.py
@@ -21,18 +21,15 @@ TEMPLATE_ID_COMPOSED_DECLINED_PRIVATE = "composed_declined_private"
 JINJA_TEMPLATES_COMPOSED = {
     TEMPLATE_ID_COMPOSED_ACCEPTED_PUBLIC: {
         "path": "composed_accepted_public.html",
-        "file_name": "Liite 1 Helsinki-lisä koontiliite "
-        "julkinen.pdf",
+        "file_name": "Liite 1 Helsinki-lisä koontiliite julkinen.pdf",
     },
     TEMPLATE_ID_COMPOSED_DECLINED_PUBLIC: {
         "path": "composed_rejected_public.html",
-        "file_name": "Liite 1 Helsinki-lisä koontiliite "
-                     "julkinen.pdf",
+        "file_name": "Liite 1 Helsinki-lisä koontiliite julkinen.pdf",
     },
     TEMPLATE_ID_COMPOSED_ACCEPTED_PRIVATE: {
         "path": "composed_accepted_private.html",
-        "file_name": "Helsinki-lisä koontiliite "
-        "salassa pidettävä.pdf",
+        "file_name": "Helsinki-lisä koontiliite salassa pidettävä.pdf",
     },
     TEMPLATE_ID_COMPOSED_DECLINED_PRIVATE: {
         "path": "benefit_declined.html",  # Composed decline template reuse single company decline template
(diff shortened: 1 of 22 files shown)


backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -13,23 +14,26 @@
...
-        "salassa pidettävä.pdf",
+        "file_name": "Helsinki-lisä koontiliite salassa pidettävä.pdf",
     },
     TEMPLATE_ID_COMPOSED_DECLINED_PRIVATE: {
         "path": "benefit_declined.html",  # Composed decline template reuse single company decline template
-        "file_name": "Helsinki-lisä 2021 kielteiset päätökset koontiliite "
+        "file_name": "Helsinki-lisä kielteiset päätökset koontiliite "
         "yrityksille salassa pidettävä.pdf",
     },
 }
```
karisal-anders, line 39: The file_name values are inconsistent and they're not unique because the accepted public filename is the same as the declined public filename.  Also the decline private has "yrityksille" part in it but none of the others don't although they previously had "yhteisöille" in them. Please could you check these filenames out and update them, thank you!

backend/benefit/applications/services/pdf_templates/composed_accepted_private.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 
 {% block content %}
-<p>Työllisyydenhoidon Helsinki-lisän myöntäminen työnantajille</p>
+<p>Työllisyydenhoidon Helsinki-lisä myöntäminen työnantajille</p>
```
karisal-anders, line 4: Why was this changed? I think the former version is grammatically correct and the new is incorrect.

backend/benefit/applications/services/pdf_templates/composed_accepted_public.html
```diff
@@ -1,10 +1,10 @@
 {% extends "base.html" %}
 
 {% block content %}
-<p>Työllisyydenhoidon Helsinki-lisän myöntäminen työnantajille</p>
+<p>Työllisyydenhoidon Helsinki-lisä myöntäminen työnantajille</p>
```
karisal-anders, line 4: Same as https://github.com/City-of-Helsinki/yjdh/pull/1215/files#r943182821 i.e. former version is grammatically correct and the new one is not.

karisal-anders (commented):
Left a few comments, please take a look at them.

backend/benefit/applications/services/ahjo_integration.py
nenonja: fixed

backend/benefit/applications/services/pdf_templates/composed_accepted_private.html
nenonja: no idea why, fixed

backend/benefit/applications/services/pdf_templates/composed_accepted_public.html
nenonja: no idea why, now fixed

karisal-anders (approved):
Squashed the last two commits "review fixes" and "review fixes again" into one commit "review fixes" to reduce the amount of commits. Force pushed the changes.

Because of the current situation, time limits etc I approve this pull request as it is now even though some things could be improved still. Let's just merge this if the tests pass and things can be polished later if there is a real need for the polishing.

PR #1220, opened 2022-06-16
TETP-253 Implement suomi.fi authentication via Helsinki Profile

mattikl:
## Description :sparkles:

- modify TET settings to use Tunnistamo urls instead of those by Tunnistus service
- add `HelsinkiProfileClient` to read `user_ssn` from Helsinki Profile GraphQL API
- modify local development nginx settings  to get `https://` scheme for `callback_uri`
- improve documentation


backend/shared/shared/oidc/views/eauth_views.py
```diff
@@ -56,13 +61,37 @@ def register_user(self, person_id):
...
         user_info = get_userinfo(request)
 
+        # When authenticating via Tunnistus service, we can read `user_ssn` from `user_info`,
+        # but for authentication via Tunnistamo, we need to call Helsinki Profile GraphQL API.
         user_ssn = user_info.get("national_id_num")
+        if user_ssn is None:
+            try:
+                profile = HelsinkiProfileClient().get_profile(oidc_access_token)
+            except HelsinkiProfileException as e:
+                logger.warning(
```
charn, line 83: One option here would be to use `logger.exception(...)` [which should automatically](https://docs.python.org/3/library/logging.html#logging.Logger.exception) put the exception info to the logging message. It will be an logged at ERROR level, which is maybe something that we don't want to do here. Not saying that this needs to be changed. Just pointing out that there's this special logger which can be used inside an exception handler.

backend/shared/shared/helsinki_profile/hp_client.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        return {"user_ssn": national_identification_number}
+
+    def get_api_access_token(self, oidc_access_token):
+        """
+        Exchanges OIDC access token for API access token
+        """
+        try:
+            response = requests.get(
```
charn, line 88: I personally always like to add an explicot timeout to calls made by requests, since [by default they will not time out](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts).

charn (approved):
Looks good to me. :+1: :shipit:  Added a few minor comments.

backend/shared/shared/helsinki_profile/hp_client.py
mattikl: Thanks for pointing this out. Added the timeout, even though I think a better approach would be to define it at a higher level for all requests, but now just keeping things simple.

PR #1227, opened 2022-06-23
TETP youth accessibility fixes

pekkvil:


mattikl (approved):
These are all good improvements.

PR #1229, opened 2022-06-23
YJDH-574 | Eauth changed for Suomi.fi integration

charn:
## Description :sparkles:
This PR adds changes that are required to get the Eauthorization working with the Suomi.fi SAML authentication.

## Issues :bug:
YJDH-574 Suomi.fi integration


mattikl (approved):
Looks good. After merging TETP-253 with this, I will test that the HelsinkiOIDCAuthenticationBackend setup still works in TET.

nenonja:
as Juha is gone for vacation, I'll merge this

PR #1233, opened 2022-06-27
Yjdh 574 excel export adfs support

mikkovihonen:
Changed excel export view to use @enforce_handler_view_adfs_login instead of django admin method decorator


mattikl (approved):
Reviewed over Slack huddle.

PR #1248, opened 2022-06-29
TETP-249 Add ErrorText component. Remove test components from tet youth

pekkvil:


mattikl (approved):
Looks good. The error messages are now localized.

PR #1340, opened 2022-08-15
YJDH-642 | KS-Backend: Fix annual Talpa Excel export

karisal-anders:
### KS-Backend: Use streaming FileResponse in Excel export

Use a FileResponse in EmployerApplicationExcelDownloadView in order to
avoid getting a Gateway Timeout 504 error if the Excel export takes too
long.

### KS-Backend: Use select/prefetch_related in Excel export

EmployerApplicationExcelDownloadView:
 - Select and prefetch related objects in Excel export to hopefully
   speed up the query.

### KS-Backend: Refactor Excel exporting functions

EmployerApplicationExcelDownloadView:
 - Move shared query part from export_and_download_annual_applications
   and export_and_download_unhandled_applications to base_queryset

## Issues :bug:

[YJDH-642](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-642) ("BE: Kesäsetelin tämän vuoden hakemusten Talpa-Excelin lataaminen epäonnistuu yhdyskäytävän aikakatkaisuun")

Similar to the fix in https://github.com/City-of-Helsinki/yjdh/pull/1118

### DIFF
--- a/backend/kesaseteli/applications/tests/test_excel_export.py
+++ b/backend/kesaseteli/applications/tests/test_excel_export.py
@@ -5,6 +5,7 @@ from typing import List
 
 import openpyxl
 import pytest
+from django.http import StreamingHttpResponse
 from django.shortcuts import reverse
 from django.test import override_settings
 from django.urls.exceptions import NoReverseMatch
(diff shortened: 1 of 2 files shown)


nikomakela (approved):
by only looking at the code changes made to the files, I think the changes looks good. I'll setup a local environment to test it further (and to give better support next time). Since there are tests written and this is only some excel exporting, I think I can already approve this, even without knowing the actual product at all, so you can continue fixing the production 😄

PR #1361, opened 2022-08-25
YJDH-608: fix validation issues on form on both text inputs and date inputs

nenonja:
Check the [ticket](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-608) for more info


melniiv (approved):
from code perspective looks legit

PR #1407, opened 2022-09-27
Add help comments to root package json docker-compose scripts

nenonja:
## Description :sparkles:
While re-remebering myself what docker-compose scripts do, I added some helping comments to them.


karisal-anders (approved):
We discussed this but it'd be helpful information to include in the commit and/or PR that package.json does not support comments so adding comments to it is done by a workaround i.e. adding empty commands whose names are prefixed with "// " to give them the appearance of comments.

Also there are some typos.

Do these empty scripts have any side-effects i.e. are they totally safe and harmless to add?

Taking all of these into account I still approve this.

PR #1410, opened 2022-09-28
HL-562: Fix session expiration error

nenonja:
See details here: https://helsinkisolutionoffice.atlassian.net/browse/HL-562
This is hard to reproduce and probably concerns only local environment because user has changed to anonymous user after session has expired. The situation happens at least when the flag DISABLE_AUTHENTICATION is 1.

### DIFF
--- a/backend/benefit/applications/api/v1/serializers.py
+++ b/backend/benefit/applications/api/v1/serializers.py
@@ -95,7 +95,10 @@ class AttachmentField(FileField):
 
         url_pattern_name = "v1:applicant-application-download-attachment"
         request = self.context.get("request")
-        if request and get_request_user_from_context(self).is_handler():
+        user = get_request_user_from_context(self)
+        if settings.DISABLE_AUTHENTICATION or (
+            user and user.is_authenticated and user.is_handler()
+        ):
             url_pattern_name = "v1:handler-application-download-attachment"
 
         path = reverse(
@@ -463,7 +466,6 @@ class ApplicationBatchSerializer(serializers.ModelSerializer):
 
 
 class BaseApplicationSerializer(DynamicFieldsModelSerializer):
-
     """
     Fields in the Company model come from YTJ/other source and are not editable by user, and are listed
     in read_only_fields. If sent in the request, these fields are ignored.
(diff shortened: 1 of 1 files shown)


karisal-anders (commented):
This would take into account DISABLE_AUTHENTICATION=True and do the AnonymousUser removal more cleanly using [is_authenticated](https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated):
```diff
diff --git a/backend/benefit/applications/api/v1/serializers.py b/backend/benefit/applications/api/v1/serializers.py
index d4466404..7c2428e4 100644
--- a/backend/benefit/applications/api/v1/serializers.py
+++ b/backend/benefit/applications/api/v1/serializers.py
@@ -95,7 +95,10 @@ class AttachmentField(FileField):

         url_pattern_name = "v1:applicant-application-download-attachment"
         request = self.context.get("request")
-        if request and get_request_user_from_context(self).is_handler():
+        user = get_request_user_from_context(self)
+        if settings.DISABLE_AUTHENTICATION or (
+            user.is_authenticated and user.is_handler()
+        ):
             url_pattern_name = "v1:handler-application-download-attachment"

         path = reverse(
```

karisal-anders:
> This would take into account DISABLE_AUTHENTICATION=True and do the AnonymousUser removal more cleanly using [is_authenticated](https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated):
> 
> ```diff
> diff --git a/backend/benefit/applications/api/v1/serializers.py b/backend/benefit/applications/api/v1/serializers.py
> index d4466404..7c2428e4 100644
> --- a/backend/benefit/applications/api/v1/serializers.py
> +++ b/backend/benefit/applications/api/v1/serializers.py
> @@ -95,7 +95,10 @@ class AttachmentField(FileField):
> 
>          url_pattern_name = "v1:applicant-application-download-attachment"
>          request = self.context.get("request")
> -        if request and get_request_user_from_context(self).is_handler():
> +        user = get_request_user_from_context(self)
> +        if settings.DISABLE_AUTHENTICATION or (
> +            user.is_authenticated and user.is_handler()
> +        ):
>              url_pattern_name = "v1:handler-application-download-attachment"
> 
>          path = reverse(
> ```

Sorry, my bad, didn't take the `user is None` case into account.
So `user.is_authenticated and user.is_handler()` → `user and user.is_authenticated and user.is_handler()` would take care of that too.

karisal-anders (approved):

PR #1412, opened 2022-09-29
(WIP): HL-553: Error in manual calculator

nenonja:
## Description :sparkles:
HL-553: Error in manual calculator. Backend validation fails when both calculation abject and paysubsidies objecs are sent to backend. Let's strip the either object out before sending application.  Which one is stripped depends on manual calculation checkbox value.


nenonja:
This cannot be done by changing the frontend but instead the need is in the backend. See the comment on https://helsinkisolutionoffice.atlassian.net/browse/HL-553 , so I delete this pr.

PR #1483, opened 2022-10-27
HL-566 | Force pytest settings

karisal-anders:
### HL-Backend: Force settings for pytest runs 

Combine force_settings and setup_test_environment autouse fixtures into
the latter and add following settings to it:
 - settings.DISABLE_TOS_APPROVAL_CHECK = False
 - settings.NEXT_PUBLIC_MOCK_FLAG = False

Remove all now unnecessary uses of:
 - @override_settings(NEXT_PUBLIC_MOCK_FLAG=False)
 - @override_settings(DISABLE_TOS_APPROVAL_CHECK=False)

Refs HL-566
 
### HL-Backend: Add black/coverage/flake8/pep8/pytest excludes 

Add pyproject.toml to configure black because black AFAIK can not be
configured using setup.cfg at the moment.

Exclude the following directories in black/coverage/flake8/pep8/pytest:
 - .pytest_cache
 - var/media

This speeds up
 - "black . && isort . && flake8" from minutes to ~10 seconds
 - "pytest . --collect-only" from minutes to ~6 seconds
in Docker on Windows 10 using Docker Desktop and WSL 2. I saw around
100k files in /app/var/media (i.e. the MEDIA_ROOT directory) at one
point locally when testing so excluding a directory with that many files
helps.

Refs HL-566

## Issues :bug:

HL-566

### DIFF
--- a/backend/benefit/applications/tests/test_application_batch_api.py
+++ b/backend/benefit/applications/tests/test_application_batch_api.py
@@ -8,7 +8,6 @@ import pytest
 import pytz
 from django.conf import settings
 from django.http import StreamingHttpResponse
-from django.test import override_settings
 from rest_framework.reverse import reverse
 
 from applications.api.v1.serializers import ApplicationBatchSerializer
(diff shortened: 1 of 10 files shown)


tuomas777 (approved):
LGTM 👍 🐳

PR #1486, opened 2022-11-01
HL-553, HL-566 | Update mock flag related permission handling & fix moving application to handling

karisal-anders:
### Shared backend: Add is_staff & is_superuser support to OIDC mock user 

Add settings to control OIDC mock user's is_staff & is_superuser values:
 - settings.OIDC_MOCK_USER_IS_STAFF -> OIDC mock user's is_staff
   - If the setting is not set at all then will default to False
 - settings.OIDC_MOCK_USER_IS_SUPERUSER -> OIDC mock user's is_superuser
   - If the setting is not set at all then will default to False

The OIDC mock user is created by MockAuthenticationRequestView which is
mapped to oidc's authenticate endpoint
(See backend/shared/shared/oidc/urls.py) if NEXT_PUBLIC_MOCK_FLAG is set

Refs HL-566
 
### HL-Backend: Update NEXT_PUBLIC_MOCK_FLAG related permission handling 

BaseApplicationViewSet.get_queryset:
 - Refactor logical structure for readability, should not change effect

ApplicationStatus.is_editable_status:
 - Previously NEXT_PUBLIC_MOCK_FLAG allowed to edit all statuses. Now
   NEXT_PUBLIC_MOCK_FLAG allows to edit both handler's and applicant's
   statuses.
 - Update related test_is_editable_status_with_anonymous_user test

UserSerializer.is_terms_of_service_approval_needed:
 - Don't require terms of service to be approved when
   NEXT_PUBLIC_MOCK_FLAG is on

Refs HL-566
 
### Shared backend: Conditionally mock endpoints using MockEnabledProxyView 

Add MockEnabledProxyView and tests:
 - A view that changes on-the-fly between the real and mocked view based
   on the current value of NEXT_PUBLIC_MOCK_FLAG

Refs HL-566
 
### HL-Backend: Don't validate pay subsidies in RECEIVED to HANDLING change 

PaysubsidySerializer:
 - Make start_date and end_date validation errors specific to the fields
 - Skip start_date and end_date validation when application's status is
   changed from RECEIVED to HANDLING

Tests:
 - Add RECEIVED to HANDLING status change case to
   test_application_status_change_as_handler
 - Add test_ignore_pay_subsidy_dates_in_received_to_handling_transition
   test to check that pay subsidies' start and end dates can be empty
   when application's status changes from RECEIVED to HANDLING
 - Move duplicated test code to _set_two_pay_subsidies_with_empty_dates

Translations:
 - Update translations:
   - "python manage.py makemessages --no-location -l fi -l sv -l en"
 - Add Finnish and Swedish translations for
   - "Start date cannot be empty"
   - "End date cannot be empty"

Refs HL-553

## Issues :bug:

HL-553, HL-566

### DIFF
--- a/backend/benefit/applications/api/v1/views.py
+++ b/backend/benefit/applications/api/v1/views.py
@@ -108,9 +108,9 @@ class BaseApplicationViewSet(AuditLoggingModelViewSet):
 
     def get_queryset(self) -> QuerySet[Application]:
         user = self.request.user
-        if not settings.NEXT_PUBLIC_MOCK_FLAG and not user.is_authenticated:
-            return Application.objects.none()
-        return Application.objects.all().select_related("company", "employee")
+        if settings.NEXT_PUBLIC_MOCK_FLAG or user.is_authenticated:
+            return Application.objects.all().select_related("company", "employee")
+        return Application.objects.none()
 
     EXCLUDE_FIELDS_FROM_SIMPLE_LIST = [
         "applicant_terms_approval",
--- a/backend/benefit/applications/enums.py
+++ b/backend/benefit/applications/enums.py
@@ -17,7 +17,9 @@ class ApplicationStatus(models.TextChoices):
     @classmethod
     def is_editable_status(cls, user, status):
         if settings.NEXT_PUBLIC_MOCK_FLAG:
-            return True
+            return cls.is_handler_editable_status(
+                status, None
+            ) or cls.is_applicant_editable_status(status)
         elif not user.is_authenticated:
             return False
         elif user.is_handler():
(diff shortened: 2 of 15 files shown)


nenonja (approved):
We mostly looked this together with @karisal-anders and only debate was about the `MockEnabledProxyView` wheter should proxy design pattern should be used or factory design pattern, but both do the same and avoid that the view code would be splitted in if-else clauses. 
On the other that that It would be good to test these changes throughfully, that every state transition works in applications. One good case to test  is to move application NEED_MORE_INFO state and then back to HANDLING -state and all goes well without validation errors.

PR #1524, opened 2022-11-17
latest features to main

nenonja:


karisal-anders (approved):
These develop→main branch merges are in my view not necessary to code review as the pull requests these are based on should have already been code reviewed and the way the main branch is being used in this repository is as a snapshot of the develop branch (i.e. the main branch should be identical to a specific version of the develop branch although I know from experience that this has not always been so, but if it isn't then that should be corrected i.e. the differences from main should be incorporated back into develop branch so that main would be identical to a version of develop). Of course it's good to check that one doesn't break things etc but I leave the responsibility for checking this pull request out in the degree you see necessary up to you and I approve this with that in mind—please take that into account before continuing with the merge!

PR #1566, opened 2022-11-30
HL-587 | TET-Backend: Use organization name as a fallback for showing user name

karisal-anders:
### TET-Backend: Use organization name as a fallback for showing user name

Give a reasonable fallback when GDPR data has been removed

## Issues :bug:

HL-587

### DIFF
--- a/backend/tet/tet/views.py
+++ b/backend/tet/tet/views.py
@@ -18,13 +18,16 @@ class UserInfoView(View):
                 "given_name": user.first_name,
                 "family_name": user.last_name,
                 "email": user.email,
-                "name": f"{user.first_name} {user.last_name}",
+                "name": f"{user.first_name} {user.last_name}".strip(),
                 "is_ad_login": user.is_staff,
                 "organization_name": get_organization_name(request),
             }
 
             if not (userinfo["given_name"] or userinfo["family_name"]):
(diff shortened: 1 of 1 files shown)


backend/tet/tet/views.py
```diff
@@ -18,14 +18,17 @@ def get(self, request):
                 "given_name": user.first_name,
                 "family_name": user.last_name,
                 "email": user.email,
-                "name": f"{user.first_name} {user.last_name}",
+                "name": f"{user.first_name} {user.last_name}".strip(),
                 "is_ad_login": user.is_staff,
                 "organization_name": get_organization_name(request),
             }
 
             if not (userinfo["given_name"] or userinfo["family_name"]):
```
nenonja, line 26: Should this be `if not (userinfo["given_name"] or not userinfo["family_name"]):`

karisal-anders: !(x or y) = !x and !y i.e. both are empty so it's just checking that there's no name information at all present and then falls back to using email.

nenonja: Oh I didnt see the brackets... perhaps !x and !y could be better if someone else is as blind as me 👓

nenonja (approved):

PR #1569, opened 2022-11-30
HL-581 | Add GDPR API to Helsinki-lisä

karisal-anders:
## Description :sparkles:

This is at the point where this should be tested in test environment.

## Issues :bug:

HL-581


sirtawast:
The branch could go through a rebase 💅  as the commits are a bit incoherent.

backend/benefit/users/api/v1/views.py
```diff
@@ -31,3 +40,33 @@ def _get_current_user(self, request):
...
+class UserUuidGDPRAPIView(GDPRAPIView):
+    """
+    GDPR-API view that is used from Helsinki-profiili to query what data this app has
+    on person.
+    """
+
+    permission_classes = [BFGDPRScopesPermission]
+    authentication_classes = []
+
+    def get_object(self) -> AbstractBaseUser:
```
mjturt, line 54: So, this works now that if you are authenticated as someone and know someone else Helsinki-profile UUID, you can query that person's data. Of course, this is not production ready, but it would be nice to see at this point that the auth even works in a testing environment at all.

backend/benefit/users/tests/test_gdpr_api.py
```diff
@@ -0,0 +1,42 @@
...
+
+def extract_val(resp: dict, key: str) -> str:
+    """Extract value from GDPR API response."""
+    user = resp["children"]
+    value_pair = next(item for item in user if item["key"] == key)
+    return value_pair["value"]
+
+
+@override_settings(NEXT_PUBLIC_MOCK_FLAG=True)
+def test_get_profile_data_from_gdpr_api(api_client):
```
mjturt, line 22: I'm going to add authentication tests when confirmed that it works in testing environment.

backend/shared/shared/common/tests/factories.py
```diff
@@ -9,7 +9,7 @@ class UserFactory(factory.django.DjangoModelFactory):
 
     first_name = factory.Faker("first_name")
     last_name = factory.Faker("last_name")
-    username = factory.Faker("user_name")
+    username = factory.Faker("uuid4")
```
mjturt, line 12: As the Helsinki-profile UUID is saved as `username` it makes sense that our mock-data uses UUID there as well. BTW. @karisal-anders you know why is it like that? :D

backend/shared/shared/oidc/views/mock_views.py
```diff
@@ -64,6 +66,7 @@ def get(self, request, *args, **kwargs):
             user = UserFactory(
                 is_staff=getattr(settings, "OIDC_MOCK_USER_IS_STAFF", False),
                 is_superuser=getattr(settings, "OIDC_MOCK_USER_IS_SUPERUSER", False),
+                username=uuid4(),
```
mjturt, line 69: Same here.

mjturt (commented):
Co-author comments.

backend/shared/shared/common/tests/factories.py
karisal-anders: Helsinki-profile has two different UUID values:
 - Helsinki profile ID
 - User ID

The shared backend's HelsinkiOIDCAuthenticationBackend.create_user will create a Django user whose username contains the [sub claim](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) of the OIDC which is the user ID, not the Helsinki profile ID. The Helsinki profile ID is not in the OIDC claims at all but can be read using some other query—HELSINKI_PROFILE_API_URL is probably related to this but might not currently be used for that purpose but for getting a user's national identity number instead AFAIK [1]. Some discussion on the profile ID and user ID in city of Helsinki's slack channel #helsinki-profiili: [#1](https://helsinkicity.slack.com/archives/CB3HG29K8/p1669637782539189?thread_ts=1669635753.777089&cid=CB3HG29K8), [#2](https://helsinkicity.slack.com/archives/CB3HG29K8/p1669638242415039?thread_ts=1669635753.777089&cid=CB3HG29K8).

[1] backend/README.md:
> call `settings.HELSINKI_PROFILE_API_URL` with the access token [to get the national identification number](https://github.com/City-of-Helsinki/yjdh/blob/develop/backend/shared/shared/helsinki_profile/hp_client.py#L24)

karisal-anders:
I don't know whether you are aware of the GDPR API tester but it doesn't hurt to mention it as I have heard good things about it—haven't used it myself though:
 - Announcement on #helsinki-profiili: https://helsinkicity.slack.com/archives/CB3HG29K8/p1674115908002619
 - GDPR API Description: https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/80969736/GDPR+API+Description
 - **GDPR API Tester**: https://github.com/City-of-Helsinki/profile-gdpr-api-tester/
 - Example backend: https://github.com/City-of-Helsinki/example-backend-profile/
 - Django app for implementing Helsinki profile GDPR API: https://github.com/City-of-Helsinki/helsinki-profile-gdpr-api

backend/shared/shared/common/tests/factories.py
karisal-anders: I don't know why the Helsinki profile's user ID—not the profile ID but the user ID—is passed in the sub claim, this documentation may possibly shed some light on this matter: https://github.com/City-of-Helsinki/open-city-profile/blob/develop/open_city_profile/static/open-city-profile/gdpr-api-openapi.yaml

backend/benefit/users/api/v1/views.py
```diff
@@ -31,3 +40,33 @@ def _get_current_user(self, request):
...
+        """Get user by Helsinki-profiili UUID that is stored as username."""
```
karisal-anders, line 55: "Helsinki-profiili UUID" is very slightly inaccurate description as this is the user ID received from Helsinki profiili and that user ID is an UUID, see https://github.com/City-of-Helsinki/yjdh/pull/1569/files#r1094424211

Maybe "Helsinki-profiili user ID" or "Helsinki profile user ID" would be more accurate. If you wanted you could also document that there's a "Helsinki profile profile ID" also which is a different UUID value than the "Helsinki profile user ID" just to make it exactly clear to the user what's going on here, but that's up to you.

backend/benefit/users/tests/test_gdpr_api.py
```diff
@@ -0,0 +1,42 @@
...
+from users.models import format_date
+
+User = get_user_model()
+
+
+def extract_val(resp: dict, key: str) -> str:
+    """Extract value from GDPR API response."""
+    user = resp["children"]
+    value_pair = next(item for item in user if item["key"] == key)
+    return value_pair["value"]
```
karisal-anders, line 18: "Explicit is better than implicit" from Zen of Python could be used to argue here that not abbreviating the variable names here would make for more idiomatic Python code, e.g. resp → response (→ gdpr_api_response just to bring more semantic meaning to the parameter) and extract_val → extract_value. Now the docstring is more easily in line with the name of the function and its parameters and the code is thus more easily understandable IMHO. Also combining the last two lines of the function into one would make the function shorter but that's a bit arguable whether it reads better or not:
```python
extract_value(gdpr_api_response: dict, key: str) -> str:
    """Extract value from GDPR API response."""
    user: dict = gdpr_api_response["children"]
    return next(item["value"] for item in user if item["key"] == key, None)
```
How do you see this?

backend/shared/shared/common/tests/factories.py
karisal-anders: About changing the UserFactory's username to uuid4 for all use cases... hmm... I'm not sure whether all paths where users are created use username that is uuid4. Maybe it would be better to have e.g. `HelsinkiProfileUserFactory` which would be:
```python
class HelsinkiProfileUserFactory(UserFactory):
    username = factory.Faker("uuid4")  # Helsinki profile linked users' username is an uuid4
```
and keep the `UserFactory` as is. What do you think?

mjturt (commented):
Author comments #2.

backend/benefit/users/api/v1/views.py
```diff
@@ -31,3 +40,30 @@ def _get_current_user(self, request):
         if settings.NEXT_PUBLIC_MOCK_FLAG and isinstance(request.user, AnonymousUser):
             set_mock_user_name(request.user)
         return request.user
+
+
+class UserUuidGDPRAPIView(GDPRAPIView):
+    """GDPR API view that is used from Helsinki profile to query and delete user data."""
```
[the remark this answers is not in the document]
karisal-anders: Personally I don't like that multiline format. I'd much rather have this:
```python
"""
Multiline
comment
"""
```
but as this is a personal preference and the documentation you gave was not directly stating that these are the docstring formats that should be used but rather hinted at them through examples I would rather keep this lax myself i.e. not mandate a certain docstring format in the repository. What do you think?

backend/benefit/users/tests/test_gdpr_api.py
```diff
@@ -0,0 +1,46 @@
...
+    """Test that the profile data is valid without authentication."""
+    user = HelsinkiProfileUserFactory()
+    response = api_client.get(reverse("gdpr_v1", kwargs={"uuid": user.username}))
+    response_dict = json.loads(response.content)
+    assert response.status_code == 200
+    assert extract_value(response_dict, "FIRST_NAME") == user.first_name
+    assert extract_value(response_dict, "LAST_NAME") == user.last_name
+    assert extract_value(response_dict, "EMAIL") == user.email
+    assert extract_value(response_dict, "DATE_JOINED") == format_date(
+        user.date_joined  # type: ignore
```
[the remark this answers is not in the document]
karisal-anders: Isn't this a [mypy related ignore](https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker)? Are you running the project through mypy? I haven't used mypy myself and I haven't seen it used much in the projects I've worked on so far. Does it work in the yjdh repository currently? Doesn't it give a lot of warnings if run? It might be interesting to include it in a project but as I have no experience with it I don't know what are the hurdles one has to go to through to really get a project work with it without any errors or warnings. We could have a talk about this as I am interested to hear about your experience with mypy if you have worked with it?

.pre-commit-config.yaml
```diff
@@ -1,21 +1,25 @@
+fail_fast: true
```
mjturt, line 1: With python tooling, there are some issues. Isort configuration isn't used on pipelines and causes couple configuration mismatch issues. Needs to be fixed in another pull-request.

Also, we have husky and pre-commit on top of each other. I am uncertain if this is the best solution.

backend/benefit/users/api/v1/authentications.py
```diff
@@ -0,0 +1,26 @@
+"""
+Contents copied from tet.tet.views. More info there.
+As kesaseteli doesn't use helusers package, kesaseteli tests fail when these put
```
mjturt, line 3: Even if kesäseteli doesn't import the module from shared files, the tests fail.

backend/shared/shared/oidc/mixins.py
```diff
@@ -12,9 +12,9 @@ def __convert_amr_to_list(self, id_token):
         To fix the claim validation issue, convert the id_token amr to list
         when a string is given.
         """
-        if id_token["amr"] and not isinstance(id_token["amr"], list):
+        if id_token.get("amr") and not isinstance(id_token["amr"], list):
```
mjturt, line 15: `amr` wasn't there when testing with gdpr-api-tester so added this fail proofing.

mjturt (commented):
Comments 3.

backend/benefit/users/tests/test_gdpr_api.py
```diff
@@ -1,41 +1,45 @@
...
 
-from shared.common.tests.factories import UserFactory
+from shared.common.tests.factories import HelsinkiProfileUserFactory
 from users.models import format_date
 
 User = get_user_model()
 
 
-def extract_val(resp: dict, key: str) -> str:
+def extract_value(gdpr_api_response: dict, key: str) -> Optional[str]:
```
rikuke: As much as I would like to use type hints myself, their use is not very consistent at the moment, so perhaps we should make decision to either use them everywhere or nowhere.

mjturt: I would like to see them more, also :+1: I think they should be used as much as possible. [Helsinki city's coding standards](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/OS/pages/7737901085/Koodausstandardit+coding+standards) also suggests that they should be used.

There are cases of course where they add maybe too little value for the effort. For example, some test class test methods. Also, some libraries like `factories` have poor typing support, especially with Django models.

I would say that all the new and touched code should primarily have them, but leave option for not using them, depending on the case.

Btw. this specific function is out of date and is removed in current head.

mjturt:
@nenonja Is the failing acceptance test here just the same you mentioned that has been broken for ages?

rikuke (approved):

PR #1581, opened 2022-12-05
YJDH-651 | KS-Backend: Add Excel export for confirmed youth summer applications

karisal-anders:
### KS-Backend: Add confirmed youth applications Excel export & tests 

Add new tests:
 - test_youth_excel_download_no_youth_applications
 - test_youth_excel_download_writes_audit_log
 - test_youth_excel_download_content

Add parametrization to extisting test:
 - test_excel_view_download_with_unauthenticated_user

Also:
 - Disable VTJ in test_youth_applications_detail_valid_pk test to fix
   the test case
 - Add audit log support for list function in AuditLoggingModelViewSet
   - AuditLoggingModelViewSet inherits ModelViewSet which inherits
     ListModelMixin which has a list function so it makes sense to add
     an audit log entry also for list calls

Refs YJDH-651
 
### All backends: Extend and synchronize black/flake/pep8/pytest settings 

Similar to change in b1084fb. Should
speed up running "black . && isort . && flake8" because of excluding
e.g. var/media also in the kesaseteli backend.

Now:
 - Identical pyproject.toml files for configuring black present in:
   - benefit
   - kesaseteli
   - tet
 - setup.cfg files identical between benefit/kesaseteli/tet/shared
   except for DJANGO_SETTINGS_MODULE
 - precommit-setup.cfg identical to shared/setup.cfg

Refs YJDH-651

## Issues :bug:

YJDH-651

## Testing :alembic:

Steps:
 - Set `NEXT_PUBLIC_MOCK_FLAG=True` to `.env.kesaseteli`
 - `yarn youth:up --build`
 - Generate some youth applications test data using Django shell, see "Generating test data" below
 - Open https://localhost:8000/excel-download/
 - Click on "Lataa Excel kaikista nuorten vahvistetuista kesäsetelihakemuksista" button
 - Open the downloaded file "Nuorten kesäsetelihakemukset-YYYY-MM-DD.xlsx" where YYYY-MM-DD is today

### Generating test data:
`docker exec -it kesaseteli-backend python manage.py shell_plus`:
```python
from common.tests.factories import *
InactiveYouthApplicationFactory()  # This should not be exported
apps = (
	ActiveYouthApplicationFactory.create_batch(size=12)
	+ [
		YouthApplicationFactory(status=status)
		for status in YouthApplicationStatus.active_values()
	]
	+ [
		ActiveVtjTestCaseYouthApplicationFactory(last_name=vtj_test_case)
		for vtj_test_case in VtjTestCase.values
	]
)
# Update created_at specially as its initial save uses auto_now_add
for app in apps:
	app.created_at -= timedelta(seconds=random.randint(0, 60 * 60 * 24 * 10))
	app.save(update_fields=["created_at"])
	app.refresh_from_db()
```

### DIFF
--- a/backend/kesaseteli/applications/api/v1/serializers.py
+++ b/backend/kesaseteli/applications/api/v1/serializers.py
@@ -1,8 +1,11 @@
 import json
+from datetime import datetime
+from typing import Optional, Union
 
 import filetype
 from django.conf import settings
 from django.db import transaction
+from django.utils import timezone, translation
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
 from PIL import Image, UnidentifiedImageError
@@ -604,4 +607,89 @@ class YouthApplicationHandlingSerializer(serializers.ModelSerializer):
         model = YouthApplication
         fields = [
             "encrypted_handler_vtj_json",
+        ]
+
+
+class YouthApplicationExcelExportSerializer(serializers.ModelSerializer):
+    application_year = serializers.SerializerMethodField()
+    application_date = serializers.SerializerMethodField()
+    birth_year = serializers.IntegerField(source="birthdate.year")
+    birthdate = serializers.SerializerMethodField()
+    confirmation_date = serializers.SerializerMethodField()
+    handling_date = serializers.SerializerMethodField()
+    additional_info_providing_date = serializers.SerializerMethodField()
+    additional_info_user_reasons = serializers.SerializerMethodField()
+    summer_voucher_serial_number = serializers.SerializerMethodField()
+    is_unlisted_school = serializers.SerializerMethodField()
+
+    @staticmethod
+    def to_isoformat_localdate(value: datetime) -> str:
+        return timezone.localdate(value).isoformat() if value else ""
+
+    @staticmethod
+    def to_local_year(value: datetime) -> int:
+        return timezone.localdate(value).year
+
+    def get_summer_voucher_serial_number(self, obj: YouthApplication) -> Optional[int]:
+        if not obj.has_youth_summer_voucher:
+            return None
+        return obj.youth_summer_voucher.summer_voucher_serial_number
+
+    def get_birthdate(self, obj: YouthApplication) -> str:
+        return obj.birthdate.isoformat()
+
+    def get_application_year(self, obj: YouthApplication) -> int:
+        return self.to_local_year(obj.created_at)
+
+    def get_application_date(self, obj: YouthApplication) -> str:
+        return self.to_isoformat_localdate(obj.created_at)
+
+    def get_confirmation_date(self, obj: YouthApplication) -> str:
+        return self.to_isoformat_localdate(obj.receipt_confirmed_at)
+
+    def get_additional_info_providing_date(self, obj: YouthApplication) -> str:
+        return self.to_isoformat_localdate(obj.additional_info_provided_at)
+
+    def get_handling_date(self, obj: YouthApplication) -> str:
+        return self.to_isoformat_localdate(obj.handled_at)
+
+    def get_additional_info_user_reasons(self, obj: YouthApplication) -> str:
+        return ", ".join(sorted(obj.additional_info_user_reasons))
+
+    def get_is_unlisted_school(self, obj: YouthApplication) -> str:
+        with translation.override("fi"):
+            return str(_("Kyllä") if obj.is_unlisted_school else _("Ei"))
+
+    @staticmethod
+    def get_placeholder_value(field: str) -> Union[int, str]:
+        if field in ["application_year", "birth_year"]:
+            return 1  # Placeholder integer value
+        else:
+            return "placeholder_value"
+
+    class Meta:
+        model = YouthApplication
+        fields = read_only_fields = [
+            "additional_info_description",
+            "additional_info_providing_date",
+            "additional_info_user_reasons",
+            "application_date",
+            "application_year",
+            "birth_year",
+            "birthdate",
+            "confirmation_date",
+            "email",
+            "first_name",
+            "handling_date",
+            "id",
+            "is_unlisted_school",
+            "language",
+            "last_name",
+            "phone_number",
+            "postcode",
+            "school",
+            "status",
+            "summer_voucher_serial_number",
+            "vtj_home_municipality",
+            "vtj_last_name",
         ]
--- a/backend/kesaseteli/applications/models.py
+++ b/backend/kesaseteli/applications/models.py
@@ -675,17 +675,17 @@ class YouthApplication(LockForUpdateMixin, TimeStampedModel, UUIDModel):
         )
 
     @property
-    def vtj_last_name(self):
+    def vtj_last_name(self) -> Optional[str]:
         if vtj_last_names := self._vtj_values("$.Henkilo.NykyinenSukunimi.Sukunimi"):
             return vtj_last_names[0]
-        return None
+        return ""
 
     @property
     def attends_helsinkian_school(self) -> bool:
         return not self.is_unlisted_school
 
     @property
-    def vtj_home_municipality(self):
+    def vtj_home_municipality(self) -> Optional[str]:
         if vtj_home_municipality := self._vtj_values("$.Henkilo.Kotikunta.KuntaS"):
             return vtj_home_municipality[0]
         return ""
--- a/backend/kesaseteli/applications/tests/test_excel_export.py
+++ b/backend/kesaseteli/applications/tests/test_excel_export.py
@@ -1,4 +1,5 @@
-import datetime
+import random
+from datetime import date, datetime, timedelta
 from decimal import Decimal
 from io import BytesIO
 from typing import List
@@ -9,9 +10,16 @@ from django.http import StreamingHttpResponse
 from django.shortcuts import reverse
 from django.test import override_settings
 from django.urls.exceptions import NoReverseMatch
+from django.utils import translation
+from django.utils.timezone import localdate
 from freezegun import freeze_time
 
-from applications.enums import EmployerApplicationStatus, ExcelColumns
+from applications.enums import (
+    EmployerApplicationStatus,
+    ExcelColumns,
+    VtjTestCase,
+    YouthApplicationStatus,
+)
 from applications.exporters.excel_exporter import (
     APPLICATION_LANGUAGE_FIELD_TITLE,
     EMPLOYMENT_END_DATE_FIELD_TITLE,
@@ -34,17 +42,27 @@ from applications.exporters.excel_exporter import (
     SPECIAL_CASE_FIELD_TITLE,
     WORK_HOURS_FIELD_TITLE,
 )
-from applications.models import EmployerSummerVoucher
+from applications.models import EmployerSummerVoucher, YouthApplication
 from applications.tests.test_models import create_test_employer_summer_vouchers
+from applications.views import YouthApplicationExcelExportViewSet
 from common.tests.factories import (
+    ActiveVtjTestCaseYouthApplicationFactory,
+    ActiveYouthApplicationFactory,
     EmployerApplicationFactory,
     EmployerSummerVoucherFactory,
+    InactiveYouthApplicationFactory,
+    YouthApplicationFactory,
 )
 from common.urls import handler_403_url
+from shared.audit_log.models import AuditLogEntry
 
 
 def excel_download_url():
     return reverse("excel-download")
+
+
+def youth_excel_download_url():
+    return reverse("youth-excel-download")
 
 
 def get_field_titles(fields: List[ExcelField]) -> List[str]:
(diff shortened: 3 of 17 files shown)


backend/kesaseteli/applications/models.py
```diff
@@ -675,17 +675,17 @@ def is_applicant_dead_according_to_vtj(self) -> bool:
         )
 
     @property
-    def vtj_last_name(self):
+    def vtj_last_name(self) -> Optional[str]:
```
nenonja, line 678: Looks for me that the return value was previously optional string but now it's just string because `None` was replaced with "". Or could `vtj_last_names[0]` still return `None`?

nenonja: Also the [documentation](https://docs.python.org/3/library/typing.html#typing.Optional) says that Optional[str] could be also expressed as `str | None` which is at least for me a lot cleaner

backend/kesaseteli/applications/tests/test_excel_export.py
```diff
@@ -257,6 +282,110 @@ def employer_summer_voucher_sorting_key(voucher: EmployerSummerVoucher):
...
+        expected_output_columns_names = (
+            YouthApplicationExcelExportView.output_column_names()
+        )
+    assert output_column_names == expected_output_columns_names
+
+    source_fields = YouthApplicationExcelExportView.source_fields()
+    for data_row, app in zip(data_rows, apps):
+        for output_cell, source_field in zip(data_row, source_fields):
+            output_value = output_cell.value
+            if source_field == "application_date":
```
nenonja, line 340: Does python support switch-statement instead of multiple ifs? I think it would be more readable

backend/kesaseteli/applications/tests/test_excel_export.py
```diff
@@ -257,6 +282,110 @@ def employer_summer_voucher_sorting_key(voucher: EmployerSummerVoucher):
...
+                        int(output_value)
+                        == app.youth_summer_voucher.summer_voucher_serial_number
+                    )
+            elif source_field == "birth_year":
+                assert int(output_value) == app.birthdate.year
+            elif source_field == "birthdate":
+                assert date.fromisoformat(output_value) == app.birthdate
+            elif source_field == "is_unlisted_school":
+                assert output_value in ["Kyllä", "Ei"]
+                assert output_value != "Ei" or not app.is_unlisted_school
```
nenonja, line 360: double negations, without negations would be more readable?
`output_value == "Kyllä" or app.is_unlisted_school`

backend/kesaseteli/applications/tests/test_youth_applications_api.py
```diff
@@ -414,6 +414,7 @@ def test_youth_applications_process_valid_pk(
         )
 
 
+@override_settings(NEXT_PUBLIC_DISABLE_VTJ=True)
```
nenonja, line 417: Why this? Something related has changed?

backend/kesaseteli/applications/views.py
```diff
@@ -163,3 +170,123 @@ def export_and_download_annual_applications(
...
+        return YouthApplication.objects.active().order_by("created_at", "pk")
+
+    def serializer(self, apps: QuerySet[YouthApplication]):
+        return (self.generate_data_row(app) for app in apps)
+
+    @enforce_handler_view_adfs_login
+    def get(
+        self, request, *args, **kwargs
+    ) -> Union[HttpResponse, HttpResponseRedirect, StreamingHttpResponse]:
+        queryset = self.get_queryset()
```
nenonja, line 230: Should we log to audit log about who has downloaded the excel especially because it has sensitive information about the youth persons?

backend/kesaseteli/applications/views.py
```diff
@@ -163,3 +170,123 @@ def export_and_download_annual_applications(
...
+                xlsx_template=self.xlsx_template(queryset),
+                serializer=self.serializer,
+                batch_size=settings.EXCEL_DOWNLOAD_BATCH_SIZE,
+            ),
+            content_type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        )
+        response["Content-Disposition"] = f"attachment; filename={self.xlsx_filename}"
+        return response
+
+    @property
```
nenonja, line 246: Maybe this unnecessary nagging, but I would probably move the excel-related functions to another excel-utility class because of S in [SOLID principles](https://en.wikipedia.org/wiki/SOLID) eg. "every class should have only one responsibility"

backend/kesaseteli/applications/views.py
nenonja: Not sure if the logging should be here or `urls.py`

nenonja (changes requested):
Some comments.

nenonja:
I downloaded the excel and usability-wise I noticed that some enum values were not translated. If it's okay to product owners and they understand the enum values then there is no reason to translate them which probably would need some work.

And if we decide to implement the translations to enums, it's good to know that frontend has translated at least some of them, so the same translations could be used.

backend/kesaseteli/applications/tests/test_youth_applications_api.py
karisal-anders: This function's test cases using staff_client/superuser_client and NEXT_PUBLIC_MOCK_FLAG=False [weren't passing](https://github.com/City-of-Helsinki/yjdh/actions/runs/3639752757/jobs/6143534465) so I added this, I would think it's not related to any code changes in this pull request i.e. that these cases should've been broken earlier already.

It's because of [fetch_vtj_json](https://github.com/City-of-Helsinki/yjdh/blob/5f932b6c25499f9c8ef8aa7827e72b612a6c9573/backend/kesaseteli/applications/models.py#L267) being called which needs NEXT_PUBLIC_DISABLE_VTJ=True or NEXT_PUBLIC_MOCK_FLAG=True to not use real VTJ integration and when NEXT_PUBLIC_MOCK_FLAG=False and ks-review.yml has [NEXT_PUBLIC_DISABLE_VTJ: 0](https://github.com/City-of-Helsinki/yjdh/blob/5f932b6c25499f9c8ef8aa7827e72b612a6c9573/.github/workflows/ks-review.yml#L46) the test cases used real VTJ integration which failed in review environment because the VTJ integration wasn't properly configured there. IMHO it's a good idea not to use real VTJ integration in this test so ensuring that it is not used is in my opinion good.

backend/kesaseteli/applications/tests/test_excel_export.py
karisal-anders: I was trying to use logical implication here i.e. a ⇒ b <=> !a or b:
```
assert output_value != "Ei" or not app.is_unlisted_school
assert output_value != "Kyllä" or app.is_unlisted_school
```
If Python had logical implication operator I could have written:
```
assert (output_value == "Ei") ⇒ not app.is_unlisted_school
assert (output_value == "Kyllä") ⇒ app.is_unlisted_school
```

But I can actually combine the two cases into a more concise and clearer assertion because we've checked already that output_value must be either "Kyllä" or "Ei":
```
assert (output_value == "Kyllä") == app.is_unlisted_school
```

backend/kesaseteli/applications/models.py
karisal-anders: > Looks for me that the return value was previously optional string but now it's just string because `None` was replaced with "". Or could `vtj_last_names[0]` still return `None`?

I left the Optional there exactly because I was not 100% sure whether vt_last_names[0] will always return a string or could it possibly return None.

backend/kesaseteli/applications/tests/test_excel_export.py
karisal-anders: There's no switch statement, but if we used Python 3.10 or later we could perhaps use [match/case](https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching) but [we're using Python 3.8](https://github.com/City-of-Helsinki/yjdh/blob/5f932b6c25499f9c8ef8aa7827e72b612a6c9573/backend/docker/kesaseteli.Dockerfile#L2). Also some kind of a dictionary lookup could perhaps be used instead and the checks be put into functions but it depends on the code whether that would be readable or not. Having a lot of if/elif/else statements and nested ones in them does make the logic a bit hard to follow.

backend/kesaseteli/applications/views.py
karisal-anders: I'd rather not refactor the Excel exporting code in this pull request as there is a lot of overlap with the employer applications' Excel export in the same file. I knew when making this code that there's going to be overlap and code duplication but it was clearer and less time costly to do it this way now rather than try to refactor the two Excel export functionalities together in a clean way—I did try to combine them together first but that implementation became convoluted and I decided against using it. So yes, this code could be improved upon but I'd rather not go into it in this pull request now to save time as there's a lot that could be combined between `YouthApplicationExcelExportView` and `EmployerApplicationExcelDownloadView`.

karisal-anders:
> I downloaded the excel and usability-wise I noticed that some enum values were not translated. If it's okay to product owners and they understand the enum values then there is no reason to translate them which probably would need some work.

I showed the output of the Excel export to the product owners and they were ok with it so I think it's ok. The enum values themselves are not very clear even with translations, e.g. AWAITING_MANUAL_PROCESSING was only used when additional info request and automatic accepting was not implemented yet.

backend/kesaseteli/applications/views.py
karisal-anders: That's a legitimate point, I'll see about how to add some audit logging here...

backend/kesaseteli/applications/tests/test_excel_export.py
karisal-anders: Combined the asserts to a single one as above

backend/kesaseteli/applications/views.py
karisal-anders: Added audit logging

backend/kesaseteli/applications/api/v1/serializers.py
```diff
@@ -605,3 +608,97 @@ class Meta:
...
+    additional_info_providing_date = serializers.SerializerMethodField(
+        "get_additional_info_providing_date"
+    )
+    additional_info_user_reasons = serializers.SerializerMethodField(
+        "get_additional_info_user_reasons"
+    )
+    summer_voucher_serial_number = serializers.SerializerMethodField(
+        "get_summer_voucher_serial_number"
+    )
+    is_unlisted_school = serializers.SerializerMethodField("get_is_unlisted_school")
```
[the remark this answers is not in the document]
karisal-anders: Removed the unnecessary parameters

backend/kesaseteli/applications/views.py
```diff
@@ -163,3 +170,123 @@ def export_and_download_annual_applications(
...
+    def source_fields() -> Tuple[str, ...]:
+        return list(
+            zip(*YouthApplicationExcelExportView.source_fields_and_output_names())
+        )[0]
+
+    @staticmethod
+    def output_column_names() -> Tuple[str, ...]:
+        return list(
+            zip(*YouthApplicationExcelExportView.source_fields_and_output_names())
+        )[1]
```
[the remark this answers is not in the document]
karisal-anders: I didn't use dict because I thought it doesn't preserve the order but googling told that from [Python 3.7](https://docs.python.org/3/whatsnew/3.7.html) onwards
> the insertion-order preservation nature of dict objects has been declared to be an official part of the Python language spec.

so it could be used as we're using [Python 3.8](https://github.com/City-of-Helsinki/yjdh/blob/5f932b6c25499f9c8ef8aa7827e72b612a6c9573/backend/docker/kesaseteli.Dockerfile#L2) but I went with using OrderedDict to make it explicit that the preservation of the order of the insertion of the keys is needed i.e. the functionality will work wrong—the Excel output's columns will be in wrong order—without it.

backend/kesaseteli/applications/api/v1/serializers.py
```diff
@@ -605,3 +608,97 @@ class Meta:
         fields = [
             "encrypted_handler_vtj_json",
         ]
+
+
+class YouthApplicationExcelExportSerializer(serializers.ModelSerializer):
+    application_year = serializers.SerializerMethodField("get_application_year")
+    application_date = serializers.SerializerMethodField("get_application_date")
+    birth_year = serializers.SerializerMethodField("get_birth_year")
```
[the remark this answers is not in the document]
karisal-anders: Thanks! Changed to IntegerField

karisal-anders:
Please could you re-review @nenonja & @tuomas777? Thanks!

backend/kesaseteli/applications/views.py
```diff
@@ -163,3 +173,140 @@ def export_and_download_annual_applications(
...
+
+    def update(self, request, *args, **kwargs):
+        return HttpResponse(status=status.HTTP_405_METHOD_NOT_ALLOWED)
+
+    @classmethod
+    def source_fields_and_output_names(cls) -> typing.OrderedDict[str, str]:
+        with translation.override("fi"):
+            return collections.OrderedDict(
+                [
+                    ("application_date", str(_("Hakupvm"))),
```
tuomas777, line 202: Probably no point to change as this works perfectly fine also, but just FYI I think in situations like this you could use `gettext` instead of `gettext_lazy` to get rid of those `str(...)`s.

tuomas777 (approved):
LGTM, great job! 👍 🚀🦎

backend/kesaseteli/applications/views.py
karisal-anders: Yep, knew that and thought about it a little bit but was a bit unclear how to go about it as gettext_lazy is imported normally as `_` in the codebase, only once gettext is imported as `_`. Simply using gettext directly without renaming could be one option but `gettext` vs. `str(_` + `)` ... hmm it's just 7 vs. 6 characters so its actually quite the same character countwise. Importing gettext as _ somewhere and gettext_lazy as _ somewhere else is would not be a very good idea IMHO as it would make the use of _(...) context specific and thus the code less easily understandable (i.e. having to check which version of gettext has been imported).

karisal-anders: Tried it out to see what using gettext directly would look like but didn't like the results i.e. they looked less readable to me. So, going with the str(_(...)) although not optimal.

nenonja (approved):

PR #1596, opened 2022-12-09
yjdh-649: improve dockerfile for platta

nenonja:
Referring to discussion here: https://app.slack.com/client/T02DHG0DR/threads/thread/C02EHHFPQPL-1670827189.380059
and Platta ticket: https://helsinkisolutionoffice.atlassian.net/browse/PLATTA-4420

What is done:
  - don't install lerna on production stage of dockerfile
  - start server without using yarn script


karisal-anders (approved):
Looks ok to me, probably good to get more details about what's going on with the Kesäseteli's two failing tests on Platta.

PR #1615, opened 2022-12-15
HL-573 | Shared/TET backend: Determine ADFS login with "ADFS login" group

karisal-anders:
### Shared/TET backend: Determine ADFS login with "ADFS login" group 

Create "ADFS login" group and add the user to it when logging in using
HelsinkiAdfsAuthCodeBackend. This determines whether user was logged in
using ADFS at some point or not. Use this in TET's UserInfoView to set
return value is_adfs_login.

Return Forbidden 403 HTTP error in TET's UserInfoView if user was logged
in using ADFS at some point but did not qualify as a handler at last
ADFS login.

Add documentation about azure_adfs's optional Django settings:
 - HANDLERS_GROUP_NAME:
   - This group's membership signifies that user qualified as a
     handler at last ADFS login.
   - If set and a user group by this name exists then will be added to
     the logged in user's groups if the user qualifies as a handler or
     removed if the user does not.
 - ADFS_LOGIN_GROUP_NAME:
   - This group's membership signifies that user has logged in using
     ADFS at some point.
   - A group with this name, or "ADFS login" if not set, will be added
     to the logged in user's groups.

## Issues :bug:

HL-573

### DIFF
--- a/backend/shared/shared/azure_adfs/auth.py
+++ b/backend/shared/shared/azure_adfs/auth.py
@@ -12,6 +12,31 @@ from django_auth_adfs.exceptions import MFARequired
 LOGGER = logging.getLogger(__name__)
 
 
+def adfs_login_group_name():
+    """
+    Name of the user group that signifies that the user has logged in using ADFS i.e.
+    HelsinkiAdfsAuthCodeBackend at some point
+
+    :return: Value of settings.ADFS_LOGIN_GROUP_NAME if it exists, otherwise
+             "ADFS login"
+    """
+    return getattr(django_settings, "ADFS_LOGIN_GROUP_NAME", "ADFS login")
+
+
+def is_adfs_login(user) -> bool:
+    """
+    Has user logged in using ADFS i.e. HelsinkiAdfsAuthCodeBackend at some point?
+
+    .. warning::
+       Return value of True does NOT mean that user is currently logged in and active
+       also, it only means that user has logged in using ADFS at some point in the past.
+
+    :return: True if user has logged in using HelsinkiAdfsAuthCodeBackend as some point,
+             otherwise False.
+    """
+    return user.groups.filter(name=adfs_login_group_name()).exists()
+
+
 class HelsinkiAdfsAuthCodeBackend(AdfsAuthCodeBackend):
     def get_member_objects_from_graph_api(
         self, user_oid: str, graph_api_access_token: str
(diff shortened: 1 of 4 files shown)


backend/shared/shared/azure_adfs/auth.py
```diff
@@ -12,6 +12,27 @@
...
+    Name of the user group that signifies that the user has logged in using ADFS i.e.
+    HelsinkiAdfsAuthCodeBackend
+
+    :return: Value of settings.ADFS_LOGIN_GROUP_NAME if it exists, otherwise
+             "ADFS login"
+    """
+    return getattr(django_settings, "ADFS_LOGIN_GROUP_NAME", "ADFS login")
+
+
+def is_adfs_login(user) -> bool:
```
nenonja, line 26: `was_logged_in_using_adfs` would be better name because user is not necessary active anymore

nenonja (approved):
LGTM. One comment but minor, so approved as well

backend/shared/shared/azure_adfs/auth.py
```diff
@@ -148,6 +169,13 @@ def get_graph_api_access_token(self, access_token):
         return response_json["access_token"]
 
     def assign_local_groups(self, user):
+        # Set group membership which tells if user was logged in using ADFS
+        adfs_login_group, _ = Group.objects.get_or_create(name=adfs_login_group_name())
+        if user.is_authenticated:
+            user.groups.add(adfs_login_group)
+        else:
+            user.groups.remove(adfs_login_group)
```
nenonja, line 177: You could put the same type of comment `# safety - shouldn't be ever needed` as on line 188

nenonja: or remove else clause on both cases

backend/shared/shared/azure_adfs/auth.py
karisal-anders: It was concluded after discussion that `is_adfs_login` is ok as e.g. `is_authenticated` is used by Django itself rather than e.g. `was_authenticated`. So, `is_adfs_login` is in line with the precedent set by Django.

backend/shared/shared/azure_adfs/auth.py
karisal-anders: Changed the comments ([#1](https://github.com/City-of-Helsinki/yjdh/pull/1615/commits/19e97161dffc6d3ff38cebfaac0122a447c7af35#diff-7bf3fa8dc236385602f1b56075d236b632c9a655d62aba3756e6146520598e40R181), [#2](https://github.com/City-of-Helsinki/yjdh/pull/1615/commits/19e97161dffc6d3ff38cebfaac0122a447c7af35#diff-7bf3fa8dc236385602f1b56075d236b632c9a655d62aba3756e6146520598e40L160-R193)) to be more in line with reality AFAIK.

PR #1623, opened 2022-12-16
HL-576: remove sentry auth token as it doesnt work and is unneeded

nenonja:
## Description :sparkles:
Sentry source maps doesnt work. It is probably because of the sentry auth token which is configured wrong on . Source maps doesnt need protection because source code is in public, so sentry auth token can be removed from `@sentry/nextjs` config: https://docs.sentry.io/platforms/javascript/guides/nextjs/sourcemaps/

`@sentry/nextjs will generate and upload source maps automatically, in order to enable errors to have readable stacktraces.`

See also the conversation: https://app.slack.com/client/T02DHG0DR/C012SR6PHSB/thread/C012SR6PHSB-1671108162.824849


karisal-anders (approved):
Approved. Is the comment
>  //   release, url, org, project, authToken, configFile, stripPrefix,

in frontend/next.config.js up to date after removing the authToken?

PR #1627, opened 2022-12-19
TET-backend: Set .env.tet.example HELSINKI_PROFILE_API_URL

karisal-anders:
## Issues :bug:

HL-595


nenonja (approved):
I tested that the link worked. Approved.

PR #1631, opened 2022-12-21
HL-600 | Add TET GDPR API

karisal-anders:
### All backends: Normalize case & sort requirements*.in file contents 

Normalize package names to all lowercase letters in requirements*.in
which should be safe because PEP 428 at
https://peps.python.org/pep-0426/#name says:
 - "All comparisons of distribution names MUST be case insensitive"

Sort the package names in requirements*.in to ascending order.

Afterwards ran:
 - pip-compile requirements.in
 - pip-compile requirements-dev.in
 - pip-compile requirements-prod.in
but the requirements*.txt files remained unchanged for all backends.

Refs HL-600 (Looked at requirements*.in when implementing this ticket)
 
### TET-Backend: Add django-helusers and drf-oidc-auth to requirements 

Afterwards ran:
 - pip-compile requirements.in
 - pip-compile requirements-dev.in
 - pip-compile requirements-prod.in

Refs HL-600
 
### TET-Backend: Add dummy TET GDPR API view callable using user UUID 

Examples of endpoints:
  - GET & DELETE: /gdpr-api/v1/user/60404d9b-9e75-43be-bdb4-aae6a1a2f9f6
    - 60404d9b-9e75-43be-bdb4-aae6a1a2f9f6 is user UUID & User.username

Refs HL-600

## Issues :bug:

HL-600

### DIFF
--- a/backend/benefit/requirements-dev.in
+++ b/backend/benefit/requirements-dev.in
@@ -3,12 +3,12 @@ black
 flake8
 ipython
 isort
+langdetect
+openpyxl # for the Excel testcases from the handlers
 pip-tools
 pytest
 pytest-cov
 pytest-django
 pytest-freezegun
 requests-mock
-Werkzeug
-openpyxl # for the Excel testcases from the handlers
-langdetect
+werkzeug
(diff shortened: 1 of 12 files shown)


nenonja (approved):
Approved... Just make sure to make all the tests pass. 
Also we discussed that the time schedule is so tight that let's skip the tests this time.

PR #1681, opened 2023-01-02
HL-607 | Use alpine-slim image instead of a default one

sirtawast:
## Description :sparkles:

Achieve a smaller docker fingerprint by switching to alpine-slim instead of the default distro.

## Screenshots :camera_flash:

Screenshot 2023-01-02 at 13 18 25


mjturt (approved):
Looks good! We should probably also do more comprehensive Dockerfile refactoring in the future and modify ticket [HL-160](https://helsinkisolutionoffice.atlassian.net/browse/HL-160) to meet our needs.

PR #1684, opened 2023-01-03
HL-606 | Fix CORS locally on media redirect cases

sirtawast:
## Description :sparkles:

* Fix CORS issue on HTTP => HTTPS redirect to :8000
* Fix wrong pathing to media file
* Add new media file `service_unavailable.pdf` for placeholder terms
* Only display action buttons if the terms file has been uploaded manually

## Screenshots :camera_flash:

Screenshot 2023-01-03 at 15 01 39


mjturt:
Looks good, but I would wait with this until the whole team is present. If the fixtures are run, placeholder PDF would override any already uploaded PDF's (might be issue e.g. in the production). Also not sure about the `isContinueAllowed` logic here.

Could we extract the CORS fix to separate PR from this? That I think we could merge :)

sirtawast:
Closed for now, see if @rikuke seeds were enough.

PR #1696, opened 2023-01-04
YJDH-655 | Fix YouthApplication factories' need_additional_info

karisal-anders:
### KS-Backend: Fix need_additional_info generating in factories 

Previously determine_target_group_social_security_number function
sometimes returned social security numbers for two different years.
When NEXT_PUBLIC_DISABLE_VTJ was set to True and
determine_target_group_social_security_number returned a social security
number with year difference of 17 from the current year rather than 16
it made YouthApplication.need_additional_info return True.

Now the determine_target_group_social_security_number only returns
social security numbers with a year difference of 16 from the current
year and AbstractYouthApplicationFactory based factories that expect a
certain need_additional_info value in their created object work with
both NEXT_PUBLIC_DISABLE_VTJ=False and NEXT_PUBLIC_DISABLE_VTJ=True.

Add test_need_additional_info test to test_factories.py to check that
need_additional_info is properly generated in factories.

Add test test_get_random_social_security_number_for_year to test that
get_random_social_security_number_for_year generates valid non-temporary
Finnish social security numbers for a given year.

Shared backend code:
 - Add create_finnish_social_security_number to create a Finnish social
   security number based on birthdate and individual number
 - Add tests for create_finnish_social_security_number
   - test_valid_create_finnish_social_security_number
   - test_valid_consecutive_create_finnish_social_security_number
   - test_invalid_create_finnish_social_security_number

Refs YJDH-655
 
### KS-Backend: Make use of get_faker function to speed up tests 

Not needing to instantiate Faker object but using the same Faker
object sped up locally running all tests using
"pytest . --pyargs shared -vv" from ~320s to ~150s.
Tried once on Windows 10 using Docker Desktop and WSL 2.

Refs YJDH-655 (Thought of the shared Faker object during this ticket)

### HL/KS/Shared-Backends: Move test utils to tests/utils.py 

Move these functions only used in tests to shared.common.tests.utils:
 - create_finnish_social_security_number
 - normalize_whitespace
 - set_setting_to_value_or_del_with_none
 - utc_datetime

KS-Backend:
 - Move get_random_social_security_number_for_year to common.tests.utils

Refs YJDH-655 (Refactoring)

## Issues :bug:

[YJDH-655](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-655)

[YJDH-655]: https://helsinkisolutionoffice.atlassian.net/browse/YJDH-655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/kesaseteli/common/tests/test_factories.py
+++ b/backend/kesaseteli/common/tests/test_factories.py
@@ -191,6 +191,38 @@ def test_youth_application_factory(  # noqa: C901
 
 
 @pytest.mark.django_db
+@override_settings(NEXT_PUBLIC_MOCK_FLAG=True)
+@pytest.mark.parametrize(
+    "youth_application_factory,expected_need_additional_info",
+    [
+        (AdditionalInfoRequestedYouthApplicationFactory, True),
+        (AdditionalInfoProvidedYouthApplicationFactory, True),
+        (InactiveNoNeedAdditionalInfoYouthApplicationFactory, False),
+        (InactiveNeedAdditionalInfoYouthApplicationFactory, True),
+    ],
+)
+@pytest.mark.parametrize("next_public_disable_vtj", [False, True])
+@pytest.mark.parametrize(
+    "freeze_date", ["2021-01-01", "2021-06-15", "2021-12-31", "2023-07-16"]
+)
+def test_need_additional_info(
+    settings,
+    make_youth_application_activation_link_unexpired,
+    youth_application_factory,
+    expected_need_additional_info: bool,
+    next_public_disable_vtj: bool,
+    freeze_date: str,
+):
+    settings.NEXT_PUBLIC_DISABLE_VTJ = next_public_disable_vtj
+    for _ in range(10):  # Run more than once because factories generate random data
+        with freeze_time(freeze_date):
+            youth_application: YouthApplication = youth_application_factory()
+            assert (
+                youth_application.need_additional_info == expected_need_additional_info
+            )
+
+
+@pytest.mark.django_db
 def test_youth_summer_voucher_factory():
     # Run more than once because factories generate random data
     youth_summer_vouchers = YouthSummerVoucherFactory.create_batch(size=20)
--- a/backend/kesaseteli/common/tests/test_utils.py
+++ b/backend/kesaseteli/common/tests/test_utils.py
@@ -1,11 +1,13 @@
 from datetime import datetime, timezone
 
 import pytest
+import stdnum.fi.hetu
 from django.core.exceptions import ValidationError
 
 from common.utils import (
     are_same_text_lists,
     are_same_texts,
+    get_random_social_security_number_for_year,
     has_whitespace,
     is_uppercase,
     normalize_for_string_comparison,
@@ -14,6 +16,7 @@ from common.utils import (
     validate_finnish_social_security_number,
     validate_optional_finnish_social_security_number,
 )
+from shared.common.utils import social_security_number_birthdate
 
 
 def get_empty_values():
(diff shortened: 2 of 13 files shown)


backend/shared/shared/common/utils.py
```diff
@@ -137,3 +137,33 @@ def social_security_number_birthdate(social_security_number) -> date:
     century_char = compacted_social_security_number[6]
     century = {"+": 1800, "-": 1900, "A": 2000}[century_char]
     return date(year=century + year_mod_100, month=month, day=day)
+
+
+def create_finnish_social_security_number(
```
nenonja, line 142: Perhaps this util function could be under tests folder if it's only used in tests?

backend/kesaseteli/common/utils.py
```diff
@@ -118,6 +122,23 @@ def validate_finnish_social_security_number(value):
         )
 
 
+def get_random_social_security_number_for_year(year: int) -> str:
```
nenonja, line 125: Perhaps this util function could be under tests folder if it's only used in tests?

backend/kesaseteli/common/tests/test_utils.py
```diff
@@ -155,3 +158,11 @@ def test_utc_datetime(args):
     result = utc_datetime(*args)
     assert result == datetime(*args, tzinfo=timezone.utc)
     assert result.tzinfo == timezone.utc
+
+
+@pytest.mark.parametrize("year", [1800, 2022, 2023, 2099])
+def test_get_random_social_security_number_for_year(year):
+    for _ in range(1000):
```
nenonja, line 165: another idea could be just loop all days of the year? Not sure how to do it in python but that would be more complete coverage (all possible days) and faster (365/366 items in loop)

nenonja (approved):
Good job! Some minor comments but also approved

backend/kesaseteli/common/tests/test_utils.py
nenonja: Well I read the test badly, it was about the testing _random_ ssn

backend/kesaseteli/common/tests/test_factories.py
```diff
@@ -190,6 +190,38 @@ def test_youth_application_factory(  # noqa: C901
...
+        (AdditionalInfoProvidedYouthApplicationFactory, True),
+        (InactiveNoNeedAdditionalInfoYouthApplicationFactory, False),
+        (InactiveNeedAdditionalInfoYouthApplicationFactory, True),
+    ],
+)
+@pytest.mark.parametrize("next_public_disable_vtj", [False, True])
+@pytest.mark.parametrize(
+    "freeze_date", ["2021-01-01", "2021-06-15", "2021-12-31", "2023-07-16"]
+)
+def test_need_additional_info(
```
nenonja, line 208: Perhaps we could also create a test `test_doesnt_need_additional_info` and use specific days that it should not apply and specific factories where it should not ask for additional info?

backend/shared/shared/common/utils.py
karisal-anders: Taken into account

backend/kesaseteli/common/utils.py
karisal-anders: Taken into account

backend/kesaseteli/common/tests/test_factories.py
karisal-anders: `test_need_additional_info` includes parametrization to test that `InactiveNoNeedAdditionalInfoYouthApplicationFactory` returns YouthApplication objects where their `need_additional_info` is False. `test_youth_application_factory` test above this test in the same file also tests `need_additional_info` a bit but not very extensively.

PR #1703, opened 2023-01-09
YJDH-656 | KS-Frontend: Always show titles in employer UI's selection groups

karisal-anders:
## Description :sparkles:

Affected selection groups in employer UI:
 - target_group
   - "Vuosiluokka tai syntymävuosi" in Finnish
 - hired_without_voucher_assessment
   - "Olisitko palkannut nuoren ilman kesäseteliä?" in Finnish

Previously the showTitle parameter to these selection groups was either false or not given at all and thus falsy. Thus their titles were not shown. Now the showTitle parameter is removed completely and the selection groups' titles are always shown.

## Issues :bug:

YJDH-656


nenonja (approved):
LGTM

PR #1705, opened 2023-01-10
HL-229 | Reorder applications by date at handler and applicant list view 

sirtawast:
* Reorder by date
  * ~Might want to refactor to use a GET "sortby", "orderby" param from client~
  * ☝️  Done
* Change label from 'created' to 'saved'

screencapture-localhost-3000-2023-01-10-09_15_48

### DIFF
--- a/backend/benefit/applications/api/v1/views.py
+++ b/backend/benefit/applications/api/v1/views.py
@@ -35,7 +35,6 @@ from messages.models import MessageType
 from shared.audit_log.viewsets import AuditLoggingModelViewSet
 from users.utils import get_company_from_request
 
-
 class BaseApplicationFilter(filters.FilterSet):
 
     status = filters.MultipleChoiceFilter(
@@ -154,7 +153,9 @@ class BaseApplicationViewSet(AuditLoggingModelViewSet):
         )
 
         serializer = self.serializer_class(qs, many=True, context=context)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+        ordered = sorted(serializer.data, key=lambda application:application["submitted_at"], reverse=True)
+
+        return Response(ordered, status=status.HTTP_200_OK)
 
     @action(
         methods=("POST",),
--- a/backend/benefit/applications/api/v1/views.py
+++ b/backend/benefit/applications/api/v1/views.py
@@ -35,6 +35,7 @@ from messages.models import MessageType
 from shared.audit_log.viewsets import AuditLoggingModelViewSet
 from users.utils import get_company_from_request
 
+
 class BaseApplicationFilter(filters.FilterSet):
 
     status = filters.MultipleChoiceFilter(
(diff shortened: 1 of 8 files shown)


backend/benefit/applications/api/v1/views.py
```diff
@@ -154,7 +154,23 @@ def simplified_application_list(self, request):
         )
 
         serializer = self.serializer_class(qs, many=True, context=context)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+
+        order_by = request.GET.get("order_by")
+        is_order_descending = True if order_by and order_by.lower() == "desc" else False
```
mjturt, line 159: You can use `bool(order_by and order_by.lower() == "desc")`.

backend/benefit/applications/api/v1/views.py
```diff
@@ -154,7 +154,23 @@ def simplified_application_list(self, request):
...
+
+        order_key = (
+            request.GET.get("order_key")
+            if request.GET.get("order_key")
+            else "created_at"
```
mjturt, line 164: Here, you can use `.get()` method's default parameter, for example `request.GET.get("order_key", "created_at")`.

backend/benefit/applications/api/v1/views.py
mjturt: Also, here usage of `.get()`'s default value could maybe simplify some things :)

mjturt (changes requested):
Nice work on the Python side too :grinning: Just thinking, was the option to change ordering by parameter in the spec? I would consider consulting PO's first that there's a possible future need for this before merging, as this adds (possibly unnecessary) complexity to the method. Less is more! :D

mjturt:
Test coverage _seems_ to be dropped quite a lot here too. It could also be nice to get ordering tested.

sirtawast:
> Nice work on the Python side too 😀 Just thinking, was the option to change ordering by parameter in the spec? I would consider consulting PO's first that there's a possible future need for this before merging, as this adds (possibly unnecessary) complexity to the method. Less is more! :D

@mjturt yeah it's a bit of an overkill. But take it as an exercise. I just want to re-cap some primitive python things so I can be of actual help and not just blindly stare at my display 😂 

Anyway, I think at some point someone would like to list the applications per company name or something, so then it's just a matter of implementing it in frontend.

backend/benefit/applications/api/v1/views.py
```diff
@@ -154,7 +154,19 @@ def simplified_application_list(self, request):
         )
 
         serializer = self.serializer_class(qs, many=True, context=context)
-        return Response(serializer.data, status=status.HTTP_200_OK)
+
+        order_key = request.GET.get("order_key", "created_at")
+
+        order_by = request.GET.get("order_by", "asc")
+        is_order_descending = order_by.lower() == "desc" or False
```
mjturt, line 161: `.lower()` might fail if `order_by` is not string, and that's probably why there's the `or False` at the end? You could define a serializer for the query parameters to handle validation etc. It would handle HTTP responses etc. automatically. [Example](https://tinystruggles.com/posts/drf_recipes/#use-a-serializer-for-the-query-parameters).

Same thing with the `order_key`, what happens if it's not a real key?

backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -2103,3 +2088,58 @@ def _expected_sort_key(obj):
     )
     returned_application_ids = [elem["id"] for elem in response.data]
     assert expected_application_ids == returned_application_ids
+
+
+def test_handler_application_ordering_by_key(handler_api_client):
```
mjturt, line 2093: :+1:

mjturt (commented):
Nice to see some tests too :+1:

rikuke (approved):

sirtawast:
Moved to https://github.com/City-of-Helsinki/yjdh/pull/1733 because of @mjturt current absence.

PR #1707, opened 2023-01-10
yjdh-657: fix endless loop on kesaseteli login page

nenonja:
To make sure error handler doesn't cause infinite loop, let's do the following:
- if we are already on login page and get auth error, do nothing (because on login page we are not authorized yet)
- if we are routing to the same path where we already are, use shallow route (to avoid unnecessary refresh)
- if we are routing to exact same path where we already are, do nothing (to avoid loop)

### DIFF
--- a/frontend/shared/src/error-handler/error-handler.ts
+++ b/frontend/shared/src/error-handler/error-handler.ts
@@ -6,6 +6,7 @@ import { isError } from 'shared/utils/type-guards';
 type Params = {
   error: Error | unknown;
   t: TFunction;
+  pathname: string;
   goToPage: GoToPageFunction;
   onAuthError?: () => void;
   onServerError?: () => void;
@@ -13,9 +14,13 @@ type Params = {
 };
 
 const handleError = (args: Params): void => {
-  const { error, t, goToPage } = args;
+  const { error, t, pathname, goToPage } = args;
   const {
-    onAuthError = () => goToPage('/login?sessionExpired=true'),
+    onAuthError = () => {
+      if (pathname !== '/login') {
+        goToPage('/login?sessionExpired=true');
+      }
+    },
     onServerError = () => goToPage('/500'),
     onCommonError = () =>
       showErrorToast(
--- a/frontend/shared/src/hooks/useGoToPage.ts
+++ b/frontend/shared/src/hooks/useGoToPage.ts
@@ -7,24 +7,19 @@ import GoToPageFunction from 'shared/types/go-to-page-function';
 const useGoToPage = (): GoToPageFunction => {
   const router = useRouter();
   const locale = useLocale();
-  const isRouting = useIsRouting();
-
+  const isAlreadyRouting = useIsRouting();
   return React.useCallback(
-    (pagePath, options): void => {
-      if (isRouting) {
-        // if already routing dont re-route;
+    (pagePath, operation = 'push'): void => {
+      // if already routing, or we're trying to route to same path (including query params), do nothing
+      if (isAlreadyRouting || pagePath === router.asPath) {
         return;
       }
       const path = `${locale}${pagePath || '/'}`;
-      if (options) {
-        const { operation, asPath, ...routerOptions } = options;
-        const op = operation ?? 'push';
-        void router[op](path, asPath ?? path, routerOptions);
-      } else {
-        void router.push(path);
-      }
+      // use shallow route when we are already on the same path (ignoring query params)
+      const shallow = pagePath === router.pathname;
+      void router[operation](path, undefined, { shallow });
     },
-    [locale, router, isRouting]
+    [locale, router, isAlreadyRouting]
   );
 };
 
--- a/frontend/kesaseteli/employer/src/__tests__/application.test.tsx
+++ b/frontend/kesaseteli/employer/src/__tests__/application.test.tsx
@@ -36,7 +36,11 @@ describe('frontend/kesaseteli/employer/src/pages/application.tsx', () => {
       const spyPush = jest.fn();
       renderPage(ApplicationPage, { push: spyPush });
       await waitFor(() =>
-        expect(spyPush).toHaveBeenCalledWith(`${DEFAULT_LANGUAGE}/login`)
+        expect(spyPush).toHaveBeenCalledWith(
+          `${DEFAULT_LANGUAGE}/login?sessionExpired=true`,
+          undefined,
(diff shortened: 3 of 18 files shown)


frontend/shared/src/hooks/useGoToPage.ts
```diff
@@ -8,11 +8,10 @@ const useGoToPage = (): GoToPageFunction => {
   const router = useRouter();
   const locale = useLocale();
   const isRouting = useIsRouting();
-
   return React.useCallback(
     (pagePath, options): void => {
-      if (isRouting) {
-        // if already routing dont re-route;
+      // if already routing, or we're trying to route to same path, do nothing;
+      if (isRouting || pagePath === router.pathname) {
```
karisal-anders, line 14: What about the case if the query parameters aren't the same?

frontend/shared/src/error-handler/error-handler.ts
```diff
@@ -6,16 +6,21 @@ import { isError } from 'shared/utils/type-guards';
...
   onCommonError?: () => void;
 };
 
 const handleError = (args: Params): void => {
-  const { error, t, goToPage } = args;
+  const { error, t, pathname, goToPage } = args;
   const {
-    onAuthError = () => goToPage('/login?sessionExpired=true'),
+    onAuthError = () => {
+      if (pathname !== '/login') {
```
karisal-anders, line 20: Is it so that the `pathname` here does not include query parameters so even if we're already on page `/login?sessionExpired=true` the pathname will be `/login`?

frontend/shared/src/hooks/useGoToPage.ts
nenonja: fixed now. Let's us `router.asPath` which checks the query param as well

nenonja: Screenshot 2023-01-11 at 3 42 06 PM

frontend/shared/src/hooks/useGoToPage.ts
```diff
@@ -21,7 +20,9 @@ const useGoToPage = (): GoToPageFunction => {
         const op = operation ?? 'push';
         void router[op](path, asPath ?? path, routerOptions);
       } else {
-        void router.push(path);
+        // use shallow route when we are already on the same page
+        const shallow = pagePath.startsWith(router.pathname);
```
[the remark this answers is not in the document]
nenonja: and here we use `pagePath === router.pathname` which ignores query params but takes subpaths into account

frontend/shared/src/error-handler/error-handler.ts
nenonja: It doesn't

frontend/shared/src/hooks/useGoToPage.ts
```diff
@@ -21,10 +20,12 @@ const useGoToPage = (): GoToPageFunction => {
         const op = operation ?? 'push';
         void router[op](path, asPath ?? path, routerOptions);
       } else {
-        void router.push(path);
+        // use shallow route when we are already on the same path (ignoring query params)
+        const shallow = pagePath === router.pathname;
+        void router.push(path, undefined, { shallow });
```
karisal-anders, line 25: If there are no options given, router is not yet routing and pagePath has query parameters but they have changed then what happens? Doesn't it first not go into the `if (isAlreadyRouting || pagePath === router.asPath)` because the router is not yet routing and the query parameters have changed, then it doesn't go into the `if (options)` clause because there are no options but into its `else` clause? And then in that part it will use shallow routing because the base pathname is the same? So doesn't this use shallow routing for an URL whose query parameters have changed in this case? Or did I miss or misinterpret something?

frontend/shared/src/error-handler/error-handler.ts
```diff
@@ -6,16 +6,21 @@ import { isError } from 'shared/utils/type-guards';
...
+        goToPage('/login?sessionExpired=true');
```
karisal-anders, line 21: Now it goes to the `/login?sessionExpired=true` page when opening the employer UI by going to e.g. https://localhost:3000, right? I mean wouldn't it work the same in production also if employers would open the https://kesaseteli.hel.fi/ page and they weren't logged in yet they would get the "Käyttäjäsessio vanhentui. Kirjaudu uudelleen sisään" error notification rather than the "Palvelun käyttäminen edellyttää vahvaa tunnistautumista" info notification? Just thinking whether this is ok.

nenonja: I think that the backend will return 401 on both cases, so it's impossible to make the difference on ui-side. So it will show the `Käyttäjäsessio vanhentui. Kirjaudu uudelleen sisään` error on both cases, when the session exists but expired, or session didnt exist in the first place.

karisal-anders: Alright, this is another problem so this doesn't have to be tackled in this PR, IMHO we can leave it be like this now.

frontend/shared/src/error-handler/error-handler.ts
karisal-anders: Ok, good

frontend/shared/src/hooks/useGoToPage.ts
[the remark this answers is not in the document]
karisal-anders: This is ok

frontend/shared/src/hooks/useGoToPage.ts
karisal-anders: That's good

frontend/shared/src/hooks/useGoToPage.ts
karisal-anders: The `options` were used e.g. in ApplicationPage:

```typescript
  if (applicationQuery.isSuccess) {
    if (applicationQuery.data.status !== 'draft' && applicationId) {
      void goToPage(`/thankyou?id=${applicationId}`, { operation: 'replace' });
    }
```

After discussion using the shallow routing for the otherwise same URL but with only the query parameters changed is ok, I wasn't sure whether the page contents would get updated correctly in that situation with shallow routing but they should so that's ok.

nenonja: Fixed now. Options were reduced to only allow the operation to be changed which can be  `push` or `replace`. `push` is default when operation type is not given.

karisal-anders (approved):
LGTM!

PR #1719, opened 2023-01-13
Hl 626 command line tool for seeding applications

rikuke:
This  PR adds a `seed` command that can run to seed the database with some application data, to help with frontend development and manual testing. The command by default creates 10 applications for each of the seven different application statuses, the amount can be increased with the `--number` flag. Running the command also adds one attachment for each application and creates an attachment .pdf-file, based on the `valid_pdf_file.pdf`-file located in the `applications/tests` directory. Worth noting is that running  the seed command clears any existing application data from the database and clears the local `var/media`-folder to prevent dummy .pdf files accumulating.

The feature can be tested with pytest by running `pytest applications/tests/test_application_tasks.py ` and/or by running the  command with `python manage.py seed --number=20` and confirming from the database that 140 applications have been created. The seeded data can also  be tested from the application frontend at `localhost:3000` by clicking the listed applications and confirming that the attachment links work.

This feature can and will be further developed as more diverse test data is required.

### DIFF
--- a/backend/benefit/applications/management/commands/seed.py
+++ b/backend/benefit/applications/management/commands/seed.py
@@ -9,7 +9,7 @@ from applications.enums import ApplicationStatus
 from applications.models import Application, ApplicationBasis
 from applications.tests.factories import (
     AdditionalInformationNeededApplicationFactory,
-    ApplicationFactory,
+    ApplicationWithAttachmentFactory,
     CancelledApplicationFactory,
     DecidedApplicationFactory,
     HandlingApplicationFactory,
@@ -53,7 +53,7 @@ def run_seed(number):
 
     factories = (
         AdditionalInformationNeededApplicationFactory,
-        ApplicationFactory,
+        ApplicationWithAttachmentFactory,
         CancelledApplicationFactory,
         DecidedApplicationFactory,
         HandlingApplicationFactory,
--- a/backend/benefit/applications/management/commands/seed.py
+++ b/backend/benefit/applications/management/commands/seed.py
@@ -45,6 +45,6 @@ def run_seed(number):
     """Delete all existing applications and create applications for all statuses"""
     clear_applications()
 
-    for i in range(number):
+    for _ in range(number):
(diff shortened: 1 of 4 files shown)


mjturt (approved):
Looks good to me! :)

backend/benefit/applications/management/commands/seed.py
```diff
@@ -0,0 +1,50 @@
...
+            "--number",
+            type=int,
+            default=10,
+            help="Number of applications to create",
+        )
+
+    def handle(self, *args, **options):
+        total_created = len(ApplicationStatus.values) * options["number"]
+        if os.getenv("DEBUG") != "1":
+            self.stdout.write("Seeding is only allowed in debug mode")
```
mjturt, line 24: Actually, just out of curiosity, why not just use `print()`? I believe that `stdout.write()` doesn't put the newline at the end of the line (`print()` has an `end` parameter if that's the reason)?

sirtawast (changes requested):
The `submitted_at` seems to be missing from most of the applications. That needs to be set. I got it showing by adding `bf_applications_applicationlogentry` for each application.

backend/benefit/applications/management/commands/seed.py
rikuke: No other reason for using stdout.write besides that Django documentation uses stdout.write in their examples for creating management commands. The only practical difference is indeed the newline after calling print(). I guess command line tools could at some scenarios have progress indicators like "#########" that need to be printed on the same line.

rikuke:
Alright, I'll add some `submitted_at` information to the applications.

sirtawast:
Wonder why the tests fail -- looks to be pretty weird python errors.

sirtawast:
@rikuke ok this is random but 

https://github.com/City-of-Helsinki/yjdh#known-errors

will fix the error.  Good thing it crossed my mind, was wondering if this **EVER** happened in real life 😂

PR #1755, opened 2023-01-27
Seeder improvements

rikuke:
## Description :sparkles:
Improve the previously added seeder command so that it works better in the development environment where 
attachments are uploaded into Azure blob storage. 
- Delete attachments through the FileModelField class
- Check the DEBUG env-variable with django.conf
- Add a test for the DEBUG variable
- Modify seeder so that cancelled applications have  `modified_at` date older than 30 days.

## Testing :alembic:
Run `pytest applications/tests/test_application_tasks.py`.

### DIFF
--- a/backend/benefit/applications/management/commands/seed.py
+++ b/backend/benefit/applications/management/commands/seed.py
@@ -1,6 +1,7 @@
 from datetime import timedelta
 
 import faker
+import pytz
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


sirtawast (approved):

backend/benefit/applications/management/commands/seed.py
```diff
@@ -31,23 +33,32 @@ def add_arguments(self, parser):
...
     which are not deleted with on_delete=CASCADE and delete the media folder"""
-    Application.objects.all().delete()
+    applications = Application.objects.all()
+    for application in applications:
+        for attachment in application.attachments.all():
+            attachment.attachment_file.delete()
+    applications.delete()
+
     ApplicationBasis.objects.all().delete()
-    shutil.rmtree(settings.MEDIA_ROOT)
```
sirtawast, line 46: I assume this fixes a bug wherer `python manage.py seed --number=0` is used to clear all applications. Good stuff.

sirtawast (approved):

PR #1804, opened 2023-02-03
Hl 552 sentry config

rikuke:
## Description :sparkles:
According to the Sentry [documentation](https://docs.sentry.io/platforms/javascript/guides/nextjs/sourcemaps/) , Sentry NextJS integration uses the Sentry CLI under the hood to tag a release and generate & send sourcemaps when `Sentry.init()` is run.  In the current build stage, the `NEXTJS_DISABLE_SENTRY`  variable is always true, possibly due to the variable being set 1 in the shared frontend Dockerfile, so the `Sentry.init() `is not run. This PR also adds the env variables required by the Sentry CLI.

## Testing :alembic:
Start a new build with these changes to the Dockerfile and observe if Sentry is initialised when `next build` command is run.


mjturt (approved):
:+1:

PR #1829, opened 2023-02-08
Hl-624 SSN validator from python-stdnum

mjturt:
Changed SSN validator and fixed related tests. Ran black on files, so it looks that there are plenty of changes. I will leave comment on relevant changes.


backend/benefit/applications/api/v1/serializers.py
```diff
@@ -307,25 +282,12 @@ class Meta:
...
-        """
-        For more info about the checksum validation, see "Miten henkilötunnukset tarkistusmerkki lasketaan?" in
-        https://dvv.fi/henkilotunnus
-        """
         if value == "":
             return value
 
-        m = re.match(self.SSN_REGEX, value)
-        if not m:
+        if not hetu.is_valid(value):
```
mjturt, line 288: Use new SSN validator.

mjturt (commented):
Author's comments.

backend/benefit/applications/tests/test_applications_api.py
```diff
@@ -576,9 +576,10 @@ def test_application_put_edit_employee(api_client, application):
     """
     modify existing application
     """
+    new_ssn = "260778-323Y"
```
mjturt, line 579: And here let's use non-temporary SSN :)

sirtawast: Did you run any tests with 2023 SSN-style as per https://dvv.fi/hetu-uudistus?

sirtawast: I guess they are already tested thoroughly in the lib though 🤔

mjturt: Actually, it looks like python-stdnum doesn't support the new middle characters. Python-stdnum is well maintained, so should we make a pull request there to add the support, or rather just fix our own implementation and put it under shared code, what do you think?

I think there's no rush, we can wait python-stdnum to be updated:

_Tilanne 1/2023: Tämänhetkisen arvion mukaisesti uuden välimerkkiasetuksen mukaisia uusia henkilötunnuksen välimerkkejä annetaan ensimmäisen kerran maaliskuussa 2024._

rikuke (approved):

PR #1832, opened 2023-02-09
fix: change youth summer voucher email's year from 2022 to 2023

karisal-anders:
### fix: change youth summer voucher email's year from 2022 to 2023

refs None

## Issues :bug:

`-`

### DIFF
--- a/backend/kesaseteli/applications/locale/en/LC_MESSAGES/django.po
+++ b/backend/kesaseteli/applications/locale/en/LC_MESSAGES/django.po
@@ -700,7 +700,7 @@ msgstr ""
 
 msgid ""
 "Kaupunki korvaa 325 euroa työnantajalle, joka palkkaa Kesäseteliin "
-"oikeutetun nuoren töihin 1.6.–15.8.2022 välisenä aikana."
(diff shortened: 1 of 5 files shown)


backend/kesaseteli/applications/locale/en/LC_MESSAGES/django.po
```diff
@@ -700,7 +700,7 @@ msgstr ""
 
 msgid ""
 "Kaupunki korvaa 325 euroa työnantajalle, joka palkkaa Kesäseteliin "
-"oikeutetun nuoren töihin 1.6.–15.8.2022 välisenä aikana."
+"oikeutetun nuoren töihin 1.6.–15.8.2023 välisenä aikana."
```
nenonja, line 703: I wonder if the year could be read the year from server's clock so it doesnt have to be changed every year?

nenonja (commented):
I wonder if the year could be read the year from server's clock so it doesnt have to be changed every year?

nenonja (approved):
Agreed that email template should be rechecked everything year. Perhaps this could be improved in the future so that the text could be changed from django admin.

PR #1850, opened 2023-02-16
HL-567 | Fix ADFS in local environment, add log in/out to UI

sirtawast:
* Header now displays user's full name and has a dropdown link to sign out
* Changes Django Auth callback links to work on localhost
* If HTTPS=on causes any harm on remote env's, remove this line or use an env
* HTTPS should be generally be always on though

## Test 

In `.env.benefit-backend`

* `ADFS_LOGIN_REDIRECT_URL=https://localhost:3100`
* `NEXT_PUBLIC_MOCK_FLAG=0`

Go to https://localhost:3100 and press log in. Use test user credentials. You should see log out button in right hand corner.


rikuke (approved):
Looks good!

PR #1852, opened 2023-02-17
HL-643 Python tooling

mjturt:
So, the problem with Python tools was not that configuration files weren't used. The problem was that `isort` didn't recognize local modules in some cases/environments, resulting in wrong import order. Running hooks per project should fix that.


README.md
```diff
@@ -141,14 +141,23 @@ Running `yarn tet-youth down` brings down all services.
...
 1. Run `yarn install` and `yarn prepare` on project root
 2. Try `git commit -m foo`. It does not commit anything for real but pre-commit hook should be triggered.
 
 If the pre-commit hook hangs, or you want to push changes with failing tests, you can disable the hook with
 `npx husky uninstall`. Running `yarn prepare` reactivates the hook.
 
+### Pre-commit (backend)
+
+1. Install [pre-commit](https://pre-commit.com/)
+2. Run `pre-commit install`
```
rikuke, line 157: Maybe we could install this with the `requirements-dev.in`-file ?

mjturt: Yeah, true :+1:

.pre-commit-config.yaml
```diff
@@ -8,18 +8,64 @@ repos:
...
+      - id: black
+        name: (TET) black
+        args: [--preview]
+        files: ^backend/tet
+      - id: black
+        name: (Shared) black
+        args: [--preview]
+        files: ^backend/shared
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
```
karisal-anders, line 27: Benefit/Kesaseteli/TET backends use an older isort 5.10.1 at the moment according to their requirements-dev.txt files. Maybe black/isort/flake8 versions could be updated to their newest versions in this file and in all the backends' requirements-dev.txt files to make them be in sync. At the least the versions should be the same for all IMHO.

rikuke (approved):

.pre-commit-config.yaml
mjturt: Versions should be in sync now. Also added the pre-commit dependency to kesaseteli/tet also.

PR #1853, opened 2023-02-17
Hl 655 add s3

rikuke:
## Description :sparkles:
This PR add S3 file system support in anticipation of the upcoming staging / production environments.
As the current dev/test environments use Azure Blob storage, the S3 support is optionally enable by setting the 
`USE_S3` env-variable  to True and then adding  the  S3-specific connection variables:

- `S3_ENDPOINT_URL`
- `S3_ACCESS_KEY_ID`
- `S3_SECRET_ACCESS_KEY`
- `S3_STORAGE_BUCKET_NAME`

The PR also adds a [MinIO](https://min.io/) -container that can be used to test the S3-storage in the local development environment.

## Issues :bug:
MinIO cannot keep up with the amount of mock files generated by the test factories so running the `python manage.py seed `and `pytest` commands takes a looong time or fails.  So before running backend tests or the seed command, switch back to the local disk as a default storage by setting `USE_S3=0`. This is clumsy, and a possible workaround could be to set the `USE_S3 `to False when running tests like this ` USE_S3=False pytest . `.

## Testing :alembic:

0. Enable the S3 storage by settting the necessary variables from `.env.benefit-backend.example`.
1. Build the containers
2. Visit the minIO console at the `http://localhost:9090`
3. Login with username `minio-root` and password `minio-pass`
4. Navigate to Object Browser and select `local-s3-bucket`
5. From the applicant UI, upload some attachments
6. From the minIO console, click refresh and verify that the files uploaded from the UI are present

To confirm that a file is stored from the backend:
1.  Run` python manage.py shell`
2. `from django.core.files.storage import default_storage`
3. `file = default_storage.open('testfile', 'w')`
4. `file.write('test text')`
5. `file.close()`
6.  Refresh the bucket in the minIO console and verify that testfile was uploaded.

## Screenshots :camera_flash:
Screenshot 2023-02-20 at 9 46 32
Screenshot 2023-02-20 at 10 21 23

Screenshot 2023-02-20 at 10 21 35


mjturt (commented):
Changes itself looks good, just thinking that should the tests/factories be fixed first if they are doing something sketchy. I remember seeing thousands of mock PDF files somewhere, which is probably not wanted behaviour. Something is definitely wrong if minio hangs.

mjturt:
We can merge this first also, but then fixing sketchy tests should be considered as priority.

mjturt (approved):

rikuke:
Yeah PR HL-627 reduces the number of pdf-files generated somewhat, but not completely and does not solve this.  The way that  the model factories are used to generate pdf files is bit complex and should definitely be refactored, there's actually a case to be made that tests should rarely or never generate any files at all because any file operations in tests should should just mocked.  But yeah lets merge this for now and open another ticker for making the factories rely on file I/O less.

PR #1860, opened 2023-02-22
HL-677: Failing MinIO tests

mjturt:
As explained in the new readme, the tests freeze time and that's why MinIO requests fail.

Some discussion related to the issue [here](https://github.com/spulec/freezegun/issues/331). The [PR to freezegun](https://github.com/spulec/freezegun/pull/430) that would allow S3 related utilities to be ignored from time freeze (and probably fix this) is still open. 

## Possible solutions

1. Leave it as it is for now. Run tests minio disabled. I think this is ok, at least until we know more about our production environment, but no longer.
2. Remove time freezing from tests and let tests run in current time. Probably requires great effort.
3. Replace freezegun with a library that can ignore S3 utils properly. Freezegun has an "ignore library" configuration option, it's just not working for botocore etc. One possibility mentioned in freezegun's PR discussion would be [time-machine](https://github.com/adamchainz/time-machine). Most likely less effort than option 2.
4. Clock faking isn't a long-term solution.


rikuke (approved):

rikuke:
Yeah because of the renewed uncertainty about the production environment, we can let this be at it is for the time being and just set the `USE_S3=False` to use the local disk as before. When we know what type of storage we'll be using in prod, we can revisit this and either remove Minio and Boto etc or fix this issue.

PR #1873, opened 2023-02-23
HL-532 | Fix bunch of accessability issues with attachments

sirtawast:
## Description :sparkles:

* Change order of headings to incremental

* Change attachment list so that it's wrapped in `<ul>` and each item is a `<li>`

* Add `<title>` to every page (was previously undefined)

### DIFF
--- a/frontend/benefit/applicant/src/styles/globals.css
+++ b/frontend/benefit/applicant/src/styles/globals.css
@@ -1,0 +1,19 @@
+html,
+body {
+  padding: 0;
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+ul {
+  list-style: none;
+}
+
+ul, li {
+  margin: 0;
+  padding: 0;
+}
+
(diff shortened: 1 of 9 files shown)


frontend/benefit/applicant/src/styles/globals.css
```diff
@@ -0,0 +1,18 @@
...
+}
+
+ul {
+  list-style: none;
+}
+
+ul, li {
+  margin: 0;
+  padding: 0;
+}
```
mjturt, line 18: No newline at the end of the file. We have `.editorconfig` that specifies UNIX style lines to be used (meaning that every line should end with a newline character, even the last one).  Did you use VS Code? You could install some editorconfig plugin, for example https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig.

mjturt (changes requested):
The code itself looks good to me, just a small file format thingy.

frontend/benefit/applicant/src/styles/globals.css
sirtawast: Weird, I though this was linter stuff, usually is.

mjturt (approved):

PR #1891, opened 2023-02-28
HL-525 | Add a11y attributes to custom Stepper component

sirtawast:
## Description

Turns out the shared component [Stepper is a copy](https://hds.hel.fi/components/stepper/) from the one in design system.

* Make stepper look like HDS Stepper in 2023
* Add aria-label and aria-step to tell screen reader about current progress
* Remove inline css

### DIFF
--- a/frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
+++ b/frontend/benefit/applicant/src/components/applications/pageContent/PageContent.tsx
@@ -135,7 +135,7 @@ const PageContent: React.FC = () => {
           </$PageHeading>
         </$HeaderItem>
         <$HeaderItem>
-          <Stepper steps={steps} activeStep={currentStep} />
+          <Stepper steps={steps} activeStep={currentStep} as="ol" />
         </$HeaderItem>
       </$PageHeader>
       {id && application?.createdAt && !isSubmittedApplication && (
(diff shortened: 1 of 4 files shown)


mjturt:
Is there a reason why the HDS component was not used (or is not possible to use now)?

nenonja (changes requested):
Please note that the kesaseteli employer service also uses wizard+ stepper. For some reason the kesaseteli service arent deployed, the pipeline should trigger all the frontend services when frontend/shared folder is changed. Now I cannot try does the wizard still work on kesaseteli side or not... perhaps pipeline could be fixed so that the change would trigger all the deploys?

PS: Some tests fail on kesaseteli.

nenonja:
> Is there a reason why the HDS component was not used (or is not possible to use now)?

It's a copy because during the time it was implemented, it didnt exist on HDS. I think HDS even used ours as a base of their component? Anyway, HDS component could be used now if possible. There might be some futher customization though that doesnt work on HDS component, not sure? I think that the wizard steps are clickable, that might be one of the custom features.

sirtawast:
Tests pass now. 

I'll see if the HDS one is applicable as we don't have the priviledge to jump between the steps (I think).

nenonja (changes requested):
Stepper is visually broken on kesaseteli employer:
https://ks-empl-1891.test.kuva.hel.ninja/fi/application?id=071ba6ba-cdd9-43c1-b671-f18ae8615da5
Screenshot 2023-03-01 at 10 20 57 AM

sirtawast:
Closed, use HDS in the future. https://github.com/City-of-Helsinki/yjdh/pull/1894

PR #1892, opened 2023-03-01
Hl-672: Mock authenctications

mjturt:
## Description :sparkles:

Mock authentication when `NEXT_PUBLIC_MOCK_FLAG` is on also on the handler side. Remove mocking from user info endpoint, as now there should always be a real user with request objects when logged in.

Based on [python upgrade](https://github.com/City-of-Helsinki/yjdh/pull/1878) branch, so that should be merged first


backend/shared/shared/azure_adfs/mock_views.py
```diff
@@ -0,0 +1,37 @@
...
+
+class MockOAuth2CallbackView(View):
+    def get(self, request):
+        if not request.user.is_authenticated:
+            user = UserFactory(
+                username=f"handler_{factory.Faker('user_name')}",
+                is_staff=True,
+            )
+            auth.login(
+                request, user, backend="django.contrib.auth.backends.ModelBackend"  # type: ignore
```
karisal-anders, line 28: Why is the `# type: ignore` needed here? Couldn't this be done without the type ignore?

backend/shared/shared/azure_adfs/tests/test_adfs_login.py
```diff
@@ -27,7 +27,9 @@ def test_get_member_objects(requests_mock):
     auth_backend = HelsinkiAdfsAuthCodeBackend()
 
     member_objects_response = {
-        "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(Edm.String)",
+        "@odata.context": (
+            "https://graph.microsoft.com/v1.0/$metadata#Collection(Edm.String)"
+        ),
```
karisal-anders, line 32: Why all the formatting changes in this file i.e. putting strings into parentheses? Is black doing this or something else?

backend/shared/shared/azure_adfs/urls.py
```diff
@@ -1,8 +1,39 @@
...
+        name="login",
+    ),
+    re_path(
+        r"^logout$",
+        MockEnabledProxyView(
+            real_view_class=OAuth2LogoutView, mock_view_class=MockOAuth2LogoutView
+        ),
+        name="logout",
+    ),
     path("", include("django_auth_adfs.urls")),
```
karisal-anders, line 38: Where did the login and logout endpoints come from before this change? Did they come from `path("", include("django_auth_adfs.urls"))`? If they did then are you sure the login and logout mocks are not being overridden by values from `path("", include("django_auth_adfs.urls"))`?

backend/benefit/users/api/v1/views.py
```diff
@@ -15,7 +13,6 @@
 from users.api.v1.authentications import HelsinkiProfileApiTokenAuthentication
 from users.api.v1.permissions import BFGDPRScopesPermission
 from users.api.v1.serializers import UserSerializer
-from users.utils import set_mock_user_name
```
karisal-anders, line 18: It seems `set_mock_user_name` is dead code after this change so it could be removed IMHO as clean up.

backend/shared/shared/azure_adfs/urls.py
```diff
@@ -1,8 +1,39 @@
...
+        ),
+        name="login",
+    ),
+    re_path(
+        r"^logout$",
+        MockEnabledProxyView(
+            real_view_class=OAuth2LogoutView, mock_view_class=MockOAuth2LogoutView
+        ),
+        name="logout",
+    ),
```
karisal-anders, line 37: Writing tests for these mocked versions would be good for maintainability, like e.g. in backend/shared/shared/oidc/tests/test_mock_views.py but for these endpoints.

backend/shared/shared/azure_adfs/mock_views.py
mjturt: This is due to the factory-boy's poor typing support. `user` is `UserFactory` type and `auth.login()` requires basic user object.

backend/shared/shared/azure_adfs/tests/test_adfs_login.py
mjturt: Yes, new black version formats these also. The line was over 88 lines and in `dict` you need to put value in parentheses if it's on the different line that key.

backend/shared/shared/azure_adfs/urls.py
mjturt: Yes, they came from there `django_auth_adfs`. URLs are not overridden, as this works on the "first that matches" principle.

backend/benefit/users/api/v1/views.py
mjturt: Ok, then I will remove it.

karisal-anders (approved):
Good to further the mocking also to the ADFS part! 👍  I would clean up the commits before merging but otherwise looks ok to merge!

PR #1893, opened 2023-03-01
HL-685 | Never send a note via email to an applicant

sirtawast:
## Description :sparkles:

Found a issue in a handler session today. Should only send email when appropriate.

### DIFF
--- a/backend/benefit/messages/views.py
+++ b/backend/benefit/messages/views.py
@@ -7,7 +7,7 @@ from rest_framework.exceptions import NotFound
 from applications.models import Application
 from common.permissions import BFIsApplicant, BFIsHandler, TermsOfServiceAccepted
 from messages.automatic_messages import send_email_to_applicant
-from messages.models import Message
+from messages.models import Message, MessageType
 from messages.permissions import HasMessagePermission
 from messages.serializers import MessageSerializer, NoteSerializer
 from shared.audit_log.viewsets import AuditLoggingModelViewSet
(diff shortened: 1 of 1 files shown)


backend/benefit/messages/views.py
```diff
@@ -49,7 +49,9 @@ class HandlerMessageViewSet(ApplicantMessageViewSet):
 
     def perform_create(self, serializer):
         message = serializer.save()
-        send_email_to_applicant(message.application)
+        # Never send email if it's a handler's note!
+        if message.message_type != MessageType.Note:
```
sirtawast, line 53: @mjturt or should this be inclusive  rather than exclusive?

backend/benefit/messages/views.py
```diff
@@ -49,7 +48,12 @@ class HandlerMessageViewSet(ApplicantMessageViewSet):
 
     def perform_create(self, serializer):
         message = serializer.save()
-        send_email_to_applicant(message.application)
+        # Never send email if it's a handler's note!
+        if (
+            message.message_type == MessageType.HANDLER_MESSAGE
```
mjturt, line 53: Not sure about the logic here. If `MessageType.APPLICANT_MESSAGE` evaluates as true, this comparison always evaluates as true, despite the contents of `message`'s `message_type` attribute.

I believe that the intention here was to compare `message_type` to both `MessageType.HANDLER_MESSAGE` and `MessageType.APPLICANT_MESSAGE` which would be in python:
```python
if message.message_type == MessageType.HANDLER_MESSAGE or message.message_type == MessageType.APPLICANT_MESSAGE:
```
This is a bit too verbose, so it could be shorthanded using `list`'s `in` operator:
```python
if message.message_type in [MessageType.HANDLER_MESSAGE, MessageType.APPLICANT_MESSAGE]:
```

mjturt (changes requested):

rikuke (approved):

PR #1911, opened 2023-03-08
Add http-cache-semantics to audit-ci whitelist

sirtawast:
## Description :sparkles:

Audit-ci is failing constantly. Add `http-cache-semantics` package to whitelist.

image

### DIFF
--- a/frontend/audit-ci.jsonc
+++ b/frontend/audit-ci.jsonc
@@ -8,6 +8,7 @@
     "loader-utils",         // next
     "moment",               // testcafe
     "jsonwebtoken",         // testcafe 
+    "http-cache-semantics", // testcafe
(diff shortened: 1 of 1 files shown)


nenonja (approved):
I was thinking that if these are coming from testcafe or any test package, audit should not complain about them in the first place. Only production packages should be scanned because only them are included in the production build. Perhaps that's something to improve with audit-ci? But this pr looks good anyway!

sirtawast:
> I was thinking that if these are coming from testcafe or any test package, audit should not complain about them in the first place. Only production packages should be scanned because only them are included in the production build. Perhaps that's something to improve with audit-ci? But this pr looks good anyway!

Good call! Actually, lerna also depends on this 😄

PR #1913, opened 2023-03-10
HL-675 | Fix and rename AHJO batch file exports

sirtawast:
## Description :sparkles:

This turned out to be quiet the Django excursion for me.

* Backend: Freeze public and secret attachment names to 1 and 2, respectively
* Backend: Rename attachments from 3..n
* Backend: Split de minimis aided and regularly paid applications to separate PDF's
* Frontend: Change the exported batch zip's filename from report.zip to `hl 1970-01-01 00.00.00.zip` format to avoid confusion and so that they can be listed chronologically in the file explorer


frontend/benefit/handler/src/components/applicationReview/deminimisView/DeminimisView.tsx
```diff
@@ -79,8 +78,7 @@ const DeminimisView: React.FC<ApplicationReviewViewProps> = ({ data }) => {
       ) : (
         <$GridCell $colSpan={12}>
           <$ViewField>
-            {t(`${translationsBase}.fields.deMinimisAidsNo`)}{' '}
-            <$ViewFieldBold>{t('common:utility.no')}</$ViewFieldBold>
+            {t(`${translationsBase}.fields.deMinimisAidsNo`)}
```
sirtawast, line 81: No means no. No need to add "No" to no.

backend/benefit/applications/services/ahjo_integration.py
```diff
@@ -185,36 +236,52 @@ def generate_pdf(
 
 
 def generate_single_declined_file(
-    company: Company, apps: List[Application]
+    company: Company, apps: List[Application], attachment_number="n"
```
mjturt, line 239: Same here.

mjturt (changes requested):
Python starting to work well for you :+1: Just a couple of minor things.

mjturt (approved):
OK when the tests pass :+1:

PR #1914, opened 2023-03-13
HL-678 | New looks for handler front page view

sirtawast:
## Description :sparkles:

Add a tiny bit of contrast with background color on table views. There was a mechanism already in place to change background color per context but wires were not fully connected.

## Before

screencapture-localhost-3100-2023-03-13-16_18_12

screencapture-localhost-3100-2023-03-13-16_19_13

## After

screencapture-localhost-3100-2023-03-13-16_16_06

screencapture-localhost-3100-2023-03-13-16_20_05

### DIFF
--- a/frontend/benefit/handler/public/locales/en/common.json
+++ b/frontend/benefit/handler/public/locales/en/common.json
@@ -129,7 +129,11 @@
         "unreadMessagesCount": ""
       },
       "messages": {
-        "empty": "Ei eriä tällä hetkellä"
+        "empty": {
+          "received": "Ei yhtään saapunutta hakemusta",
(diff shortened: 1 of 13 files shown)


backend/benefit/messages/views.py
```diff
@@ -48,7 +48,12 @@ class HandlerMessageViewSet(ApplicantMessageViewSet):
 
     def perform_create(self, serializer):
         message = serializer.save()
-        send_email_to_applicant(message.application)
+        # Never send email if it's a handler's note!
```
rikuke, line 51: Wasn't this done in PR#1909 already?

mjturt (approved):
Looks good.

PR #1942, opened 2023-03-28
Downgrade elastic search client to 7.x  for audit logs to be uploaded

rikuke:
## Description :sparkles:
Audit logs are uploaded to Elastic Search server, but the current version of the Python client, 8.x ,is not compatible with the ES version on the server, 7.x, so the client needs to be downgraded to 7.x.


mjturt (changes requested):
Should the `requirements.txt` file be changed also?

mjturt (approved):

sirtawast (approved):

PR #1944, opened 2023-03-29
Hl 705 encrypt name

rikuke:
## Description :sparkles:
This PR encrypts the first and last name of the employee in the database the same way that the SSN is encrypted.
Also adds a test to confirm that a handler can still search/filter applications by employee first and last name.
## Testing :alembic:
Fill an application and submit it. Check from the database that the first and lastnames are encrypted.
## Screenshots :camera_flash:
Screenshot 2023-03-29 at 13 15 35
Screenshot 2023-03-29 at 13 15 22

Hash keys for the first and last name need to be set as env variables in production.


mjturt (approved):

backend/benefit/applications/migrations/0029_history_date_format_changes.py
```diff
@@ -0,0 +1,68 @@
+# Generated by Django 3.2.18 on 2023-03-29 07:21
```
sirtawast, line 1: Looks good. How are these new date migrations related?

sirtawast: Or is it something that should've been included in earlier work?

PR #1979, opened 2023-04-17
Fix issue with benefit service suddenly bailing out with CSRF errors when POSTing

sirtawast:
## Description :sparkles:

* Problems started suddenly with port :3100 (handler UI)
* I then changed `CSRF_TRUSTED_ORIGINS` to `https://localhost:3000` and `https://localhost:3100`
  * Mistake: `https://` is [only a django >= 4 thing](https://docs.djangoproject.com/en/4.2/ref/settings/#csrf-trusted-origins) but [django 3 works differently](https://docs.djangoproject.com/en/3.2/ref/settings/#csrf-trusted-origins)
  * Both services were now broken
* Fix issue in `settings.py` to take the ENV in string format (did not parse right with multiple entries coming from .env file)
* Change default `settings.py` to localhost:3000 and localhost:3100 (is this bad 🤔 )


rikuke (approved):
This is OK, the CRS_TRUSTED_ORIGINS are set in the devops library for dev/test/stage/prod separately so no harm done if localhost:3001  and 3100 are default values here.

PR #1995, opened 2023-04-25
Hl 739 talpa csv

rikuke:
## Description :sparkles:
Remove unnecessary columns from the csv report that is downloaded from the handlers reports view and sent to Talpa.


mjturt (approved):
:+1:

PR #2045, opened 2023-05-25
HL-617 | Fix flashing unstyled HTML elements on first page load

sirtawast:
## Description :sparkles:

Fix unstyled elements popping up before styles had been loaded by adding some CSS to add a mask and Javascript to remove the mask as client gets ready.

# Before

ugly-load

# After

smooth-load


frontend/benefit/applicant/src/styles/globals.css
```diff
@@ -13,3 +13,6 @@ ul, li {
   padding: 0;
 }
 
+.app-load-wrapper > div {
+  display: none;
```
sirtawast, line 17: Hide loading spinner in applicant app

mjturt (approved):

PR #2095, opened 2023-06-06
HL-756 | Validate state_aid_max_percentage

rikuke:
## Description :sparkles:
The handler form fails silently if there is no option selected from the State Aid Maximum Percentage select element.
This PR extends the backend validations for the pay subsidy calculation to validate that the State Aid Maximum Percentage is selected and displays an error if is it not.

## Testing :alembic:
Navigate to the handler form, open an application for handling, try to create a calculation without selectin the State Aid Maximum Percentage.
## Screenshots :camera_flash:
Screenshot 2023-06-06 at 15 11 36

Investigating the issue, I noticed that there is some code on the frontend for validating the input fields, but enabling the frontend validation breaks the layout and the validations  as the date inputs use the same keys.
Fixing this is assigned to a separate ticket [HL-826](https://helsinkisolutionoffice.atlassian.net/browse/HL-826?atlOrigin=eyJpIjoiOTczYWI5ZTZlNWRhNGVjY2JlMWU3YzkwMGVjMThhYzkiLCJwIjoiaiJ9)

[HL-826]: https://helsinkisolutionoffice.atlassian.net/browse/HL-826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/benefit/calculator/api/v1/serializers.py
+++ b/backend/benefit/calculator/api/v1/serializers.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 from dateutil.relativedelta import relativedelta
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
@@ -260,6 +262,11 @@ class PaySubsidySerializer(serializers.ModelSerializer):
             and not self._is_manual_mode()
         )
 
+    def _is_invalid_state_aid_max(self, state_aid_max_input: Union[None, int]) -> bool:
+        return state_aid_max_input is None or state_aid_max_input not in [
+            choice[0] for choice in STATE_AID_MAX_PERCENTAGE_CHOICES
+        ]
+
     def validate(self, data):
         request = self.context.get("request")
         if request is None:
@@ -276,12 +283,7 @@ class PaySubsidySerializer(serializers.ModelSerializer):
             state_aid_max_input = self.context["request"].data["calculation"][
                 "state_aid_max_percentage"
             ]
-            if (
-                state_aid_max_input is None
-                or state_aid_max_input == ""
-                or state_aid_max_input
-                not in [choice[0] for choice in STATE_AID_MAX_PERCENTAGE_CHOICES]
-            ):
+            if self._is_invalid_state_aid_max(state_aid_max_input):
                 raise serializers.ValidationError(
                     {
                         "state_aid_max_percentage": _(
--- a/backend/benefit/calculator/api/v1/serializers.py
+++ b/backend/benefit/calculator/api/v1/serializers.py
@@ -280,7 +280,7 @@ class PaySubsidySerializer(serializers.ModelSerializer):
                 raise serializers.ValidationError(
                     {"end_date": _("End date cannot be empty")}
                 )
-            state_aid_max_input = self.context["request"].data["calculation"][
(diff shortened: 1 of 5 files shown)


backend/benefit/calculator/api/v1/serializers.py
```diff
@@ -272,6 +272,15 @@ def validate(self, data):
                 raise serializers.ValidationError(
                     {"end_date": _("End date cannot be empty")}
                 )
+            calculation = self.context["request"].data["calculation"]
```
mjturt, line 275: Seems that `request` is already in the context and `data` also maybe?

backend/benefit/calculator/api/v1/serializers.py
```diff
@@ -272,6 +272,15 @@ def validate(self, data):
...
+            if calculation["state_aid_max_percentage"] is None:
```
mjturt, line 276: What about other falsy values like empty strings?

backend/benefit/calculator/api/v1/serializers.py
rikuke: Screenshot 2023-06-13 at 11 48 25

rikuke: In this stage of executio, data does not have any knowledge of the calculation.

backend/benefit/calculator/api/v1/serializers.py
```diff
@@ -272,6 +273,22 @@ def validate(self, data):
                 raise serializers.ValidationError(
                     {"end_date": _("End date cannot be empty")}
                 )
+            state_aid_max_input = self.context["request"].data["calculation"][
+                "state_aid_max_percentage"
+            ]
+            if (
```
mjturt, line 279: Could be written:

```
if (
    not state_aid_max_input or state_aid_max_input
    not in [choice[0] for choice in STATE_AID_MAX_PERCENTAGE_CHOICES]
):
    ...
```

or if the `STATE_AID_MAX_PERCENTAGE_CHOICES`  were defined as a Django enum, like something like this:
```
class STATE_AID_MAX_PERCENTAGE_CHOICES(models.IntegerChoices):
    FIFTY = 50, "50%"
    HUNDRED = 100, "100%"
```

then there would be attributes `.values` and `.choices` available and the statement could be written:
```
if not state_aid_max_input or state_aid_max_input not in STATE_AID_MAX_PERCENTAGE_CHOICES.values:
    ...
```
but… don't know if it's really worth the trouble.

backend/benefit/locale/fi/LC_MESSAGES/django.po
```diff
@@ -512,6 +512,9 @@ msgstr "Aloituspäivä ei voi olla tyhjä"
 msgid "End date cannot be empty"
 msgstr "Päättymispäivä ei voi olla tyhjä"
 
+msgid "State aid maximum percentage cannot be empty"
```
mjturt, line 515: Should there be translations for other languages too?

mjturt (approved):
Some minor nitpicks. Approving.

PR #2097, opened 2023-06-07
TETP-276(bug): JobPostingSearch keywords workFeaturesList

mikkojamG:
## Description :sparkles:

JobPostingSearch keywords can put cause application to crash if no matching keywords found in workFeaturesList

## Issues :bug:

In useEffect should check to both keywords and workFeaturesList are not undefined

## Testing :alembic:

Testing on local machine


arpargo (approved):
LGTM

PR #2115, opened 2023-06-13
HL-836 | Fix pdf viewer issues 

sirtawast:
## Description :sparkles:

* Resolve rendering issues
* Update react-pdf again

## Issues

~Loading the worker from `public/` is not elegant and webpack should probably load it but webpack is keen on generating the file with hash included in the name for cache busting purposes. That won't allow the resource to be loaded from a single static path so we need to revisit that after NextJs12 upgrade, which has changes in the asset pipeline.~

UPDATE: Loads from webpack now, with considerably smaller JS footprint ~500KB.

### DIFF
--- a/frontend/benefit/applicant/package.json
+++ b/frontend/benefit/applicant/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@frontend/shared": "*",
     "@frontend/benefit-shared": "*",
-    "@react-pdf/renderer": "^3.1.10",
+    "@react-pdf/renderer": "^3.1.12",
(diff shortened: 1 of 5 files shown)


mjturt (commented):
Don't really like the idea of keeping deps in repo. This should be eventually fixed properly, but should we for now consider this as a temporary fix? What @rikuke thinks?

Some info here https://github.com/wojtekmaj/react-pdf/issues/97

mjturt (approved):
Approved, if tests really fail, they should be fixed before merge.

PR #2141, opened 2023-06-21
HL-856 | Show deminimis section for company and association

rikuke:
## Description :sparkles:
Making it mandatory to choose a value from deminimis aid radio button in #2132 made it possible for an association user to get stuck on the first step of the form. This PR fixes the validation for the deminimis input so that when user is from an association, and has no business activity, the deminimis yes/no radio input is not visible and is not require. When the user represents a company, the deminimis yes/no radio input is visible and  required .
## Screenshots :camera_flash:
Screenshot 2023-06-21 at 17 55 14


sirtawast:
Backend build process failed because of kubernetes hangup. Could not rerun build as it was too old. Force pushed to run pipelines again.

sirtawast (approved):

PR #2143, opened 2023-06-22
feat: use openDrawer param on application page to open sidebar

sirtawast:
## Description :sparkles:

* Open sidebar with `openDrawer` param on application view
* Add link to the icon indicating there's new messages
* If there are new messages, add `openDrawer` param in the text and icon link 

sidebar-links

### DIFF
--- a/frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
+++ b/frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
@@ -30,12 +30,13 @@ export interface ApplicationListProps {
 
 const buildApplicationUrl = (
   id: string,
-  status: APPLICATION_STATUSES
+  status: APPLICATION_STATUSES,
+  openDrawer = false
 ): string => {
   if (status === APPLICATION_STATUSES.DRAFT) {
     return `${ROUTES.APPLICATION_FORM}?id=${id}`;
   }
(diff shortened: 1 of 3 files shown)


frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
```diff
@@ -30,12 +30,13 @@ export interface ApplicationListProps {
...
   id: string,
-  status: APPLICATION_STATUSES
+  status: APPLICATION_STATUSES,
+  openDrawer = false
 ): string => {
   if (status === APPLICATION_STATUSES.DRAFT) {
     return `${ROUTES.APPLICATION_FORM}?id=${id}`;
   }
-  return `${ROUTES.APPLICATION}?id=${id}`;
+  return `${ROUTES.APPLICATION}?id=${id}${openDrawer ? '&openDrawer=1' : ''}`;
```
mjturt, line 39: Starts to feel a bit fragile. Also, in query params if the parameter exists, it can be considered true.

mjturt (approved):

frontend/benefit/handler/src/components/applicationList/ApplicationList.tsx
sirtawast: Good catch! Resolving this with a function to check for param's value where it's used/needed.

PR #2156, opened 2023-07-23
HL-642: Handler state check icons

mjturt:
## Description :sparkles:

In handling view, the check icons can now be clicked.

image


sirtawast:
Not sure why context / provider pattern is needed here as data is handled on application item level only and not needed globally (my guess at least?). Other than that, looks good to me! Good work 👍

mjturt:
> Not sure why context / provider pattern is needed here as data is handled on application item level only and not needed globally (my guess at least?). Other than that, looks good to me! Good work +1

It's for the review state and related update function. Context provider reduces a lot of prop drilling here at least.

sirtawast (approved):

PR #2287, opened 2023-09-22
feat: add platform definitions to docker compose files

arpargo:
To ensure that mac m1&m2 chips use amd images instead of the arm ones. This way we all should be using the same base images.


karisal-anders (approved):

rikuke (approved):
Confirmed that this works with Apple Silicon as well.

PR #2653, opened 2023-12-21
Update Dockerfile

Vitals9367:
This should stop OpenShift build from testing different image registries and just pull it from dockerhub straight away


sirtawast:
@Vitals9367 merged to main already, closing this.

PR #2918, opened 2024-04-05
fix(applicant): add more logic to disable/enable save and close button

sirtawast:
## Description :sparkles:

_Save and close_ button was disabled if any field on dataset `employee` or `company` was changed to an invalid value. 

Logic is as follows:

Rather than trying to compare `touched`, a single field, against a `errors` data set:

1. flatten both nested object structures to single level objects where the key is of dot notation
2. use substring to match the touched and errored field
3. ... and use substring rather than equals to ignore any min, max, required, ... validation related additional data which the original `errors` object contains.


terovirtanen:
**API branch is deployed to platta: https://helsinkilisa-pr2918.api.dev.hel.ninja/healthz** :rocket::rocket::rocket:

terovirtanen:
**HANDLER branch is deployed to platta: https://helsinkilisa-ui-handler-pr2918.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**APPLICANT branch is deployed to platta: https://helsinkilisa-ui-pr2918.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://helsinkilisa-ui-handler-pr2918.dev.hel.ninja :laughing::tada::tada::tada:

terovirtanen:
## TestCafe result is __success__ for https://helsinkilisa-ui-pr2918.dev.hel.ninja :laughing::tada::tada::tada:

frontend/benefit/applicant/src/components/applications/forms/application/utils.ts
```diff
@@ -1,8 +1,42 @@
...
+      obj[key] !== null &&
+      !Array.isArray(obj[key])
+    ) {
+      flattenObject(obj[key] as unknown as FlattenedObject, newKey, result);
+    } else {
+      appendResult[newKey] = obj[key];
+    }
+  });
+  return result;
+}
```
EmiliaMakelaVincit, line 25: The implementation is sound but the typing is iffy, yeah. The input is already assigned to FlattenedObject – assumed to be an object with only string values – which doesn't make sense since the whole point of the function is to flatten nested objects onto one level.

Here's how I would type this. The effective FormikTouched and FormikErrors types already are the same shape as the SimpleObject type given, which means the casts in the next function can also be removed altogether.

```ts
type SimpleObjectNode = SimpleObject | Array<SimpleObjectNode> | string | number | boolean | null;
type SimpleObject = { [prop: string]: SimpleObjectNode };

function isSimpleObject(a: SimpleObjectNode): a is SimpleObject {
  return typeof a === 'object' && a !== null && !Array.isArray(a);
}

function flattenObject(
  obj: SimpleObject,
  parentKey = '',
  result: { [prop: string]: Exclude<SimpleObjectNode, SimpleObject> } = {}
): SimpleObjectNode {
  const appendResult = result;
  Object.keys(obj).forEach((key) => {
    const newKey = `${parentKey}${parentKey ? '.' : ''}${key}`;
    const item = obj[key];
    if (isSimpleObject(item)) {
      flattenObject(item, newKey, result);
    } else {
      appendResult[newKey] = item;
    }
  });
  return result;
}
```

In our use case, where we only care about the object-ness of each leaf and the keys of the output, we could even define SimpleObjectNode as `SimpleObject | unknown` and I believe it would still work.

frontend/benefit/applicant/src/components/applications/forms/application/utils.ts
```diff
@@ -1,8 +1,42 @@
...
+ * of single fields that are touched. Thus substring is used to match touched fields against error fields.
+ * */
+export const findIntersectionOfTouchedAndErroredFields = (
   touched: FormikTouched<Partial<Application>>,
   errors: FormikErrors<Partial<Application>>
 ): string[] =>
-  Object.keys(touched).filter((field) => Object.keys(errors).includes(field));
+  Object.keys(flattenObject(touched as FlattenedObject)).filter((field) =>
+    Object.keys(flattenObject(errors as FlattenedObject)).some((errorField) =>
+      errorField.includes(field)
```
EmiliaMakelaVincit, line 40: With the above suggestion, these could just be passed as is.

EmiliaMakelaVincit (approved):
Functionally, seems good. Do consider using the given suggestion for the type woes before merging.

frontend/benefit/applicant/src/components/applications/forms/application/utils.ts
EmiliaMakelaVincit: Oh, and `flattenObject` is being run for errors quite a few times if there are many touched fields. I doubt there's much of a reason to fear performance issues here, but the flattened objects could also be optionally assigned to variables first before this two-level lookup happens.

terovirtanen:
**HANDLER branch is deployed to platta: https://helsinkilisa-ui-handler-pr2918.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://helsinkilisa-ui-handler-pr2918.dev.hel.ninja :laughing::tada::tada::tada:

PR #2926, opened 2024-04-10
fix: failing ahjo token refresh

rikuke:
## Description :sparkles:
The retrieval of the AHJO rest api access token was previously failing because the expiration date was based on created_at and the token  refresh was updating an existing token which kept the create_at date unchanged. This PR changes the logic so that an access token is only ever created and custom exceptions are raised if anything token-related is not behaving as intended or is configured incorrectly. 
Also adds `expiry_datetime()` and `has_expired()` methods and a  new string representation for the `AhjoToken` class, which can be used in error logs etc.
## Testing :alembic:
With (huolto.hel.fi) BIG-IP EDGE VPN turned on:
Get an ahjo code from  [here](https://johdontyopoytahyte.hel.fi/ids4/Account/Login?ReturnUrl=%2Fids4%2Fconnect%2Fauthorize%2Fcallback%3Fscope%3Dopenid%2520offline_access%26response_type%3Dcode%26redirect_uri%3Dhttps%253A%252F%252Fhelsinkilisa%252Fdummyredirect.html%26client_id%3Dhelsinkilisa)

Set an ahjo_code in  Django admin.
In python shell:
```
from applications.services.ahjo_authentication import  AhjoConnector
connector = AhjoConnector()
connector.is_configured() # should return True if all necessary env variables are set

# this only needs to be run once, if this returns an invalid grant error, get and save a new ahjo_code in django admin
connector.get_initial_token()
# check that you have a token in the db now
token = connector.get_token_from_db()
print(token)
# after initial token is set, the token can be refreshed N number of times with:
connector.refresh_token()
```

### DIFF
--- a/backend/benefit/applications/services/ahjo_authentication.py
+++ b/backend/benefit/applications/services/ahjo_authentication.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Dict, Union
 
 import requests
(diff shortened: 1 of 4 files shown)


terovirtanen:
**APPLICANT branch is deployed to platta: https://helsinkilisa-ui-pr2926.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
## TestCafe result is __success__ for https://helsinkilisa-ui-pr2926.dev.hel.ninja :laughing::tada::tada::tada:

sirtawast (approved):

backend/benefit/applications/services/ahjo_authentication.py
```diff
@@ -23,12 +44,12 @@ def __init__(self) -> None:
...
 
-        self.grant_type_for_auth_token: str = "authorization_code"
-        self.grant_type_for_refresh_token: str = "refresh_token"
+        self.grant_type_for_auth_token: str = AUTH_TOKEN_GRANT_TYPE
+        self.grant_type_for_refresh_token: str = REFRESH_TOKEN_GRANT_TYPE
         self.headers: Dict[str, str] = {
             "Content-Type": "application/x-www-form-urlencoded",
         }
-        self.timeout: int = 10
+        self.timeout: int = 60
```
sirtawast, line 52: Should this be ENV controlled?

rikuke: I'm thinking of refactoring this at some point to  a AhjoClient class which would handle all Ahjo requests as now they are done all over the ahjo_integration.py , so I could do this at that point.

PR #3111, opened 2024-06-24
fix: remove list items that are handled with new ahjo mode

sirtawast:
## Description :sparkles:

Filter out applications that should not be shown if old ahjo batch process is used.


terovirtanen:
**API branch is deployed to platta: https://helsinkilisa-pr3111.api.dev.hel.ninja/healthz** :rocket::rocket::rocket:

PR #3979, opened 2026-04-09
fix(deps): upgrade pdfjs-dist and react-pdf to fix CVE security alert

timwessman:
Upgrade pdfjs-dist from 3.11.174 to 4.8.69 and react-pdf from ^8.0.2 to ^9.2.1 to fix Dependabot alerts #391 and #409 (arbitrary JavaScript execution via malicious PDF).

Update webpack rule and worker setup to use new URL/import.meta.url pattern, update CSS import paths, and replace private _pdfInfo.numPages with the public numPages API from pdfjs-dist v4.

Also fix commitlint hook to resolve config from frontend/ where packages are installed.


karisal-anders (approved):
Approval from Kesäseteli side, @ToivoMattilaHelsinki FYI this changes `./husky/commit-msg` just so you know as you're working on pre-commit related stuff currently.

PR #3983, opened 2026-04-09
YJDH-820 | Renew audit logging

karisal-anders:
## Description :sparkles:

- Add audit logging using `django-auditlog` & `django-auditlog-extra`.
- Remove collection of audit log using the old audit logging solution i.e. `AuditLoggingModelViewSet` from `shared.audit_log`.

TODO (Not in this PR):
- Use django-resilient-logger to send new audit logs to Elasticsearch
  - Not possible at the moment as the Elasticsearch we're using in production is Elasticsearch v7, and django-resilient-logger requires Elasticsearch v8, which is incompatible with Elasticsearch v7.
- Specific audit logging concerns related to Django admin
  - These were not handled by the previous audit logging solution in use, so no use case / support has been removed for these in this PR. The only thing related to Django admin in the previous audit logging solution is that there is a `AuditLogEntryAdmin` that is used for e.g. showing the previous audit logging solution's audit log entries in Django admin. This has been left intact.

## Related

[YJDH-820](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-820)

[YJDH-820]: https://helsinkisolutionoffice.atlassian.net/browse/YJDH-820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/kesaseteli/applications/services.py
+++ b/backend/kesaseteli/applications/services.py
@@ -1,7 +1,9 @@
 import logging
 from typing import TYPE_CHECKING
 
+from auditlog.models import LogEntry
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.template import Context, Template
 from django.template.exceptions import TemplateDoesNotExist
 from django.template.loader import get_template
@@ -13,10 +15,14 @@ from applications.models import EmailTemplate, School, SummerVoucherConfiguratio
 from common.utils import send_mail_with_error_logging
 
 if TYPE_CHECKING:
+    from django.contrib.contenttypes.models import ContentType
+
     from applications.models import YouthApplication
 
 
 LOGGER = logging.getLogger(__name__)
+
+User = get_user_model()
 
 
 class TargetGroupValidationService:
@@ -211,3 +217,51 @@ class SchoolService:
                 else:
                     existing_count += 1
         return created_count, existing_count
+
+
+class AuditAccessLogService:
+    """
+    Service for creating ACCESS audit log entries with additional_data.
+    """
+
+    @staticmethod
+    def create_access_log_entry_with_no_related_object_instance(
+        *,
+        actor: User,
+        actor_email: str,
+        content_type: "ContentType",
+        additional_data: dict,
+    ) -> LogEntry:
+        """
+        Create an ACCESS audit log entry with no related object instance,
+        but with additional data.
+        """
+        return LogEntry.objects.create(
+            action=LogEntry.Action.ACCESS,
+            actor=actor,
+            actor_email=actor_email,
+            content_type=content_type,
+            additional_data=additional_data,
+        )
+
+    @staticmethod
+    def create_access_log_entry_with_related_object_and_additional_data(
+        *,
+        accessed_instance,
+        actor: User,
+        actor_email: str,
+        additional_data: dict,
+    ) -> LogEntry:
+        """
+        Create an ACCESS audit log entry with related object instance and
+        additional data into which "is_sent" and "request_path" info
+        can be set later.
+        """
+        return LogEntry.objects.log_create(
+            accessed_instance,
+            force_log=True,
+            action=LogEntry.Action.ACCESS,
+            actor=actor,
+            actor_email=actor_email,
+            additional_data=additional_data,
+        )
--- a/backend/kesaseteli/applications/views.py
+++ b/backend/kesaseteli/applications/views.py
@@ -7,6 +7,7 @@ from typing import List, Union
 
 import xlsx_streaming
 from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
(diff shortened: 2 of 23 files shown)


backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -748,3 +752,6 @@
 SUMMER_VOUCHER_DEFAULT_VOUCHER_VALUE = 350
 SUMMER_VOUCHER_DEFAULT_MIN_WORK_COMPENSATION = 500
 SUMMER_VOUCHER_DEFAULT_MIN_WORK_HOURS = 60
+
+# Load auditlog settings
+from kesaseteli.auditlog_settings import *  # noqa: E402, F403
```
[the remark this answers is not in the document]
karisal-anders: The point is fair, but the so called best fix is bad—actually worse than the original code.

karisal-anders:
[Force pushed](https://github.com/City-of-Helsinki/yjdh/compare/4ba5138e3858e62708899fe803414cadd00b10dd..09eedc76f77f315f0ef1ae2702916eec603fe1ab) changing `ContentType.objects.get` → `ContentType.objects.get_for_model`.

backend/kesaseteli/kesaseteli/auditlog_settings.py
```diff
@@ -0,0 +1,56 @@
...
+    "auth.permission",
+    {
+        "model": "auth.user",
+        "exclude_fields": ["last_login"],  # To reduce excessive logging
+    },
+    "auth.user_groups",
+    "auth.user_user_permissions",
+    "companies.company",
+    "contenttypes.contenttype",
+    "sequences.sequence",
```
nikomakela: Content types and sequences should be ignored.  Sequence number is anyway included in summer voucher serial number field. It does not make sense to auditlog change of auto incerement number. Content types are pure system instances.

backend/kesaseteli/applications/views.py
```diff
@@ -261,28 +277,32 @@ def serializer(self, apps: QuerySet[YouthApplication]):
...
-            )
-            return response
+        # Manually log access because of mass access to sensitive information:
+        LogEntry.objects.create(
+            action=LogEntry.Action.ACCESS,
+            actor=request.user,
+            actor_email=request.user.email,
+            content_type=ContentType.objects.get_for_model(YouthApplication),
+            additional_data={"method": f"{self.__class__.__name__}.list"},
+        )
```
nikomakela: It's strange that it's allowed to create and save LogEntry objects without `object_ok`, because they are `blank=False` and `null=False` in django-auditlog (ref: https://github.com/jazzband/django-auditlog/blob/7e33d8261ea9496dabeed0283205f006219f149b/auditlog/models.py#L348-L353). However, when tested, it really seems to be possible -- the object is persisted in database, with null and empty string set to those fields.

I understand that this is LogEntry (and the 2-3 other similar in this same PR) are to write "access log" to "audit log", meaning "this user exported excel with lots of application and voucher instances". The normal way would be to write access log of every instance independently, but it would be a heavy write process (and excel file can anyway be shared to anyone after that).

Since this is not a normal way to use django-auditlog, I would at least wrap this feature to a new service function to add 1 abstraction level. It could be a (service) function that "writes access log without pk" or "logs access without pk" or "logs access to an endpoint or process", etc. Later, it would be easier to find these custom django-auditlog usages when they are wrapped (and documented) like this.

nikomakela (approved):
2 types to be ignored and a suggestion to add a service function to write access log without pk. Approved in advance.

backend/kesaseteli/applications/views.py
karisal-anders: Moved to AuditAccessLogService in [force push](https://github.com/City-of-Helsinki/yjdh/compare/09eedc76f77f315f0ef1ae2702916eec603fe1ab..abb469e7cae94a592f953b9b9782d8d5d0e4a811)

backend/kesaseteli/kesaseteli/auditlog_settings.py
karisal-anders: Taken into account with [force push](https://github.com/City-of-Helsinki/yjdh/compare/09eedc76f77f315f0ef1ae2702916eec603fe1ab..abb469e7cae94a592f953b9b9782d8d5d0e4a811)

backend/kesaseteli/applications/services.py
```diff
@@ -13,11 +15,15 @@
 from common.utils import send_mail_with_error_logging
 
 if TYPE_CHECKING:
+    from django.contrib.contenttypes.models import ContentType
```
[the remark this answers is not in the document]
karisal-anders: It's being used in static type checking later in the file with `"ContentType"`.

PR #3996, opened 2026-04-14
Fix commitlint config path in github actions

karisal-anders:
### Fix commitlint config path in github actions

This fixes running commitlint in github actions with the wanted
commitlint.config.mjs, and thus allows "deps" type and does not break
on dependabot PRs anymore.

Refs: PR #3978 (moved commitlint.config.mjs from root to /frontend/)

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -20,4 +20,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: wagoid/commitlint-github-action@v6
         with:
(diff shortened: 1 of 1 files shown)


voneiden (approved):

nikomakela (approved):

timwessman:
Sorry, that was my fault for moving the commitlint config to the frontend folder unnecessarily. I reverted that change https://github.com/City-of-Helsinki/yjdh/pull/3997 ..so after that revert, this pr is not needed

karisal-anders:
Using PR #3997 instead.

PR #3997, opened 2026-04-14
chore: revert commitlint mess

timwessman:
I made an unnecessary mess by moving the commitlint config to the frontend folder. Now let's revert it back to the root.


voneiden:
Closes #3996

karisal-anders (approved):

PR #4004, opened 2026-04-17
Render birthdate instead of SSN in emplyers form and fix summer voucher validation

nikomakela:
YJDH-841.

# Backend

1. Improved serializer validation and linking of employer summer voucher
into employer application. An employer user and his colleagues should be
able to edit company's applications, sumemr vouchers and application
attachments.
2. Added a birthdate field into employer application endpoint's
response.
3. Improvements to admin site views.
4. Employer application filter now defaults showing companys applications without explicitly defining `only_mine` -parameter
5. Fix integrity errors with audit logging in some tests.

# Frontend

Replace SSN with a birthdate field. Social security number should not be
visible anywhere in employers UI anymore. When employee data is fetched,
a birthdate is shown instead of SSN.

Other misc updates:
1. Added timeout to API requests
2. Handle and postcode fields as texts of 5 chars length and mock the
values in tests with postcode mocker.
3. Fixes to reset form values method and loading the values from session
storage.

# Shared

1. Fix input filling and improve post code mocking.

### DIFF
--- a/backend/kesaseteli/applications/admin.py
+++ b/backend/kesaseteli/applications/admin.py
@@ -10,6 +10,7 @@ from django.http import HttpResponse, HttpResponseRedirect
 from django.shortcuts import render
 from django.urls import path, reverse
 from django.utils import timezone
+from django.utils.formats import date_format
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
 from django.utils.translation import ngettext
@@ -371,6 +372,8 @@ class YouthApplicationAdmin(admin.ModelAdmin):
         "id",
         "first_name",
         "last_name",
+        "target_group_display",
+        "birthdate",
         "masked_social_security_number",
         "status",
         "school",
@@ -381,6 +384,7 @@ class YouthApplicationAdmin(admin.ModelAdmin):
     list_filter = [
         "created_at",
         "modified_at",
+        "target_group",
         "status",
         IsValidSchoolFilter,
         SchoolFilter,
@@ -417,6 +421,21 @@ class YouthApplicationAdmin(admin.ModelAdmin):
     # currently in school list.
     is_valid_school.short_description = _("is in current school list")
 
+    def target_group_display(self, obj):
+        return obj.get_target_group_display()
+
+    target_group_display.short_description = _("target group")
+
+    def birthdate(self, obj):
+        return obj.birthdate
+
+    birthdate.short_description = _("birthdate")
+
+    def birthdate_display(self, obj):
+        return date_format(obj.birthdate, format="DATE_FORMAT", use_l10n=True)
+
+    birthdate_display.short_description = _("birthdate")
+
     def masked_social_security_number(self, obj):
         """Mask social security number for display."""
         return mask_social_security_number(obj.social_security_number)
@@ -425,8 +444,11 @@ class YouthApplicationAdmin(admin.ModelAdmin):
 
     def get_readonly_fields(self, request, obj=None):
         """Make contact info fields readonly."""
+        custom_readonly_fields = [
+            "birthdate_display",
+        ]
         if obj:
-            return [
+            return custom_readonly_fields + [
                 f.name
                 for f in self.model._meta.fields
                 if f.name not in self.exclude and f.name not in self.contact_info_fields
@@ -458,7 +480,8 @@ class YouthSummerVoucherAdmin(admin.ModelAdmin):
     list_display = [
         "id",
         "summer_voucher_serial_number",
-        "target_group",
+        "target_group_display",
+        "birthdate",
         "youth_application_link",
         "masked_social_security_number",
         "youth_application__email",
@@ -487,6 +510,16 @@ class YouthSummerVoucherAdmin(admin.ModelAdmin):
 
     def get_queryset(self, request):
         return super().get_queryset(request).select_related("youth_application")
+
+    def target_group_display(self, obj):
+        return obj.get_target_group_display()
+
+    target_group_display.short_description = _("target group")
+
+    def birthdate(self, obj):
+        return obj.youth_application.birthdate
+
+    birthdate.short_description = _("birthdate")
 
     def masked_social_security_number(self, obj):
         """Mask social security number for display."""
--- a/backend/kesaseteli/applications/api/v1/permissions.py
+++ b/backend/kesaseteli/applications/api/v1/permissions.py
@@ -1,6 +1,7 @@
 from typing import Optional
 
 from django.http import HttpRequest
+from rest_framework.exceptions import NotFound
 from rest_framework.permissions import BasePermission
 
 from applications.enums import EmployerApplicationStatus
@@ -19,7 +20,10 @@ ALLOWED_APPLICATION_UPDATE_STATUSES = [
 
 
 def get_user_company(request: HttpRequest) -> Optional[Company]:
-    user_company = get_or_create_company_using_organization_roles(request)
+    try:
+        user_company = get_or_create_company_using_organization_roles(request)
+    except NotFound:
+        user_company = None
 
     return user_company
 
--- a/backend/kesaseteli/applications/api/v1/permissions.py
+++ b/backend/kesaseteli/applications/api/v1/permissions.py
@@ -30,12 +30,14 @@ def has_employer_application_permission(
     """
     Allow access only to DRAFT status employer applications of the user's company.
     """
-    return (
-        request.user.is_staff
-        or request.user.is_superuser
-        or (
-            employer_application.company == get_user_company(request)
-            and employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
+    user_company = get_user_company(request)
+    return bool(
+        user_company
+        and employer_application.company == user_company
+        and (
+            request.user.is_staff
+            or request.user.is_superuser
+            or employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
         )
     )
 
--- a/backend/kesaseteli/applications/api/v1/permissions.py
+++ b/backend/kesaseteli/applications/api/v1/permissions.py
@@ -32,17 +32,22 @@ def has_employer_application_permission(
     request: HttpRequest, employer_application: EmployerApplication
 ) -> bool:
     """
-    Allow access only to DRAFT status employer applications of the user's company.
+    Allow access to employer applications.
+    - Staff and superusers have full access.
+    - Standard users must belong to the same company as the application.
+    - For standard users, only DRAFT and SUBMITTED applications are viewable.
     """
+    if request.user.is_staff or request.user.is_superuser:
+        return True
+
+    # It is important to check the company permission, because
+    # at some point user might also lose the company permission.
+    # User should not be able to access applications of companies they don't belong to.
     user_company = get_user_company(request)
     return bool(
         user_company
         and employer_application.company == user_company
-        and (
-            request.user.is_staff
-            or request.user.is_superuser
-            or employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
-        )
+        and employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
     )
 
 
--- a/backend/kesaseteli/applications/api/v1/serializers.py
+++ b/backend/kesaseteli/applications/api/v1/serializers.py
@@ -17,6 +17,7 @@ from applications.enums import (
     AttachmentType,
     EmployerApplicationStatus,
     get_supported_languages,
+    YouthApplicationStatus,
 )
 from applications.models import (
     Attachment,
@@ -25,11 +26,13 @@ from applications.models import (
     School,
     SummerVoucherConfiguration,
     YouthApplication,
+    YouthSummerVoucher,
 )
 from applications.target_groups import (
     get_target_group_choices,
     get_target_group_data,
 )
+from common.fuzzy_matching import is_last_name_fuzzy_match_in_full_name
 from common.permissions import HandlerPermission
 from companies.api.v1.serializers import CompanySerializer
 from companies.services import get_or_create_company_using_organization_roles
@@ -232,6 +235,11 @@ class EmployerSummerVoucherSerializer(serializers.ModelSerializer):
         allow_blank=True,
         required=False,
     )
+    employee_name = serializers.CharField(
+        max_length=256,
+        allow_blank=True,
+        required=False,
+    )
 
     class Meta:
         model = EmployerSummerVoucher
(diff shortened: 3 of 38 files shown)


backend/kesaseteli/applications/api/v1/permissions.py
```diff
@@ -30,11 +30,18 @@ def has_employer_application_permission(
...
+
+    if employer_application.company != user_company:
+        return False
+
     return (
         request.user.is_staff
         or request.user.is_superuser
         or (
-            employer_application.company == get_user_company(request)
+            employer_application.company == user_company
```
karisal-anders: To get rid of the redundant check here (it's handled already by the early return in this function now) here's one version:
```python
def has_employer_application_permission(
    request: HttpRequest, employer_application: EmployerApplication
) -> bool:
    """
    Allow access only to DRAFT status employer applications of the user's company.
    """
    user_company = get_user_company(request)
    return bool(
        user_company
        and employer_application.company == user_company
        and (
            request.user.is_staff
            or request.user.is_superuser
            or employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
        )
    )
```

That's run through ruff so it's formatted ok.

backend/kesaseteli/applications/tests/test_employer_summer_voucher_validation.py
```diff
@@ -0,0 +1,172 @@
...
+                return_value=application,
+            ),
+        ):
+            data = EmployerApplicationSerializer(application).data
+            data["street_address"] = "New Company Address"
+
+            response = api_client.put(get_detail_url(application), data)
+            assert response.status_code == status.HTTP_200_OK, response.data
+            application.refresh_from_db()
+            assert application.street_address == "New Company Address"
```
karisal-anders: How exactly does this verify that a company colleague can update a draft application belonging to their company? Some functions have been patched to return specific values, but it's not evident to me right away that this tests what the docstring says this tests. We can discuss this live for speed also!

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -769,8 +770,14 @@ def get_queryset(self):
 
         user_company = get_user_company(self.request)
```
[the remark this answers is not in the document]
nikomakela: I will do this

```python
def get_user_company(request: HttpRequest) -> Optional[Company]:
    try:
        return get_or_create_company_using_organization_roles(request)
    except NotFound:
        return None
```

backend/kesaseteli/applications/tests/test_employer_summer_voucher_validation.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        Verify that a voucher cannot be used in two different applications.
+        """
+        # First linkage
+        EmployerSummerVoucher.objects.create(
+            application=application, youth_summer_voucher=active_voucher
+        )
+
+        # Attempt to link to a second application
+        # Attempt to link to a second application.
```
karisal-anders: Deduplication
```suggestion
        # Attempt to link to a second application.
```

backend/kesaseteli/applications/tests/test_security_with_non_staff_user.py
```diff
@@ -377,126 +378,138 @@ def get_attachment(client: Client, attachment: Attachment):
...
+    client_type,
+    context_type,
+    owner_type,
+    app_company_type,
+    expected_status,
+):
     """
     Test that the employer summer voucher's handle attachment endpoint can be
     used to delete draft employer applications' attachments but only those that
     are the user's and use the user's company.
```
karisal-anders:
```suggestion
    are the user's or use the user's company.
```

backend/kesaseteli/applications/admin.py
```diff
@@ -425,8 +444,11 @@ def masked_social_security_number(self, obj):
 
     def get_readonly_fields(self, request, obj=None):
         """Make contact info fields readonly."""
+        çustom_readonly_fields = [
```
karisal-anders: Typo `ç`→`c`:
```suggestion
        custom_readonly_fields = [
```

backend/kesaseteli/applications/admin.py
```diff
@@ -425,8 +444,11 @@ def masked_social_security_number(self, obj):
...
+            self.birthdate_display,
+        ]
         if obj:
-            return [
+            return çustom_readonly_fields + [
```
karisal-anders: `ç`→`c`:
```suggestion
            return custom_readonly_fields + [
```

frontend/kesaseteli/employer/src/components/application/steps/step1/__tests__/EmploymentForm.test.tsx
```diff
@@ -69,21 +67,73 @@ describe('EmploymentForm', () => {
...
+
+      await waitFor(() => {
+        expect(
+          screen.getByText(/common:application.form.errors.maxLength/i)
+        ).toBeInTheDocument();
+      });
+    });
   });
 
   it('activates the second part of the form when employee_phone_number is present (even if SSN is missing)', () => {
```
karisal-anders: How about this? Or maybe removing the "even" part in parenthesis?
```suggestion
  it('activates the second part of the form when employee_phone_number is present (even if birthdate is missing)', () => {
```

backend/kesaseteli/applications/tests/test_employer_summer_voucher_validation.py
nikomakela: NIce catch:

```
No, this test does not actually test what it claims to. It effectively circumvents its own objective by disabling the security mechanisms it's supposed to verify.

Does it test what it tries to test?
No. The test is "cheating" by patching the very gates it should be trying to open.

A "colleague" is never truly verified: A random user from UserFactory() is authenticated, but they pass the test only because the view's filters and permission checks have been replaced with mocks that always return the draft application.
Flawed Logic: If you were to change the authenticated user to a completely unrelated person from a different company, this test would still pass, proving it isn't actually validating "colleague" permissions.
Documentation of the Object Patches
The with patch.object(...) block performs the following overrides:

EmployerApplicationViewSet.get_queryset:

What it does: This method normally filters the database so a user only sees applications belonging to their company.
Effect of patch: It forces the ViewSet to "find" the application regardless of the authenticated user's company context. It disables the logic in views.py (lines 758-782) that validates get_user_company(request).
EmployerApplicationViewSet.get_object:

What it does: In Django Rest Framework, get_object is responsible for retrieving the instance and, crucially, calling self.check_object_permissions(request, obj).
Effect of patch: It returns the application object directly, bypassing the entire permission class system. The EmployerApplicationPermission.has_object_permission logic is never even executed.
```

I will fix it with something like this
```suggestion
    def test_update_permission_allowed_for_company_colleague(self, api_client, company):
        """
        Verify that a company colleague CAN update a draft application belonging to their company.
        """
        # 1. Create a draft application for a specific company.
        application = EmployerApplicationFactory(
            company=company, status=EmployerApplicationStatus.DRAFT
        )

        # 2. Authenticate as a completely different user (the "colleague").
        # Note: In this system, company access isn't tied to the User model itself.
        # Instead, it's tied to the "organization_roles" stored in the active SESSION.
        # The api_client fixture already has the 'company' in its session via `store_company_in_session`.
        api_client.force_authenticate(user=UserFactory())

        # 3. Even though the User changed, the 'api_client' fixture preserves its session.
        # That session already contains 'company.business_id' (set by store_company_in_session).
        # This simulates a different person logged into the same organizational context.
        data = EmployerApplicationSerializer(application).data
        data["street_address"] = "New Company Address"

        # 4. Attempt to update the application via the API.
        # The backend's permission logic will fetch the company from the session,
        # find that it matches the application's company, and allow the update.
        response = api_client.put(get_detail_url(application), data)

        assert response.status_code == status.HTTP_200_OK, response.data
        application.refresh_from_db()
        assert application.street_address == "New Company Address"
 ```

File: frontend/kesaseteli/shared/src/__tests__/utils/fake-objects.ts [hunk 9]
```diff
@@ -209,7 +209,7 @@ export const fakeValidVtjAddress = (
   merge(
     {
       LahiosoiteS: faker.address.streetAddress(),
-      Postinumero: String(faker.datatype.number(99_999)),
+      Postinumero: faker.helpers.replaceSymbols('#####'),
```
karisal-anders: Like above for readability?
```suggestion
      Postinumero: faker.address.zipCode('#####'),
```

backend/kesaseteli/applications/api/v1/permissions.py
nikomakela: used your suggestion.

backend/kesaseteli/applications/tests/test_employer_summer_voucher_validation.py
nikomakela: fixed

backend/kesaseteli/applications/tests/test_security_with_non_staff_user.py
nikomakela: fixed

backend/kesaseteli/applications/admin.py
nikomakela: fixed

backend/kesaseteli/applications/admin.py
nikomakela: fixed

frontend/kesaseteli/employer/src/components/application/steps/step1/__tests__/EmploymentForm.test.tsx
nikomakela: fixed

frontend/kesaseteli/shared/src/__tests__/utils/fake-objects.ts
nikomakela: fixed

backend/kesaseteli/applications/api/v1/permissions.py
```diff
@@ -30,12 +34,14 @@ def has_employer_application_permission(
...
-            employer_application.company == get_user_company(request)
-            and employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
+    user_company = get_user_company(request)
+    return bool(
+        user_company
+        and employer_application.company == user_company
+        and (
+            request.user.is_staff
+            or request.user.is_superuser
+            or employer_application.status in ALLOWED_APPLICATION_VIEW_STATUSES
```
ToivoMattilaHelsinki: If user_company is None (or falsy), we skip checking if the user is staff or superuser
Is this intentional?

backend/kesaseteli/applications/admin.py
```diff
@@ -425,8 +444,11 @@ def masked_social_security_number(self, obj):
...
         """Make contact info fields readonly."""
+        custom_readonly_fields = [
+            self.birthdate_display,
+        ]
         if obj:
-            return [
+            return custom_readonly_fields + [
                 f.name
                 for f in self.model._meta.fields
                 if f.name not in self.exclude and f.name not in self.contact_info_fields
```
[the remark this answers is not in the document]
nikomakela: It's definitely not raising an error here. However, I don't mind it to be a string either. 

This is with `self.birthdate_display`:
image

The same result comes with a string "birthdate_display".

backend/kesaseteli/applications/api/v1/serializers.py
```diff
@@ -288,8 +298,115 @@ def _validate_non_draft_required_fields(self, data):
...
+        serial_number = data.get("summer_voucher_serial_number")
+        if not serial_number:
+            return
+
+        try:
+            youth_summer_voucher = YouthSummerVoucher.objects.get(
+                summer_voucher_serial_number=int(serial_number)
+            )
+        except (ValueError, TypeError, YouthSummerVoucher.DoesNotExist):
+            return
```
ToivoMattilaHelsinki: What does this mean?

If we can't find a YouthSummerVoucher with the serial number, we just exit early and the data is considered ok (instead of i.e. raising a ValidationError)

Just confirming that this is on purpose and serial number we can't find is fine

ToivoMattilaHelsinki (approved):

backend/kesaseteli/applications/api/v1/permissions.py
nikomakela: The staff role is questionable, but superuser basically should have all the permissions anyway. The permission is currently only checked in API viewset and handlers do not yet have any UI for it. Superusers and staff users would probably use admin site instead, but the custom permission is not checked there.

backend/kesaseteli/applications/api/v1/permissions.py
```diff
@@ -30,12 +34,14 @@ def has_employer_application_permission(
...
         )
     )
```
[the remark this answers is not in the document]
nikomakela: The staff role is questionable anyway, but superuser basically should have all the permissions anyway. The permission is currently only checked in API viewset and handlers do not yet have any UI for it. Superusers and staff users would probably use admin site instead, but the custom permission is not checked there.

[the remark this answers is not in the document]
nikomakela: This permissions.py is inconsistent with API view's get_queryset. Since this `permissions.py` issue is also related to failing tests after `only_mine` -filter changes, I will fix the permissions so that staff and superuser have early return, and I will fix the inconsistency with view queryset, since it's related to failing tests too. READ-only access is sufficient for now, so I won't change serializer yet.

backend/kesaseteli/applications/api/v1/permissions.py
nikomakela: https://github.com/City-of-Helsinki/yjdh/pull/4004#discussion_r3100492489

nikomakela:
Fixes to current issues and a more clear refactored logic to handle company's permissions in https://github.com/City-of-Helsinki/yjdh/pull/4005. @karisal-anders @ToivoMattilaHelsinki

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -794,8 +798,13 @@ def update(self, request, *args, **kwargs):
...
             raise ValidationError("Only DRAFT applications can be updated")
-        if request.user != instance.user:
-            raise PermissionDenied("Only application creator can update it")
+
+        # Ensure that visibility and modification are strictly tied to the
+        # organization the user currently represents. Even the original creator
+        # must have an active organization association to update the application.
+        user_company = get_user_company(request)
+        if not user_company or instance.company != user_company:
+            raise PermissionDenied("Only company members can update it")
```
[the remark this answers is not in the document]
nikomakela: That is fine and it's out of scope in this PR. Staff members and superusers are not using the API anyhow and tehy don't have a UI or client for that either. Even if they would have through a handlers' UI, I don't think we would like to support them to update or destroy applications. They need to be able to read them, but probably not manage them.

backend/kesaseteli/applications/api/v1/serializers.py
nikomakela: Related PR merged into this one https://github.com/City-of-Helsinki/yjdh/pull/4004#ref-pullrequest-4283226783

PR #4007, opened 2026-04-20
feat(ks,shared): encourage for full employer street address

nikomakela:
YJDH-846.

Add a helper text to model and employer UI's field of street address.

Screenshot 2026-04-20 at 12 45 16

NOTE: a trivial field addition to shared TextInput that makes the TextInput more compliant with HDS and makes it possible to use helper text.


backend/kesaseteli/applications/models.py
```diff
@@ -1178,7 +1178,11 @@ class EmployerApplication(TimeStampedModel, UUIDModel):
     street_address = models.CharField(
         max_length=256,
         blank=True,
-        verbose_name=_("invoicer work address"),
+        verbose_name=_("postal address"),
+        help_text=_(
+            "Enter the whole postal address including street address, postcode "
+            "and city."
+        ),
```
karisal-anders, line 1188: How about defining whose address this is in verbose name and in help text?
The translations use this as the company's address, so how about "company postal address" in verbose name and "the whole company postal address" in help text? That'd make this a bit clearer IMHO.

Adding some comment above the `street_address` field definition could make things a bit clearer still, as it's been a street address historically, but now it's meant to be a full postal address, and in production database it may contain both types.

karisal-anders (approved):
Left comment about adding a bit of clarity to the `street_address` field's verbose name and helper text. Please take a look at it.

frontend/shared/src/components/forms/inputs/TextInput.tsx
```diff
@@ -104,7 +106,7 @@ const TextInput = <T extends FieldValues>({
             ? String(registerOptions.maxLength)
             : undefined
         }
-        helperText={lengthIndicator}
+        helperText={rest.helperText ?? lengthIndicator}
```
nikomakela: helperText could be destructed in constructor instead

PR #4018, opened 2026-04-22
Ignore navigation prompt on session expiration navigation

nikomakela:
YJDH-845.

When a session expires, user is logged out and taken back to login screen, the UI should not ask wait for confirmation of away navigation. It's a forced system navigation.

NOTE: hook is only used by Kesaseteli Employer's UI.

Improve `useLeaveConfirm` hook's readability, maintainability and testability.

### DIFF
--- a/frontend/shared/src/hooks/__tests__/useLeaveConfirm.utils.test.ts
+++ b/frontend/shared/src/hooks/__tests__/useLeaveConfirm.utils.test.ts
@@ -1,0 +1,101 @@
+import Router from 'next/router';
+import {
+  getPathWithoutHash,
+  isBypassUrl,
+  isInternalLink,
+  shouldBlockNavigation,
+  getAnchorPath,
+} from '../useLeaveConfirm.utils';
+
+jest.mock('next/router', () => ({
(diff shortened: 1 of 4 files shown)


tjgofore (approved):

frontend/shared/src/hooks/__tests__/useLeaveConfirm.utils.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    beforeEach(() => {
+      Router.asPath = '/current';
+    });
+
+    it('should return true if blocking is required', () => {
+      expect(shouldBlockNavigation('/other', true, false, false)).toBe(true);
+    });
+
+    it('should return false if already confirmed', () => {
+      expect(shouldBlockNavigation('/other', true, true, false)).toBe(false);
```
karisal-anders: shouldBlockNavigation function would benefit from using an object parameter / props rather than positional parameters to add clarity/readability.

karisal-anders (approved):
Left a comment, please take a look at it. Not a blocker.

PR #4025, opened 2026-04-24
fix(ks,employer): infinite loop when adfs used

nikomakela:
YJDH-840.

In Employer UI the user should use SuomiFI auth, but if he is using SSO and ADFS instead, he might end up in infinite loop of error messages. This commit fixes the situation.


karisal-anders (approved):
Please check the sentry bot's comment about path vs. router.asPath comparison whether it holds true or not.

PR #4026, opened 2026-04-24
HL-1680 Upgrade HDS to version 5

ToivoMattilaHelsinki:
## Description :sparkles:
HL-1680, YJDH-824

Kesäseteli currently uses Helsinki Design System (HDS) version 3

HDS version 3 was deprecated in the most recent release so we are migrating the project to HDS 5


frontend/kesaseteli/employer/browser-tests/actions/application.actions.ts
```diff
@@ -107,10 +106,9 @@ export const fillEmploymentDetails = async (
 export const fillStep1Form = async (
   t: TestController,
   application: Application,
-  expectedPreFill?: Partial<Employment>
 ): Promise<void> => {
   await fillEmployerDetails(t, application);
-  await fillEmploymentDetails(t, application, expectedPreFill);
```
nikomakela, line 134: Why is the expectedPreFill needed to be removed here? The idea has been that it's not important to really write all the inputs one by one in every test, when the same form is already tested once or many times before. Prefilling makes the test much faster.

Filling the inputs isn't a main scope in every test, so it can be skipped when it's not in test scope.

frontend/kesaseteli/youth/src/__tests__/utils/components/get-additional-info-page-api.ts
```diff
@@ -33,6 +46,35 @@ const getAdditionalInfoPageApi = (
...
+    const byButton = screen.queryByRole('button', {
+      name: regexp(translations.additionalInfo.form.reasons),
+    });
+
+    if (byButton) {
+      return byButton;
+    }
+
+    return screen.findByRole('combobox');
+  };
```
nikomakela: This is a bit strange way to find it, because why wouldn't it be same every time?

frontend/kesaseteli/youth/src/__tests__/utils/components/get-index-page-api.ts
```diff
@@ -30,6 +33,26 @@ const getIndexPageApi = (lang?: Language) => {
...
+    const byLabelCombobox = screen.queryByRole('combobox', {
+      name: regexp(translations.youthApplication.form.selectedSchool),
+    });
+
+    if (byLabelCombobox) {
+      return byLabelCombobox;
+    }
+
+    return screen.findByRole('combobox');
+  };
```
nikomakela: It shouldn't have many different kind of implementations. It should be the same every time so there should be no need to find it using multiple patterns.

frontend/kesaseteli/youth/src/__tests__/utils/components/get-index-page-api.ts
```diff
@@ -54,16 +77,14 @@ const getIndexPageApi = (lang?: Language) => {
...
+        expect(input).toHaveAttribute('aria-disabled', 'true');
       },
       async selectedSchoolIsEnabled(): Promise<void> {
-        const input = await screen.findByRole('combobox', {
-          name: regexp(translations.youthApplication.form.selectedSchool),
-        });
-        await waitFor(() => expect(input).toBeEnabled());
+        const input = await findSelectedSchoolCombobox();
+        await waitFor(() =>
+          expect(input).not.toHaveAttribute('aria-disabled', 'true')
```
nikomakela, line 71: "aria-disabled" is not the main attribute that should be tested, but "disabled".

```
expect(input).toBeDisabled();
```

frontend/kesaseteli/youth/src/__tests__/utils/components/get-index-page-api.ts
```diff
@@ -104,10 +121,13 @@ const getIndexPageApi = (lang?: Language) => {
...
-        const input = await screen.findByRole('combobox', {
-          name: regexp(translations.youthApplication.form.selectedSchool),
-        });
-        expect(input).toBeValid();
+        const input = await findSelectedSchoolCombobox();
+        await waitFor(() =>
+          expect(findSelectedSchoolError()).not.toBeInTheDocument()
+        );
+        await waitFor(() =>
+          expect(input).not.toHaveAttribute('aria-invalid', 'true')
```
nikomakela: "aria-invalid" might be a proepr value here, but something else than aria could be used if any available.

nikomakela:
```
// 1. Check that the attribute is set to true
expect(input).toBeInvalid(); 

// 2. Check that the error message is linked for screen readers
// This ensures aria-describedby matches the ID of your error text
expect(input).toHaveErrorMessage(/invalid email/i);
```

frontend/kesaseteli/youth/src/__tests__/index.base.test.tsx
```diff
@@ -25,7 +25,13 @@ describe('frontend/kesaseteli/youth/src/pages/index.tsx', () => {
...
       const indexPageApi = getIndexPageApi();
       await indexPageApi.expectations.pageIsLoaded();
-      const results = await axe(container);
+      const results = await axe(container, {
+        rules: {
+          'aria-allowed-role': {
+            enabled: false,
+          },
+        },
+      });
```
nikomakela: You should be able to test with `toHaveNoViolations` when using axe.
```
  const results = await axe(container);
  expect(results).toHaveNoViolations();
```

frontend/kesaseteli/shared/src/hooks/useCurrentYearSummerVoucherConfig.tsx
```diff
@@ -1,9 +1,17 @@
...
+  currentYear: number;
+} {
   const query = useSummerVoucherConfigurationQuery();
 
   const currentYear = new Date().getFullYear();
   const currentConfiguration = query.data?.find((c) => c.year === currentYear);
   return { query, currentConfiguration, currentYear };
 }
+
+export default useCurrentYearSummerVoucherConfig;
```
nikomakela: I don't see why a named export is now changed to a default, but sure it does not matter much.

frontend/kesaseteli/youth/src/components/youth-form/TargetGroupSelection.tsx
```diff
@@ -1,5 +1,5 @@
 import useRegisterInput from 'kesaseteli/youth/hooks/useRegisterInput';
-import { useCurrentYearSummerVoucherConfig } from 'kesaseteli-shared/hooks/useCurrentYearSummerVoucherConfig';
+import useCurrentYearSummerVoucherConfig from 'kesaseteli-shared/hooks/useCurrentYearSummerVoucherConfig';
```
nikomakela: Named import has been a choice that I have made earlier.

> While using only named exports is not an absolute rule, it is increasingly considered a best practice in modern JavaScript (especially TypeScript) development for improved consistency, maintainability, and tooling support

frontend/kesaseteli/youth/src/components/youth-form/YouthApplication.tsx
```diff
@@ -1,5 +1,5 @@
 import YouthForm from 'kesaseteli/youth/components/youth-form/YouthForm';
-import { useCurrentYearSummerVoucherConfig } from 'kesaseteli-shared/hooks/useCurrentYearSummerVoucherConfig';
+import useCurrentYearSummerVoucherConfig from 'kesaseteli-shared/hooks/useCurrentYearSummerVoucherConfig';
```
nikomakela: Do we have a linter for default export usage?

> While using only named exports is not an absolute rule, it is increasingly considered a best practice in modern JavaScript (especially TypeScript) development for improved consistency, maintainability, and tooling support

frontend/shared/src/components/attachments/UploadAttachment.tsx
```diff
@@ -54,12 +55,14 @@ const UploadAttachment: React.FC<UploadAttachmentProps> = ({
...
       <Button
         id={name}
+        /* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment */
         theme={theme}
         onClick={handleUploadClick}
         variant={variant}
         isLoading={isUploading}
         loadingText={loadingText}
-        iconLeft={icon}
+        iconEnd={null}
```
nikomakela: Is iconEnd needed to be defined here, if it's set to null? Isn't it the same as undefined?

frontend/shared/src/components/forms/inputs/__tests__/MultiSelectDropdown.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    expect(
+      await screen.findByRole('option', { name: 'Helsinki' })
+    ).toBeInTheDocument();
+    expect(screen.getByRole('option', { name: 'Espoo' })).toBeInTheDocument();
+    expect(screen.getByRole('option', { name: 'Vantaa' })).toBeInTheDocument();
+  });
+
+  it('maps selected options back to domain objects in onChange', async () => {
+    const user = userEvent.setup();
```
nikomakela, line 97: use imported userEvent instead

```
import userEvent from '@testing-library/user-event';
```

frontend/shared/src/components/forms/inputs/__tests__/MultiSelectDropdown.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  it('renders disabled combobox when disabled prop is true', () => {
+    renderComponent({ disabled: true });
+
+    expect(getCombobox()).toHaveAttribute('aria-disabled', 'true');
+  });
+
+  it('commits draft state to form value on menu close', async () => {
+    const user = userEvent.setup();
```
nikomakela, line 129: use imported userEvent instead

`import userEvent from '@testing-library/user-event';`

frontend/shared/src/components/forms/inputs/Dropdown.tsx
```diff
@@ -36,25 +37,55 @@ const Dropdown = <T extends FieldValues, O extends Option>({
...
+    texts: {
+      label: label || '',
+      placeholder,
+      ...(filterLabel
+        ? { filterLabel}
+        : {}),
+      ...(filterPlaceholder
+        ? { filterPlaceholder}
+        : {}),
+    },
```
nikomakela: This is really hard to read. How about defining filterPLaceholdertext and filterLabelText?

frontend/shared/src/components/forms/inputs/MultiSelectDropdown.tsx
```diff
@@ -37,48 +51,99 @@ const MultiSelectDropdown = <T extends FieldValues, O extends Option>({
...
+  const draftSelectedValuesRef = React.useRef<HdsOption[]>(selectedValues);
+
+  React.useEffect(() => {
+    setDraftSelectedValues((previousOptions) =>
+      haveSameOptions(previousOptions, selectedValues)
+        ? previousOptions
+        : selectedValues
+    );
+    draftSelectedValuesRef.current = selectedValues;
+  }, [selectedValues]);
```
nikomakela: I don't understand the consept. From this I understood that HDS already does this? Why is it needed to be replicated? I don't say that this wouldn't be needed and the code itself is okay, but I just don't understand why it is needed at all.

frontend/shared/src/components/header/Header.tsx
```diff
@@ -62,6 +64,9 @@ const Header: React.FC<HeaderProps> = ({
   title,
   titleUrl,
 }) => {
+  const router = useRouter();
+  const { locale } = router;
```
nikomakela, line 68:
```suggestion
  const { locale } = useRouter();
```

frontend/shared/src/components/modal/Modal.tsx
```diff
@@ -59,26 +59,28 @@ const Modal: React.FC<ModalProps> = ({
...
               onClick={handleToggle}
               data-testid="modalCancel"
             >
               {cancelButtonLabel}
             </Button>
             <Button
-              theme="coat"
-              variant={variant}
+              theme={ButtonPresetTheme.Coat}
+              variant={variant as ButtonVariant}
```
nikomakela: In constructor the variant is typed as DialogVariant. Should it be ButtonVariant aswell? Do the variant values match between these types?

nikomakela (commented):
Some findings. I did go through both the services; benefit and kesaseteli, but mainly concentrated on Kesäsetelit's code quality.

frontend/shared/src/components/forms/inputs/MultiSelectDropdown.tsx
ToivoMattilaHelsinki: Dropdown functionality broke with the HDS upgrade and this change made the page work again.

Ideally in the future the whole component can be removed if HDS natively supports the required functionality.
For now that is out of scope for this PR

frontend/kesaseteli/youth/src/__tests__/utils/components/get-index-page-api.ts
jorilindell: HDS Select doesn’t use the real disabled attribute — it only sets aria-disabled.
Because of that, the browser doesn’t treat it as disabled, and toBeDisabled() won’t match. That matcher only works when the actual disabled property is present.

So here we need to check aria-disabled="true" instead.

frontend/kesaseteli/youth/src/__tests__/utils/components/get-index-page-api.ts
jorilindell: I totally agree. I changed this to:
```
const findSelectedSchoolCombobox = async (): Promise<HTMLElement> =>
    screen.getByRole('button', {
      name: regexp(translations.youthApplication.form.selectedSchool),
    });
```

frontend/kesaseteli/youth/src/__tests__/utils/components/get-additional-info-page-api.ts
jorilindell: I changed this to:
```
const getReasonsDropdown = async (): Promise<HTMLElement> =>
    screen.getByRole('combobox', {
      name: regexp(translations.additionalInfo.form.reasons),
    });
```

frontend/kesaseteli/youth/src/__tests__/utils/components/get-index-page-api.ts
jorilindell: Actually, in this test we only need to verify that the input has no validation errors.
I updated the assertion to:
`await waitFor(() => expect(input).toBeValid());`

frontend/kesaseteli/youth/src/__tests__/index.base.test.tsx
jorilindell: Changed as suggested. I’m not sure why this was changed earlier, since the previous version already matched the purpose of the test.

frontend/kesaseteli/shared/src/hooks/useCurrentYearSummerVoucherConfig.tsx
jorilindell: There is a linting rule import/prefer-default-export, which is why this change was made. The rule is disabled for .ts files in
https://github.com/City-of-Helsinki/yjdh/blob/main/frontend/.eslintrc.jsx.js  
but since this is a .tsx file, those overrides don’t apply.

I disabled the rule for this file in this commit:
https://github.com/City-of-Helsinki/yjdh/pull/4026/changes/85e15013be79cb4e5593a0053ef99182487842ab

frontend/kesaseteli/youth/src/components/youth-form/TargetGroupSelection.tsx
jorilindell: There is a linting rule import/prefer-default-export, which is why this change was made. The rule is disabled for .ts files in
https://github.com/City-of-Helsinki/yjdh/blob/main/frontend/.eslintrc.jsx.js  
but since this is a .tsx file, those overrides don’t apply.

I disabled the rule for this file in this commit:
https://github.com/City-of-Helsinki/yjdh/pull/4026/changes/85e15013be79cb4e5593a0053ef99182487842ab

frontend/kesaseteli/youth/src/components/youth-form/YouthApplication.tsx
jorilindell: There is a linting rule import/prefer-default-export, which is why this change was made. The rule is disabled for .ts files in
https://github.com/City-of-Helsinki/yjdh/blob/main/frontend/.eslintrc.jsx.js  
but since this is a .tsx file, those overrides don’t apply.

I disabled the rule for this file in this commit:
https://github.com/City-of-Helsinki/yjdh/pull/4026/changes/85e15013be79cb4e5593a0053ef99182487842ab

frontend/shared/src/components/forms/inputs/Dropdown.tsx
jorilindell: I moved the logic to generate dropdown texts to a separate useMemo to make it easier to read:
```
const texts = useMemo(() => {
    const baseTexts: Partial<SelectTexts> = {
      label: label || '',
      placeholder,
    };
    if (filterLabel) {
      baseTexts.filterLabel = filterLabel;
    }
    if (filterPlaceholder) {
      baseTexts.filterPlaceholder = filterPlaceholder;
    }
    return baseTexts;
  }, [label, placeholder, filterLabel, filterPlaceholder]);
```

frontend/shared/src/components/attachments/UploadAttachment.tsx
jorilindell: The reason we originally had iconEnd={null} was a TypeScript error when variant was undefined. HDS sets a default variant inside the Button component, but since we don’t set a default value ourselves, we need to cast variant to ButtonVariant:
`variant={variant as ButtonVariant}`

frontend/kesaseteli/employer/browser-tests/actions/application.actions.ts
jorilindell: The reason I removed expectedPreFill from the function call is that it isn’t used anywhere inside the function. I added an ESLint ignore comment instead to avoid the unused‑variable warning.

If the pre‑fill logic is needed, it would be good to implement it properly, but that’s outside the scope of this PR.

frontend/shared/src/components/forms/inputs/__tests__/MultiSelectDropdown.test.tsx
jorilindell: The recommended approach in the Testing Library docs is to use userEvent.setup(). Could you add the reasoning for preferring the global userEvent here?

For reference, the docs explain that the direct APIs (like userEvent.click) internally call userEvent.setup() under the hood, which means they still create a new instance each time:

- https://testing-library.com/docs/user-event/intro 
- https://testing-library.com/docs/user-event/setup/

frontend/shared/src/components/forms/inputs/__tests__/MultiSelectDropdown.test.tsx
jorilindell: The recommended approach in the Testing Library docs is to use userEvent.setup(). Could you add the reasoning for preferring the global userEvent here?

For reference, the docs explain that the direct APIs (like userEvent.click) internally call userEvent.setup() under the hood, which means they still create a new instance each time:

- https://testing-library.com/docs/user-event/intro 
- https://testing-library.com/docs/user-event/setup/

nikomakela: Okay, sorry, didn't know that it's the recommended way to use it after v14+

> Note that, while directly invoking APIs such as userEvent.click() (which will trigger setup internally) is [still supported in v14](https://testing-library.com/docs/user-event/setup#direct-apis), this option exists to ease the migration from v13 to v14, and for simple tests. We recommend using the methods on the instances returned by userEvent.setup().
- https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

There is an (unanswered) discussion started about this matter in Github too: https://github.com/testing-library/user-event/discussions/1158.

**My reasoning was that it's simpler, cleaner and consistent with other tests that are written in same project, if a global import would be used instead.**

frontend/shared/src/components/modal/Modal.tsx
```diff
@@ -59,26 +65,32 @@ const Modal: React.FC<ModalProps> = ({
...
             </Button>
             <Button
-              theme="coat"
-              variant={variant}
+              theme={ButtonPresetTheme.Coat}
+              variant={
+                variant === 'danger'
+                  ? ButtonVariant.Danger
+                  : ButtonVariant.Primary
+              }
```
nikomakela: I guess this is okay now when variant is typed as DialogVariant which only has these 2 variant options that are introduced here. It just raises doubts. :) 

Since HDS declares these:

```ts
export type DialogVariant = 'primary' | 'danger';

export declare enum ButtonVariant {
    Primary = "primary",
    Secondary = "secondary",
    Supplementary = "supplementary",
    Success = "success",
    Danger = "danger",
    Clear = "clear"
}
```

How about typing like this in Modal constructor:

```ts
  variant?: Extract<ButtonVariant, DialogVariant>;
```

then it would be okay to have 

```ts
    <Dialog
      id={id}
      aria-labelledby={titleId}
      isOpen={isOpen}
      className={className}
      close={handleToggle}
      closeButtonLabelText={closeButtonLabelText}
      scrollable={scrollable}
      variant={variant}
      theme={theme}
    >
```

and

```ts
            <Button
              theme={ButtonPresetTheme.Coat}
              variant={variant}
              onClick={onAccept}
              disabled={actionDisabled}
              data-testid="modalSubmit"
              iconStart={submitButtonIcon}
            >
```

because 

```ts
// This results in: ButtonVariant.Primary | ButtonVariant.Danger
// (which evaluate to "primary" | "danger")
type CommonVariant = Extract<ButtonVariant, DialogVariant>;
```

Trivial TS gymnastic, but would become a bit cleaner like this.

frontend/shared/src/components/modal/Modal.tsx
jorilindell: I added logic to map the dialog variant to a ButtonVariant so we don’t need a type cast.
```
variant={
  variant === 'danger'
    ? ButtonVariant.Danger
    : ButtonVariant.Primary
}
```

nikomakela: https://github.com/City-of-Helsinki/yjdh/pull/4026#discussion_r3232546000

frontend/shared/src/components/modal/Modal.tsx
nikomakela: Earlier discussion about this same code https://github.com/City-of-Helsinki/yjdh/pull/4026#discussion_r3232609896, which is now hidden and lost in endless flood of automatic CI comments.

jorilindell: I changed the type to
`variant?: Extract<ButtonVariant, DialogVariant>  `
in this commit:
https://github.com/City-of-Helsinki/yjdh/pull/4026/changes/5e800d7a26ec83da431c025dc4e53b93f47f2e28

This forces the modal to use the ButtonVariant enum, which seems fine for our use case. I kept this change in a separate commit to make the reasoning clear.

nikomakela (approved):
I read through all the code (Benefit and Kesäseteli), but didn't test either. I saw the screenshots of changes and I'm fine with them. On behalf of the summer voucher, I can accept the changes already. It will be tested in dev-environment anyway.

JJ-gitti (approved):
UI looks now ok for Helsinki-benefit.

PR #4031, opened 2026-04-27
fix(ks,employer): prevent infinite loop on no-organisation page

nikomakela:
YJDH-840.

If I have logged in with ADFS, which means that I do not represent any company, I should be taken to no-organisation -page. However, an infinite loop takes in action instead and this commit tries to fix that.

### DIFF
--- a/frontend/kesaseteli/employer/src/hocs/withOrganisation.tsx
+++ b/frontend/kesaseteli/employer/src/hocs/withOrganisation.tsx
@@ -29,8 +29,12 @@ const withOrganisation = <P extends JSX.IntrinsicAttributes>(
       isError(err) &&
       (err.message.includes('403') || err.message.includes('Forbidden'));
 
+    const isNotFound = (err: unknown): boolean =>
+      isError(err) &&
+      (err.message.includes('404') || err.message.includes('Not Found'));
+
     const isNoOrganisation = isCompanySuccess && !company?.name;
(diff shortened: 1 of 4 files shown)


frontend/kesaseteli/employer/src/hocs/withOrganisation.tsx
```diff
@@ -27,7 +27,10 @@ const withOrganisation = <P extends JSX.IntrinsicAttributes>(
 
     const isForbidden = (err: unknown): boolean =>
       isError(err) &&
-      (err.message.includes('403') || err.message.includes('Forbidden'));
+      (err.message.includes('403') ||
+        err.message.includes('404') ||
+        err.message.includes('Forbidden') ||
+        err.message.includes('Not Found'));
```
karisal-anders: How about simply moving the isForbidden function's test directly to the only place where it is used i.e. to isAuthError, then it is evident what it does? Just so it's evident that 404 is also considered forbidden. Or renaming the function to something else, e.g. isForbiddenOrNotFound? Not a big deal, suggesting these to add slight clarity.

karisal-anders (approved):
Left a comment, please take a look at it.

frontend/kesaseteli/employer/src/hocs/withOrganisation.tsx
nikomakela: splitted to isForbidden and isNotFound.

PR #4034, opened 2026-04-28
fix(benefit): fixes to condition loading broken by nextjs update

ext-hjasplund:
## Description :sparkles:

Fix loading conditions. Move pdfs worker copying to module level.

## Motivation and Context

Loading PDF conditions on screen was broken by latest
Next JS update.

[HL-1783](https://helsinkisolutionoffice.atlassian.net/browse/HL-1783)

[HL-1783]: https://helsinkisolutionoffice.atlassian.net/browse/HL-1783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -19,6 +19,21 @@ if (process.cwd().indexOf('/app/') === 0) {
   appName = process.cwd().split('/frontend/').pop();
 }
 
+// Copy pdfjs worker to public/ at config-load time so Turbopack can serve it as a static file.
+// This must be at module level — the webpack() callback is never called by Turbopack.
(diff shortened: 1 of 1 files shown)


karisal-anders (approved):
Approval from Kesäseteli side.

As a sidenote:
It doesn't seem like environment variables are being used by the portion of code you moved out from inside the `nextConfig` function, if there would be, then that would break the usage of environment variables set from .env-files in local development. So, looks ok in that regard.

mikkojamG (approved):
Seems ok.

PR #4035, opened 2026-04-28
fix(ks): handle cancelled login and suomi-fi -logout next-path

nikomakela:
YJDH-614.

1. When a login process is cancelled, user was taken to authentication error / access denied page (API url: /saml2/acs).

2. When a logout from employer UI was called, user was taken to API root, when they should have been taken to employer UI login page.

### DIFF
--- a/backend/kesaseteli/kesaseteli/urls.py
+++ b/backend/kesaseteli/kesaseteli/urls.py
@@ -12,6 +12,7 @@ from applications.views import (
 from common.views import healthz, readiness
 from companies.api.v1.views import GetCompanyView
 from shared.suomi_fi.views import (
+    HelsinkiSaml2LogoutView,
     SuomiFiAssertionConsumerServiceView,
     SuomiFiMetadataView,
 )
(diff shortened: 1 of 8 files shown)


backend/kesaseteli/kesaseteli/urls.py
```diff
@@ -73,6 +74,13 @@
             name="suomifi_saml2_metadata",
         )
     )
+    urlpatterns.append(
+        path(
+            "saml2/logout/",
+            HelsinkiSaml2LogoutView.as_view(),
+            name="saml2_logout",
```
karisal-anders, line 81: For conformity, I think it'd be good to either use the `suomifi_` prefix here as above for acs and metadata routes, or remove the `suomifi_` prefix from all three.

I was thinking about what's happening here as djangosaml2 has the urls also and they're appended right after these three. Here's djangosaml2's appended urlpatterns:
```python
urlpatterns = [
    path("login/", views.LoginView.as_view(), name="saml2_login"),
    path("acs/", views.AssertionConsumerServiceView.as_view(), name="saml2_acs"),
    path("logout/", views.LogoutInitView.as_view(), name="saml2_logout"),
    path("ls/", views.LogoutView.as_view(), name="saml2_ls"),
    path("ls/post/", views.LogoutView.as_view(), name="saml2_ls_post"),
    path("metadata/", views.MetadataView.as_view(), name="saml2_metadata"),
]
```

So `saml2_logout` name and path are exactly the same, for acs and metadata names are different, but paths are identical.

nikomakela: It's so in purpose, so it would kind of override those endpoints.

nikomakela: ahh, sorry, I didn't find out that there was already this 

```
if settings.NEXT_PUBLIC_ENABLE_SUOMIFI:
    urlpatterns.append(
        path(
            "saml2/acs/",
            SuomiFiAssertionConsumerServiceView.as_view(),
            name="suomifi_saml2_acs",
        )
    )
```

I will remove the prefixes.

karisal-anders (approved):

mikkojamG (approved):

charn (approved):

PR #4041, opened 2026-04-30
ci(ks): execute test workflows when PR to main is created

nikomakela:
YJDH-877.

The unit tests, linters, code formatters, etc. that a global ci-node workflow job should run, were not executed in Kesäseteli frontend context when a PR was created, but only when a merge to main was done. That's too late, since we should get the feedback already before merging the PR.

### DIFF
--- a/.github/workflows/ks-empl-frontend-tests.yml
+++ b/.github/workflows/ks-empl-frontend-tests.yml
@@ -6,6 +6,12 @@ on:
   pull_request:
     paths:
       - ".github/workflows/ks-empl-frontend-tests.yml"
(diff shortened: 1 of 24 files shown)


charn (approved):

karisal-anders (approved):
frontend/next.config.js could be added as it affects all the frontends

nikomakela:
I will fix the issues that CI reported. Biggest one is the `canvas` issue. It seems only Benefit apps are really using it and it comes as a dependency to Kesaseteli from shared folder. I will try if I can just mock it in jest-config.

PR #4044, opened 2026-05-05
fix(ks,shared): suomi-fi logout relay state handling for next-url

nikomakela:
YJDH-609.

Fix Suomi.fi logout redirection by stashing 'next' URL in session

Resolved an issue where the SAML Single Logout (SLO) flow would incorrectly redirect users back to the API root (/) instead of the requested 'next' URL.

Root Cause:
The underlying djangosaml2/pysaml2 libraries use the SAML 'RelayState' parameter to track request state. During the logout flow, this parameter is populated with a generated request ID (e.g. 'id-slRc7Ma...') rather than the destination URL. When the user returns from the IdP, djangosaml2's strict URL validation rejects this ID string as a literal URL, causing a fallback to the default LOGOUT_REDIRECT_URL.

Changes:
- HelsinkiSaml2LogoutView: Updated to capture the 'next' parameter and stash it in the user's session before initiating the SAML request to the IdP.
- HelsinkiSaml2LogoutServiceView: Implemented a new callback view that intercepts the final IdP response. It restores the 'next' URL from the session to override the library's default fallback behavior.
- urls.py: Explicitly registered the saml2_ls and saml2_ls_post endpoints to ensure our custom service view shadows the default djangosaml2 implementation.
- Documentation: Added detailed docstrings explaining the "Entry Door" (Initiator) and "Return Door" (Callback) architecture.

Testing:
- Added unit tests to verify session stashing and recovery logic.
- Implemented an integration test replicating the full two-step logout flow.
- Verified that open redirect protection (is_safe_redirect_url) is maintained.

### DIFF
--- a/backend/shared/shared/suomi_fi/tests/test_views.py
+++ b/backend/shared/shared/suomi_fi/tests/test_views.py
@@ -2,10 +2,11 @@ from unittest import mock
 
 import pytest
 from django.http import HttpResponseRedirect
-from django.test import RequestFactory
+from django.test import RequestFactory, override_settings
 
 from shared.suomi_fi.views import (
     RELAY_STATE_PARAM,
+    HelsinkiSaml2LogoutServiceView,
     HelsinkiSaml2LogoutView,
     SuomiFiAssertionConsumerServiceView,
 )
@@ -45,6 +46,7 @@ class TestSuomiFiViews:
         factory = RequestFactory()
         next_url = "/dashboard"
         request = factory.get("/saml2/logout/", {"next": next_url})
+        request.session = {}
 
         view = HelsinkiSaml2LogoutView()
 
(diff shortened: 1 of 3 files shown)


backend/shared/shared/suomi_fi/views.py
```diff
@@ -94,6 +108,51 @@
...
+        initiation. Here, we pop it back out, wait for djangosaml2 to predictably
+        fail and yield the fallback URL, and then replace it with our desired `next`
+        destination (provided it passes security checks).
+        """
+        next_path = request.session.pop("saml2_logout_next_path", None)
+        if next_path and isinstance(response, HttpResponseRedirect):
+            fallback_url = resolve_url(getattr(settings, "LOGOUT_REDIRECT_URL", "/"))
+            if response.url == fallback_url:
+                if is_safe_redirect_url(request, next_path):
+                    return HttpResponseRedirect(next_path)
```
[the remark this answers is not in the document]
nikomakela: You are suggesting use of `url_has_allowed_host_and_scheme`which is exactly what the `is_safe_redirect_url` is doing.

```python
def is_safe_redirect_url(
    request: HttpRequest,
    url: str,
    allowed_hosts: Union[list, str, None] = None,
    require_https: Union[bool, str, None] = None,
) -> bool:
    """
    Check if the given URL is safe for redirecting.
    - If `allowed_hosts` is a string, it's treated as a Django setting name.
    - If `require_https` is a string, it's treated as a Django setting name.
    """
    if not url:
        return False

    # Fetch allowed_hosts from settings if needed
    if isinstance(allowed_hosts, str):
        allowed_hosts = getattr(settings, allowed_hosts, [])
    elif allowed_hosts is None:
        allowed_hosts = getattr(settings, "OIDC_REDIRECT_ALLOWED_HOSTS", [])

    # Fetch require_https from settings if needed
    if isinstance(require_https, str):
        require_https = getattr(settings, require_https, request.is_secure())
    elif require_https is None:
        require_https = getattr(
            settings, "OIDC_REDIRECT_REQUIRE_HTTPS", request.is_secure()
        )

    all_allowed_hosts = list(allowed_hosts)
    all_allowed_hosts.append(request.get_host())

    return url_has_allowed_host_and_scheme(
        url=url,
        allowed_hosts=all_allowed_hosts,
        require_https=require_https,
    )
```

backend/shared/shared/suomi_fi/tests/test_views.py
```diff
@@ -90,3 +92,110 @@ def test_handle_unsupported_slo_exception_safe_check(self):
...
+        request = factory.get("/saml2/ls/?SAMLResponse=foo&RelayState=bar")
+        request.session = {"saml2_logout_next_path": next_url}
+
+        view = HelsinkiSaml2LogoutServiceView()
+
+        fallback_response = HttpResponseRedirect("/")
+        with mock.patch(
+            "djangosaml2.views.LogoutView.get", return_value=fallback_response
+        ):
+            with mock.patch("shared.suomi_fi.views.getattr", return_value="/"):
```
karisal-anders: If you're trying to test/mock this: `fallback_url = resolve_url(getattr(settings, "LOGOUT_REDIRECT_URL", "/"))` in `HelsinkiSaml2LogoutServiceView._handle_response` then using `@override_settings("LOGOUT_REDIRECT_URL", "/")` around this test case would be clearer IMHO because it is specific to `LOGOUT_REDIRECT_URL` instead of generally overriding the `getattr` call. Or are you doing something else here?

backend/shared/shared/suomi_fi/tests/test_views.py
```diff
@@ -90,3 +92,110 @@ def test_handle_unsupported_slo_exception_safe_check(self):
...
```
karisal-anders: Same comment here as https://github.com/City-of-Helsinki/yjdh/pull/4044/changes#r3194977180

backend/shared/shared/suomi_fi/tests/test_views.py
```diff
@@ -90,3 +92,110 @@ def test_handle_unsupported_slo_exception_safe_check(self):
...
+        )
+        # We simulate the handoff by passing the session object from step 1
+        request_ls.session = request_init.session
+
+        view_ls = HelsinkiSaml2LogoutServiceView()
+        fallback_response = HttpResponseRedirect("/")
+        with mock.patch(
+            "djangosaml2.views.LogoutView.get", return_value=fallback_response
+        ):
+            with mock.patch("shared.suomi_fi.views.getattr", return_value="/"):
```
karisal-anders: Same comment here as in https://github.com/City-of-Helsinki/yjdh/pull/4044/changes#r3194977180

karisal-anders (approved):
I think the `@override_settings` instead of mocking `getattr` would be more transparent/clear to developers as to what is being actually overridden/mocked.

danipran (approved):

PR #4045, opened 2026-05-05
feat(ks,shared): use persistent Suomi.fi usernames

nikomakela:
YJDH-883.

# Issue description

The current implementation of the Employer UI uses the transient NameID from Suomi.fi as the Django username (SAML_USE_NAME_ID_AS_USERNAME = True). Because transient identifiers are session-specific, a new Django User object is created for every single login session.

Consequences:

- Duplicate User Objects: Every time an employer logs in, they are treated as a completely new user.
- Data Fragmentation: Drafts and user-linked data created in one session become inaccessible in subsequent sessions.
- Inefficiency: The system accumulates "ghost" user records that cannot be linked back to the actual individual.

# Sources & References

The following technical specifications and documentation were used to analyze the issue and design the solution:
- SAML 2.0 Core Standard: OASIS SAML v2.0 Standard (saml-core-2.0-os) for the definition of NameID formats (transient vs persistent).
- Suomi.fi Interface Description: Rajapintakuvaus - Suomi.fi-tunnistus regarding attribute-based identification and the use of the oid attribute.
- Library Documentation: Djangosaml2 Documentation for configuring attribute mapping and username handling.

Internal Analysis:
- SAML Metadata analysis of the "Kesäseteli" service.
- Database sampling of existing "Jirin" users.
- HAR (HTTP Archive) trace analysis of the dev-auth process.

# Proposed Solution

The fix involves switching the identification logic from the session-based NameID to a persistent attribute:

- Disable SAML_USE_NAME_ID_AS_USERNAME.
- Map nationalIdentificationNumber to the Django username via SAML_ATTRIBUTE_MAPPING.
- Security: Implement salting for the username field to ensure that sensitive data (SSN/hetu) is not stored in plain text while remaining persistent across sessions.

### DIFF
--- a/backend/shared/shared/suomi_fi/auth.py
+++ b/backend/shared/shared/suomi_fi/auth.py
@@ -1,15 +1,24 @@
 import hashlib
 from typing import Any
 
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 from djangosaml2.backends import Saml2Backend
 
 
 class SuomiFiSAML2AuthenticationBackend(Saml2Backend):
     def clean_user_main_attribute(self, main_attribute: Any) -> Any:
         """
-        Return hash of Suomi.fi SSO session identifier.
+        Return salted hash of Suomi.fi nationalIdentificationNumber.
 
-        SSO session identifier is over the normal 150 character username limit.
+        This prevents storing the raw national identification number as the username.
         """
-        digest = hashlib.sha3_512(main_attribute.encode()).hexdigest()
+        salt = settings.SOCIAL_SECURITY_NUMBER_HASH_KEY
+
+        if not salt or not salt.strip():
+            raise ImproperlyConfigured(
+                "Required SOCIAL_SECURITY_NUMBER_HASH_KEY is empty."
+            )
+
+        digest = hashlib.sha3_512(f"{salt}{main_attribute}".encode()).hexdigest()
         return super().clean_user_main_attribute(digest)
--- a/backend/shared/shared/suomi_fi/tests/test_auth.py
+++ b/backend/shared/shared/suomi_fi/tests/test_auth.py
@@ -1,0 +1,84 @@
+import hashlib
+import secrets
+
+import pytest
+from django.core.exceptions import ImproperlyConfigured
+
+from shared.suomi_fi.auth import SuomiFiSAML2AuthenticationBackend
+
+
+@pytest.mark.django_db
+def test_suomifi_auth_backend_hashes_main_attribute(settings):
+    settings.SOCIAL_SECURITY_NUMBER_HASH_KEY = "test_salt"
+
+    backend = SuomiFiSAML2AuthenticationBackend()
+    main_attribute = "010101-123A"
+
+    expected_hash = hashlib.sha3_512(f"test_salt{main_attribute}".encode()).hexdigest()
+
+    cleaned_attribute = backend.clean_user_main_attribute(main_attribute)
+
+    assert cleaned_attribute == expected_hash
+    assert len(cleaned_attribute) == 128  # sha3_512 hexdigest length
+
+
+@pytest.mark.django_db
+def test_suomifi_auth_backend_raises_with_empty_salt(settings):
+    settings.SOCIAL_SECURITY_NUMBER_HASH_KEY = ""
+
+    backend = SuomiFiSAML2AuthenticationBackend()
+    main_attribute = "010101-123A"
+
+    with pytest.raises(
+        ImproperlyConfigured, match="Required SOCIAL_SECURITY_NUMBER_HASH_KEY is empty."
+    ):
+        backend.clean_user_main_attribute(main_attribute)
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "salt",
+    [
+        "test-salt-123",
+        secrets.token_hex(16),
+        secrets.token_urlsafe(32),
+    ],
+)
+@pytest.mark.parametrize(
+    "ssn",
+    [
+        "010101-123A",
+        "120202-345B",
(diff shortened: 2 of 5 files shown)


backend/kesaseteli/applications/tests/test_auth.py
```diff
@@ -0,0 +1,46 @@
...
+from shared.suomi_fi.auth import SuomiFiSAML2AuthenticationBackend
+
+
+@pytest.mark.django_db
+def test_suomifi_auth_backend_authenticate_with_kesaseteli_settings():
+    """
+    Test that the Suomi.fi authentication correctly uses the nationalIdentificationNumber
+    as the username source and applies the salted hashing logic as configured in
+    Kesäseteli settings.
+    """
```
karisal-anders: How about testing a few variants for the salt so they work through `SOCIAL_SECURITY_NUMBER_HASH_KEY `?
```suggestion
@pytest.mark.parametrize(
    "salt",
    [
        "test-salt-123",
        secrets.token_hex(16),
        secrets.token_urlsafe(32),
    ],
)
def test_suomifi_auth_backend_authenticate_with_kesaseteli_settings(settings, salt):
    """
    Test that the Suomi.fi authentication correctly uses the nationalIdentificationNumber
    as the username source and applies the salted hashing logic as configured in
    Kesäseteli settings.
    """
    settings.SOCIAL_SECURITY_NUMBER_HASH_KEY = salt
```

Plus `import secrets` in the start and removing the `salt = getattr(settings, "SOCIAL_SECURITY_NUMBER_HASH_KEY", "")` line altogether as the `salt` is an input variable for the function.

karisal-anders: It looks like you've done similar tests in the shared backend part already. By not using `test_salt` salt value in both tests in shared backend you'd get coverage for different non-empty salt values in use, then this'd not be needed anymore.

backend/shared/shared/suomi_fi/tests/test_auth.py
```diff
@@ -0,0 +1,71 @@
...
+    expected_hash = hashlib.sha3_512(f"{main_attribute}".encode()).hexdigest()
+
+    cleaned_attribute = backend.clean_user_main_attribute(main_attribute)
+
+    assert cleaned_attribute == expected_hash
+
+
+@pytest.mark.django_db
+def test_suomifi_auth_backend_authenticate_uses_mapped_attribute(settings):
+    settings.SOCIAL_SECURITY_NUMBER_HASH_KEY = "test_salt"
```
karisal-anders: To test varying salt values, how about:
```suggestion
    settings.SOCIAL_SECURITY_NUMBER_HASH_KEY = "2nd_test_salt"
```

backend/shared/shared/suomi_fi/tests/test_auth.py
```diff
@@ -0,0 +1,71 @@
...
+        "givenName": ("first_name",),
+        "sn": ("last_name",),
+    }
+
+    user = backend.authenticate(
+        None, session_info=session_info, attribute_mapping=attribute_mapping
+    )
+
+    assert user is not None
+    expected_username = hashlib.sha3_512("test_salt010101-123A".encode()).hexdigest()
```
karisal-anders: To test varying salt values, how about:
```suggestion
    expected_username = hashlib.sha3_512("2nd_test_salt010101-123A".encode()).hexdigest()
```

backend/shared/shared/suomi_fi/auth.py
```diff
@@ -1,15 +1,17 @@
...
     def clean_user_main_attribute(self, main_attribute: Any) -> Any:
         """
-        Return hash of Suomi.fi SSO session identifier.
+        Return salted hash of Suomi.fi nationalIdentificationNumber.
 
-        SSO session identifier is over the normal 150 character username limit.
+        This prevents storing the raw national identification number as the username.
         """
-        digest = hashlib.sha3_512(main_attribute.encode()).hexdigest()
+        salt = getattr(settings, "SOCIAL_SECURITY_NUMBER_HASH_KEY", "")
```
karisal-anders: It's debatable whether handling empty salt here is even wanted, as then there the security would be very lax. We could in kesaseteli's settings.py or somewhere test that the SOCIAL_SECURITY_NUMBER_HASH_KEY is set properly and is not e.g. Falsy or whitespace only. Or this code could e.g. raise an exception if salt is not set:
```python
if not salt or not salt.strip():
    raise ValueError("Required SOCIAL_SECURITY_NUMBER_HASH_KEY is empty.")
```

Either way would make this safer IMHO.

karisal-anders (approved):
Left some comments, please take a look at them. I think it'd be better to not allow empty SOCIAL_SECURITY_NUMBER_HASH_KEY, but check that it is not falsy or whitespace only in e.g. kesaseteli's settings.py or in the suomi_fi/auth.py, but either way would make the salt use safer IMHO. If you want to discuss this, ping me on Slack.

backend/shared/shared/suomi_fi/auth.py
nikomakela: I added ImproperlyConfigured raising to `settings.py` if the `SOCIAL_SECURITY_NUMBER_HASH_KEY` (or `ENCRYPTION_KEY`) is not set.

backend/kesaseteli/applications/tests/test_auth.py
nikomakela: This was more like a test that a setting should be applied and a setting was used.

nikomakela: I used ImproperlyConfigured for shared package and in settings.py, so I deleted this whole test suite.

danipran (approved):

PR #4053, opened 2026-05-08
API should provide target group translations in every language

nikomakela:
YJDH-864.

* Updated translation keys (which were updated last time in year 2024)
* Provided new translations for target groups
* Summer voucher configuration endpoint and target groups endpoint are now reacting to language-header and providing the target groups in requested language.

### DIFF
--- a/backend/kesaseteli/applications/tests/test_localized_api.py
+++ b/backend/kesaseteli/applications/tests/test_localized_api.py
@@ -1,0 +1,72 @@
+import pytest
+from django.urls import reverse
+from rest_framework import status
+
+from applications.models import SummerVoucherConfiguration
+
+TARGET_GROUP_TEST_CASES = [
+    (
+        "fi",
+        "9. luokkalainen",
+        "9. luokkalaiset: 16-vuotiaat, asuinpaikan on oltava Helsinki.",
+    ),
+    (
+        "en",
+        "9th grader",
+        "9th graders: 16 years old, must live in Helsinki.",
+    ),
+    (
+        "sv",
+        "Elev i årskurs 9",
+        "9:e-klassister: 16 år gamla, måste bo i Helsingfors.",
+    ),
+]
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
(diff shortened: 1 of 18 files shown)


Riippi (approved):

karisal-anders (commented):
This is from `backend/kesaseteli/README.md` :
```
### Updating translations

In `backend/kesaseteli/`:
* Run `python manage.py makemessages --no-location -l fi -l sv -l en`
* Run `python manage.py compilemessages`
```

Using the makemessages without `--no-location` adds a lot of lines of which at least one will be out-of-sync as soon as any translation moves to another line for whatever reason (e.g. adding one line in a function containing a translation before the translation use line). I'd vote for not adding the file/line numbers to the translation files so they don't go out-of-sync so easily.

Checking this PR otherwise still, just thought I'd mention this first. If you disagree with the argument here, we can discuss live too, I'm up for a healthy discussion.

backend/kesaseteli/applications/tests/test_localized_api.py
```diff
@@ -0,0 +1,68 @@
...
+
+    url = reverse("summer-voucher-configuration")
+
+    response = api_client.get(url, HTTP_ACCEPT_LANGUAGE=language)
+    assert response.status_code == status.HTTP_200_OK
+    data = response.data[0]
+    target_group = data["target_groups"][0]
+    assert target_group["id"] == "primary_target_group"
+    assert target_group["name"] == expected_name
+    assert target_group["description"] == expected_description
```
karisal-anders, line 51: The response is all SummerVoucherConfigurations in descending year order, so the first one is the one with the largest year, which is because of `seed_default_configuration_for_tests` test setup in conftest.py year 2027.

Searching for the updated year 2026 in the response would make this better.

karisal-anders (approved):
The value of `useSummerVoucherConfigurationQuery.test.tsx` is a bit dubious IMHO, not a blocker, but just please take a look at whether there's something that could be done to make it more valuable/useful, or if it's not really useful, then you could also consider dropping it entirely—I leave this up to you to.

PR #4057, opened 2026-05-08
test: use example.org when generating emails

nikomakela:
YJDH-833.

The generated phone numbers and emails can relate to some real contact information. Since Kesäseteli is sending messages via email, at least the emails should be generated so, that they never match a real email.

Kesäseteli nor Benefit do not send SMS messages, so it's better to keep phone number faker as it is, so we get more phone numbers tested.

References:
1. Use safe email from factory boy: https://faker.readthedocs.io/en/master/providers/faker.providers.internet.html#faker.providers.internet.Provider.safe_email
2. Use exampleEmail from js-faker: https://fakerjs.dev/api/internet.html#exampleemail

### DIFF
--- a/backend/kesaseteli/common/tests/factories.py
+++ b/backend/kesaseteli/common/tests/factories.py
@@ -130,12 +130,12 @@ class EmployerApplicationFactory(
     street_address = factory.Faker("street_address")
     bank_account_number = factory.Faker("iban")
     contact_person_name = factory.Faker("name")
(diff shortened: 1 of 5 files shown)


karisal-anders (approved):
There are also `EMAIL_BACKEND="django.core.mail.backends.locmem.EmailBackend"` overrides in some of the tests for example to make sure the emails don't go to anyone.

voneiden (approved):

PR #4070, opened 2026-05-12
feat(ks): DVV compliance logging for Suomi.fi login, mandate and VTJ queries

nikomakela:
YJDH-838 YJDH-850 YJDH-851.

  - Adds DVV compliance logging for three event types required by Finnish law: Suomi.fi SAML2 login (`LOGIN`), Suomi.fi Valtuudet mandate queries (`MANDATE_QUERY`), and VTJ population information queries (`VTJ_QUERY`)
  - Uses `django-resilient-logger` for local DB storage and scheduled Elasticsearch shipping; all logging gated behind `ENABLE_AUTH_LOGGING` setting (default `False`)
  - `VTJClient` (shared library) is kept unmodified — VTJ logging is applied at the Kesäseteli call site only


backend/kesaseteli/requirements.txt
```diff
@@ -88,8 +90,12 @@ djangorestframework==3.16.1
     # via -r requirements.in
 djangosaml2==1.12.0
     # via yjdh-backend-shared
+elastic-transport==8.17.1
+    # via elasticsearch8
 elasticsearch==7.17.13
     # via -r requirements.in
+elasticsearch8==8.19.3
+    # via django-resilient-logger
```
nikomakela: Is elasticsearch 7 still needed? Can we proceed with v8 only? Should we try with v7 instead even though it's not supported by resilient-logger?

backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -353,6 +357,34 @@
...
+
+# Resilient logger — ships compliance events and django-auditlog entries to ES
+RESILIENT_LOGGER = {
+    "origin": env("AUDIT_LOG_ORIGIN"),
+    "environment": env("AUDIT_LOG_ENV"),
+    "submit_unsent_entries": True,
+    "clear_sent_entries": True,
+    "sources": [
+        {"class": "resilient_logger.sources.ResilientLogSource"},
+        {"class": "resilient_logger.sources.DjangoAuditLogSource"},
```
nikomakela, line 372: Note that also DjangoAuditLogSource is now added.

backend/kesaseteli/requirements.txt
karisal-anders: As Elasticsearch servers in use according to my testing are v8 it's slightly weird that elasticsearch v7 even works with them. Apparently they have some inter-major-version-Elasticsearch support in place (even though they document that minor versions inside a major version are compatible).

Given that servers are v8, we should not need v7, but v8. And even `elasticsearch` package is not used by anything else AFAIK but the old audit logging system which we don't use anymore—I mean we can drop it from Kesäseteli as Kesäseteli's production environment has sent all old audit log entries to Elasticsearch.

backend/kesaseteli/kesaseteli/settings.py
karisal-anders: Before actually sending the new audit logs, that are auditlogged to the django-auditlog's model, to Elasticsearch it's good to check what Elasticsearch index they're being sent to, and whether we want to send them there or not. Probably it'd be better to have a different index for the new auditlogs than the old auditlogs so they don't get mixed.

backend/kesaseteli/kesaseteli/tests/conftest.py
```diff
@@ -0,0 +1 @@
+from common.tests.conftest import *  # noqa
```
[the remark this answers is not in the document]
karisal-anders: This is how it's been done in many places in this codebase already. The point about polluting the namespace is real, but how to fix this properly? The suggested code change is bogus i.e. not good, actually worse than what is being done now (By worse I mean it does basically the same thing but with more lines and more code).

backend/kesaseteli/kesaseteli/auth_logging.py
```diff
@@ -0,0 +1,328 @@
...
+   company inside Kesäseteli. When a user logs in and their session is associated
+   with a company, Kesäseteli queries the Suomi.fi Valtuudet (eAuthorizations)
+   REST API to verify that the user holds a valid mandate for that company and to
+   retrieve the list of roles granted. This module logs both successful and failed
+   mandate queries as MANDATE_QUERY entries.
+
+3. **VTJ personal information queries**
+   When a handler (caseworker) reviews a youth application, Kesäseteli queries
+   the VTJ REST API to verify the applicant's identity and home municipality.
+   This module logs both successful and failed queries as VTJ_QUERY entries.
```
karisal-anders, line 32: This query is done in at least two places:
1. Automatically by the system itself when a youth submits their youth application to the system. This is done to check whether the youth's application can be automatically accepted or not.
2. The case written here already i.e. when a youth application was not automatically accepted and it went to manual approval/rejection by a handler and the handler opens the youth application for manual processing.

backend/kesaseteli/kesaseteli/auth_logging.py
```diff
@@ -0,0 +1,328 @@
...
+The audit trail must make the following facts recoverable after the fact:
+
+* **Who acted** – the representative (puolesta-asioija), identified by user ID.
+* **On whose behalf** – the principal company, identified by Y-tunnus (business ID).
+* **When** – timestamps of both the query sent to Suomi.fi Valtuudet and the
+  response received (handled automatically by ``django-resilient-logger``).
+* **What the result was** – the mandate roles returned and whether the query
+  succeeded or failed.
+
+See: https://kehittajille.suomi.fi/palvelut/valtuudet/tekninen-dokumentaatio/lokitusvaatimukset
```
karisal-anders, line 51: From https://kehittajille.suomi.fi/palvelut/valtuudet/tekninen-dokumentaatio/lokitusvaatimukset
> Yrityksen puolesta asiointi
> Pyyntösanoma (OrganizationRoles-kysely): 
...
> Tapahtumatietojen kannalta olennaisimmat lokitettavat tietueet pyyntösanomasta:
...
> requestId = pyynnön yksilöivä tunniste. Tämän tiedon perusteella on mahdollista löytää sama lokimerkintä DVV:n ja asiointipalvelun lokitiedoista ja verrata niitä  tarvittaessa. 

At least the `requestId` is not logged at the moment.

backend/kesaseteli/kesaseteli/auth_logging.py
```diff
@@ -0,0 +1,328 @@
...
+    minimum set of fields required by DVV for login audit trails: who logged in,
+    via which authentication backend, and from which IP address.
+
+    Args:
+        request: The current Django ``HttpRequest``. Used to read ``REMOTE_ADDR``.
+        user: The authenticated Django ``User`` instance. ``user.backend`` is
+            set by Django's auth framework immediately before the signal fires
+            and contains the dotted path of the backend that authenticated the
+            user (e.g. ``social_core.backends.saml.SAMLAuth`` for Suomi.fi
+            SAML2 logins).
```
karisal-anders: Wouldn't the used auth backend be `shared.suomi_fi.auth.SuomiFiSAML2AuthenticationBackend` because of how `NEXT_PUBLIC_ENABLE_SUOMIFI` handles `AUTHENTICATION_BACKENDS` in Kesäseteli's settings.py?
```suggestion
            user (e.g. ``shared.suomi_fi.auth.SuomiFiSAML2AuthenticationBackend``
            for Suomi.fi SAML2 logins).
```

backend/kesaseteli/kesaseteli/AUTH_LOGGING.md
```diff
@@ -0,0 +1,206 @@
...
+
+### LOGIN
+
+Written on every successful Suomi.fi SAML2 authentication.
+
+```json
+{
+  "event_type": "LOGIN",
+  "user_id": "<Django user PK>",
+  "auth_backend": "social_core.backends.saml.SAMLAuth",
```
karisal-anders: Wouldn't the used auth backend be shared.suomi_fi.auth.SuomiFiSAML2AuthenticationBackend because of how NEXT_PUBLIC_ENABLE_SUOMIFI handles AUTHENTICATION_BACKENDS in Kesäseteli's settings.py?
```suggestion
  "auth_backend": "shared.suomi_fi.auth.SuomiFiSAML2AuthenticationBackend",
```

backend/kesaseteli/README.md
```diff
@@ -190,11 +191,47 @@ When `NEXT_PUBLIC_MOCK_FLAG=true` (typically in local development):
...
+  VTJ JSON fields are stripped from `YouthApplication` entries to reduce
+  personal data volume in the log).
+
+`django-auditlog-extra` adds:
+
+- `AuditLogConfigurationHelper` — called in `KesaseteliProjectConfig.ready()`
+  to raise an error on startup if any model is neither explicitly included nor
+  excluded, preventing accidental gaps in coverage.
+- `AuditlogAdminViewAccessLogMixin` — logs admin view access for `User` and
+  `Group` admin classes in [`kesaseteli/admin.py`](kesaseteli/admin.py).
```
karisal-anders: `AuditlogAdminViewAccessLogMixin` is used in the following:
- CompanyAdmin
- SummerVoucherConfigurationAdmin
- SchoolAdmin
- EmailTemplateAdmin
- YouthApplicationAdmin
- YouthSummerVoucherAdmin
- EmployerApplicationAdmin
- EmployerSummerVoucherAdmin
- KesaseteliUserAdmin
- KesaseteliGroupAdmin

Whether it's useful to have all these listed here is debatable, e.g. just adding "...and others" or something like that could do the trick IMHO.

backend/kesaseteli/kesaseteli/auth_logging.py
```diff
@@ -0,0 +1,328 @@
...
+    type used.
+
+    ``VTJClient`` is a shared library and is intentionally unaware of this
+    logging. All logging is performed at the Kesäseteli call site so that the
+    shared client remains generic and reusable across services.
+
+    Args:
+        end_user: The identifier of the handler (caseworker) who triggered the
+            VTJ query. Derived from ``VTJClient.get_end_user(request)``: a UUID
+            string for Helsinki AD users or the Django user PK otherwise.
```
karisal-anders: In YouthApplicationViewSet.create `VTJService.fetch_vtj_json` is called like this:
```python
            # Fetch the VTJ JSON data and save it
            youth_application.encrypted_original_vtj_json = VTJService.fetch_vtj_json(
                youth_application, end_user=""
            )
```
So in this example the `end_user` is an empty string. In this case the user is "the system" so to speak as the system queries the VTJ endpoint for the data, not a caseworker.

backend/kesaseteli/kesaseteli/auth_logging.py
```diff
@@ -0,0 +1,328 @@
...
+
+@receiver(user_logged_in)
+def on_user_logged_in(sender, request, user, **kwargs):
+    """Django signal receiver — logs a LOGIN entry on every successful authentication.
+
+    Registered by importing this module in ``KesaseteliProjectConfig.ready()``
+    (``apps.py``). The ``@receiver`` decorator connects this function to
+    Django's built-in ``user_logged_in`` signal, which fires after any
+    successful call to ``django.contrib.auth.login()``, including logins
+    performed by social-auth (Suomi.fi SAML2).
```
karisal-anders: Not a big deal, but the Suomi.fi SAML2 that we use does not use social-auth.

backend/kesaseteli/companies/services.py
```diff
@@ -143,13 +147,16 @@ def get_or_create_company_using_organization_roles(request: HttpRequest) -> Comp
...
     except RequestException as e:
+        log_mandate_query_failure(request.user, e)
         LOGGER.error(
             f"Unable to fetch organization roles from eauthorizations API: {e}"
         )
         raise NotFound(
             detail="Unable to fetch organization roles from eauthorizations API"
         )
 
+    log_mandate_query(request.user, organization_roles)
```
karisal-anders: **Not necessarily a problem**, but thought I'd bring this up: this logs mandate queries also that do not query the mandate query endpoint, but fetch the data from cache in `request.session.get("organization_roles")`:
```
def get_organization_roles(request: HttpRequest) -> dict:
    """
    Docs (only in Finnish):

    Search for "Taulukossa 1
    "https://palveluhallinta.suomi.fi/fi/tuki/artikkelit/592d774503f6d100018db5dd
    """
    org_roles = request.session.get("organization_roles")

    if not org_roles:
        org_roles = request_organization_roles(request)

    return org_roles
```

backend/kesaseteli/applications/tests/test_applications_auth_logging.py
```diff
@@ -0,0 +1,127 @@
...
+    VTJ_TIMEOUT=30,
+    VTJ_PERSONAL_ID_QUERY_URL="https://example.com/vtj",
+)
+def test_fetch_vtj_json_logs_vtj_query():
+    """Successful VTJ query creates a VTJ_QUERY log entry."""
+    application = YouthApplicationFactory()
+    end_user = "test-handler-uuid"
+    vtj_response = {
+        "Henkilo": {"Henkilotunnus": {"value": application.social_security_number}}
+    }
```
karisal-anders, line 34: There's also `mock_vtj_person_id_query_found_content` for creating mocked VTJ response data, not a big deal, just noticed that the social security number is in different place than in the VTJ response data normally, where it'd be in `Henkilo.Henkilotunnus.#text` or `Hakuperusteet.Henkilotunnus.#text`.

As this test passes I take it the social security number in the VTJ response is not really used in this test, so it's not very important where in it it is.

backend/kesaseteli/kesaseteli/settings.py
nikomakela: I have made a service request about this https://helsinkisolutionoffice.atlassian.net/servicedesk/customer/portal/3/PSD-8105.

backend/kesaseteli/kesaseteli/auth_logging.py
nikomakela: I added requestId to all new loggers. I also started using signals.

backend/kesaseteli/kesaseteli/auth_logging.py
```diff
@@ -0,0 +1,353 @@
...
+
+    Args:
+        sender: The function that sent the signal.
+        request: The current Django ``HttpRequest``, used to extract the user.
+        request_id: The unique ID generated for the API query.
+        organization_roles: The dict returned by the API containing roles.
+    """
+    user = getattr(request, "user", None)
+    if not request_id:
+        return
```
karisal-anders: Would it be better to raise an exception if `request_id` is not set? The places that send the `suomifi_mandate_queried` signal currently always send non-empty `request_id` values so they'd be fine with it.

karisal-anders (approved):
Using signals is better for this functionality IMHO👍.

- Perhaps missing `request_id` could just raise an exception as there's probably no use case for it being missing (All the code paths that send the signals that use the `request_id` currently send a non-empty `request_id` AFAIK).

- The empty `end_user` could be changed to `system` (or some enumeration or constant somewhere for it) and that could be documented where ever the `end_user` info is being documented (in the docstrings etc). Not a huge deal, and not a blocker, but this would bring a bit of clarity still to the documentation and to the logged data.

Let's see what the situation with the Elasticsearch indexes (one or more, what to do with the old index, do we get a new index etc) is.

backend/kesaseteli/kesaseteli/auth_logging.py
nikomakela: added system user in these

PR #4073, opened 2026-05-13
Refresh Suomi.fi production metadata

ToivoMattilaHelsinki:
## Description :sparkles:

YJDH-891

Refresh the checked-in Suomi.fi production IdP metadata for Kesäseteli so it matches the new certificate publication announced for the May 2026 rollover.

The change updates the production metadata XML, keeps the README provenance note current, and adds regression coverage for the expected entity ID, signing cert validity windows, and SSO/SLO endpoints.

## Testing :alembic:

- Added a regression test to basically catch if the metadata would be deleted accidentally

## Screenshots :camera_flash:

Not relevant


nikomakela (approved):
Looks good to me! 👍

voneiden (approved):
New certs look correct to me :+1:

PR #4079, opened 2026-05-15
Remove old auditlog transfer task, model and admin views as redundant

nikomakela:
Depends on #4070. 
NOTE: Since CI-tasks are not ran for chained PRs, this PR will be first created as unchained.

  YJDH-838 YJDH-850 YJDH-851.

Remove (old) auditlog transfer task, model and admin views as redundant.

> - The old audit log database models (`AuditLogEntry`) and Elasticsearch log transfer tasks have been removed, as they are no longer needed by any backend project.
> - **Kesäseteli** has migrated to using `django-auditlog`.
> - **Benefit** is using a custom audit logging implementation but currently still relies on some components of this package, such as the `AuditLoggingModelViewSet` and the `log` utility.
>
> The ultimate goal is to remove this `shared.audit_log` package entirely. This will happen once Benefit fully migrates away from its dependencies on it. The schedule for Benefit's migration is currently undetermined.


karisal-anders (approved):
Approval from Kesäseteli side.

These can be removed from Kesäseteli side as unused now:
- `CLEAR_AUDIT_LOG_ENTRIES`
- `ENABLE_SEND_AUDIT_LOG`
- `"audit_log.auditlogentry"` from `AUDITLOG_EXCLUDE_TRACKING_MODELS` in auditlog_settings.py as the model does not exist anymore

The `shared.audit_log.audit_logging.log` function could be made clearer now that it requires resilient logger, as without it the `log` function does nothing. Or, instead of just removing the `else` clause for fallback to `AuditLogEntry` creation maybe an exception could be raised for clarity.

nicobav (approved):

PR #4086, opened 2026-05-19
YJDH-628 | Use SAS token safely with custom Azure blob storage class

karisal-anders:
### Use SAS token safely with custom Azure blob storage class

Using similar code as in:
- https://github.com/City-of-Helsinki/kukkuu/pull/593 in Kukkuu
- https://github.com/City-of-Helsinki/yjdh/pull/4059 in Benefit

The custom Azure blob storage class `AzureStorageWithoutQuerystringAuth` is used to avoid the risk of accidentally exposing the SAS token in the URL, which could happen with the default `AzureStorage` class from `django-storages[azure]`.

Refactoring by moving AzureStorageWithoutQuerystringAuth and its tests to shared backend code could be done in the future if wanted to, which would necessitate `django-storages[azure]` as a dependency on the shared backend.

**NOTE**:
- `DEFAULT_FILE_STORAGE`'s value can be changed to `common.storage.AzureStorageWithoutQuerystringAuth` in DevOps for clarity, currently the backend has been defensively coded to override the default SAS-token exposing AzureStorage even if wanted to be used, with the custom `AzureStorageWithoutQuerystringAuth` class.

## Related

[YJDH-628](https://helsinkisolutionoffice.atlassian.net/browse/YJDH-628)

## Testing :alembic:

Tested locally by setting the following environment variables in `.env.kesaseteli-backend`:
- `AZURE_ACCOUNT_NAME`  with the value of `AZURE_BLOB_STORAGE_NAME` from dev-environment
- `AZURE_BLOB_STORAGE_SAS_TOKEN` with the value from dev-environment's keyvault
- `AZURE_CONTAINER` with the value of `AZURE_BLOB_STORAGE_CONTAINER` from dev-environment
- `DEFAULT_FILE_STORAGE=storages.backends.azure_storage.AzureStorage`

Tried to add attachment to an employer's application using the employer UI, was able to add attachment.
Looked at the backend's data that the `post_attachment` endpoint handled,
when using the normal `AzureStorage` the `attachment_file` contained the SAS-token, with the
`AzureStorageWithoutQuerystringAuth` it did not.

**NOTE**:
- **DevOps** variables / **secrets** related to this **must be updated** for production environment **before** this can be **deployed** to **production**.

[YJDH-628]: https://helsinkisolutionoffice.atlassian.net/browse/YJDH-628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nikomakela (approved):
Looks good to me.

It's unknown to me what scope a SAS-token generated by Azure for Kesäseteli currently has. If I have understood properly, if the SAS-token is for a single blob, it would be okay to "leake" it in URL. However, if it's scope is whole blob storage / container, then it's a huge security concern. 

Since Kesäseteli is using storage only through API, all the storage request should be hidden, wouldn't they, so query string would not be shown to client? The new `AzureStorageWithoutQuerystringAuth` should anyway strip the token from URL, if it's ever leaked, which is a good thing.

NOTE: `AzureStorageWithoutQuerystringAuth` could be a shared class, but it would need changes to Benefit too. However, if we are afraid of changing things in Benefit, we will probably never refactor this.

PR #4089, opened 2026-05-20
feat(ks,api): add OpenAPI and ReDoc for kesaseteli API

ToivoMattilaHelsinki:
## Description :sparkles:

Adds an OpenAPI schema for the Kesäseteli API using drf-spectacular and exposes ReDoc documentation for the public API surface. The schema is served at `/openapi/` and ReDoc at `/api_docs/redoc/`.

This also annotates the relevant API views and adds contract/regression tests to keep the documented request and response shapes aligned with the implementation.

Refs: YJDH-782

- Adds drf-spectacular to the Kesäseteli backend and configures it as the schema generator.
- Exposes the generated OpenAPI schema at `/openapi/`.
- Exposes ReDoc at `/api_docs/redoc/`.
- Annotates the API views so the exported schema matches the supported endpoints and payloads.
- Adds schema and contract tests for the documented request/response shapes.
- Updates backend documentation to mention the new API docs endpoints.

## Issues :bug:

- No known issues.

## Testing :alembic:

- Added contract tests for:
  - youth application creation without SSN
  - employee data lookup for summer vouchers
- Added schema regression tests that verify:
  - `/openapi/` is reachable and exports an OpenAPI 3.1 schema
  - `/api_docs/redoc/` is reachable
  - unsupported CRUD operations are excluded from the exported schema
  - the schema export command validates successfully

## Screenshots :camera_flash:

- ReDoc page at `/api_docs/redoc/`

Screenshot 2026-05-20 155238

- OpenAPI schema output at `/openapi/`

Screenshot 2026-05-20 155322

- The schema is generated at runtime from backend annotations so there is no committed `openapi.yaml` artifact in the repository.

## Summary by CodeRabbit

* **New Features**
  * Employer Excel exports with selectable kinds/columns and a dedicated download page
  * Employee-data lookup endpoint
  * Youth application creation flow that supports creating without SSN

* **Documentation**
  * OpenAPI and interactive docs available at /openapi/, /api_docs/swagger/, /api_docs/redoc/
  * Excel download behavior and error messages documented in the app README

* **Tests**
  * Added API contract and regression tests; expanded export and security test coverage


backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -874,28 +1003,37 @@ def get_queryset(self):
...
+        },
+    )
     @action(
         methods=("POST",),
         detail=True,
         url_path="attachments",
         parser_classes=(MultiPartParser,),
     )
-    def post_attachment(self, request, *args, **kwargs):
+    def post_attachment(self, request: Request, *args: Any, **kwargs: Any) -> Response:
```
karisal-anders: I wouldn't add `Any` type hints as if there is no type hint, then it can be anything, and Any just says it can be anything, isn't that just redundancy? I don't see adding `Any` type hint bringing any value, but slightly perhaps removing value by adding verbosity. Do you see some value in adding `Any` type hints?

I don't see e.g. Zen of Python's "explicit is better than implicit" as enough of an argument to add Any type hints to Python code in general either.

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -335,6 +407,13 @@ def fetch_employee_data(self, request, *args, **kwargs) -> HttpResponse:
             status=status.HTTP_200_OK,
         )
```
karisal-anders: Now the data is duplicated, at least the data could be deduplicated i.e. using exactly the same data as input to both the serializer and the response. Or even possibly getting the data for the Response from the serializer.

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -417,6 +503,13 @@ def reject(self, request, *args, **kwargs) -> HttpResponse:
         else:
             return HttpResponse(status=status.HTTP_400_BAD_REQUEST)
 
+    @extend_schema(
+        responses={
+            302: OpenApiResponse(description="Redirect to application page"),
```
karisal-anders: I'm not sure how to succinctly describe it, but "Redirect to application page" case here has the following cases:
- Redirect to additional info page
- Redirect to already activated page
- Redirect to expired page
- Redirect to accepted page
- Redirect to activated page

And whether it's good to even try to include all this in the description, hmm, probably not. Just thinking whether there'd be some middle ground between abstraction "application page" and the five cases that it encompasses.

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -934,7 +1087,10 @@ def post_attachment(self, request, *args, **kwargs):
         detail=True,
         url_path="attachments/(?P<attachment_pk>[^/.]+)",
     )
-    def handle_attachment(self, request, attachment_pk, *args, **kwargs):
+    def handle_attachment(
+        self, request: Request, attachment_pk: str, *args: Any, **kwargs: Any
+    ) -> HttpResponse | Response:
```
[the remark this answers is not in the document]
karisal-anders: Looking at the function a fallback to method not allowed at the end outside of the if-elif-clauses related to the method could be a good fallback. Nothing else (like the `pass` above, it's not needed) needs to be added, just a return with method not allowed at the end of the `handle_attachment` function at the topmost indentation level.

backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -383,6 +384,23 @@
...
+    # regenerations and avoid drf-spectacular fallback names for repeated status
+    # fields.
+    "ENUM_NAME_OVERRIDES": {
+        "EmployerApplicationStatusEnum": "applications.enums.EmployerApplicationStatus",
+        "YouthApplicationStatusEnum": "applications.enums.YouthApplicationStatus",
+    },
+    "DESCRIPTION": "REST API for Kesäseteli application management",
+    "VERSION": APP_RELEASE or "0.0.1",
+    "OAS_VERSION": "3.1.0",
 }
```
karisal-anders, line 464: How about something like this?
```suggestion
    "OAS_VERSION": "3.1.0",
    "SCHEMA_PATH_PREFIX": "/v1/",  # Strip prefix for better endpoint grouping in redoc
}
```

Looked at https://drf-spectacular.readthedocs.io/en/latest/settings.html and tried out the `SCHEMA_PATH_PREFIX`. Adding the `SCHEMA_PATH_PREFIX` made the endpoint grouping better IMHO, now company has its own group, employerapplications has its own group etc:
Image

backend/kesaseteli/kesaseteli/urls.py
```diff
@@ -58,6 +62,12 @@
         ),
         name="logout",
     ),
+    path("openapi/", SpectacularAPIView.as_view(), name="schema"),
+    path(
+        "api_docs/redoc/",
+        SpectacularRedocView.as_view(url_name="schema"),
+        name="redoc",
+    ),
```
karisal-anders, line 82: Like discussed privately in Slack this addition was enough to make Swagger generated docs available:
```suggestion
    ),
    path(
        "api_docs/swagger/",
        SpectacularSwaggerView.as_view(url_name="schema"),
        name="swagger-ui",
    ),
```

Result locally:
Image

How about including both redoc and swagger so user can choose which one to use?

Enums looked ok otherwise in Swagger, only `LanguageC70Enum` was weirdly named:
- Image

Maybe handling it like some other enums in spectacular settings would make its name more normal?

backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -322,6 +379,18 @@ def fetch_employee_data(self, request, *args, **kwargs) -> HttpResponse:
...
+                "employee_birthdate": youth_application.birthdate,
+                "employee_phone_number": youth_application.phone_number,
+                "employee_home_city": youth_application.home_municipality,
+                "employee_postcode": youth_application.postcode,
+                "employee_school": youth_application.school,
+            }
+        )
+        response_serializer.is_valid(raise_exception=True)
         return Response(
             data={
```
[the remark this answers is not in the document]
karisal-anders: If `youth_application.birthdate` would be `None`, the application would be broken. It is not a valid case.

Youth application's birthdate is either from the user's permanent Finnish personal identity number (youth_application.social_security_number), if they have one, or from `non_vtj_birthdate` in the case they don't have one. There is no use case for a youth application without either of these information, nor should it ever happen using the current codebase. So, birthdate should always be available for a youth application, but if it for any reason is not, then it means the application data is broken.

backend/kesaseteli/applications/api/v1/views.py
ToivoMattilaHelsinki: Changed this to "Redirect to the relevant application status page" which is a bit more expressive.

The reader still has to know what "relevant" means (additional info, already activated, etc) but fitting all of that information to the description seems overkill and I don't have a better way to communicate the options for the status page

I'm marking this comment as resolved for now

backend/kesaseteli/applications/api/v1/views.py
ToivoMattilaHelsinki: The function now leverages the serializer for validating the data and the data from is not duplicated anymore

backend/kesaseteli/kesaseteli/urls.py
ToivoMattilaHelsinki: Swagger is now added

backend/kesaseteli/applications/api/v1/views.py
karisal-anders: That sounds like an ok compromise between succinctness and informational content

backend/kesaseteli/kesaseteli/settings.py
ToivoMattilaHelsinki: Added SCHEMA_PATH_PREFIX

backend/kesaseteli/applications/api/v1/openapi_serializers.py
```diff
@@ -0,0 +1,59 @@
+"""Schema-only serializers for Kesäseteli OpenAPI contract shapes."""
```
nikomakela: I understood that the original scope of this ticket was to just write / generate OpenAPI documentation, so I started thinking, why is there now new serializers. I do get it that these new serializers are an effective way to kind of type or structurise view's input and output types. If they are only for documentation, a file called `openapi_serializers.py` is a good one, but in that case, this file comment could explain it even more clearly, that the serializers are not in actual use, but are there only for documentation purpose. However, if they are really used in views, then the views could really be refactored so that they take more out of serializers: they could be pure DRF views instead of "custom mixins". Then they could be moved to `serializers.py`. At least some TODO-comments could be added there, if not doing the actual refactoring. It's not original scope of this PR to do refactoring, so it's your call to decide what to do.

Otherwise the code and all the changes looks good and are approvable.

backend/kesaseteli/applications/tests/test_excel_export.py
```diff
@@ -142,11 +158,9 @@ def test_excel_view_download_no_annual_applications(staff_client, columns):
     )
     EmployerApplicationFactory(status=EmployerApplicationStatus.DRAFT)
 
-    response = staff_client.get(
-        f"{excel_download_url()}?download=annual&columns={columns}"
-    )
+    response = staff_client.get(employer_excel_export_url("annual", columns))
 
-    assert response.status_code == 200
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
karisal-anders: It's a bit debatable whether 400 Bad Request is the most appropriate return status here as it is a valid possibility that there are not yet annual applications, when a new year has started and no applications have yet been processed for the year. But, not a huge deal either way, as this is only about the case when no applications are there.

backend/kesaseteli/applications/api/v1/serializers.py
```diff
@@ -953,6 +986,103 @@ class Meta:
...
+class EmployerSummerVoucherAttachmentUploadInputSerializer(serializers.Serializer):
+    """Request body (input) for uploading an attachment to an employer voucher."""
+
+    attachment_file = serializers.FileField()
+    attachment_type = serializers.ChoiceField(choices=AttachmentType.choices)
+
+
 class YouthApplicationExcelExportSerializer(serializers.ModelSerializer):
     application_year = serializers.SerializerMethodField()
     application_date = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
karisal-anders: It's definitely a broken youth application if `youth_application.birthdate` ever is `None`. All source paths that generate valid youth applications must generate a birthdate, if they don't, the youth application is invalid.

backend/kesaseteli/applications/templates/application_excel_download.html
```diff
@@ -56,18 +56,14 @@ <h2 class="mb-4">{% trans 'Työnantajien kesäsetelihakemukset' %}</h2>
...
 
                 <td>
-                    <form method="get" action="{% url 'excel-download' %}">
-                        <input type="hidden" name="download" value="unhandled">
-                        <input type="hidden" name="columns" value="reporting">
+                    <form method="get" action="{% url 'employer-excel-export' export_kind='unhandled' columns='reporting' %}">
                         <button type="submit" class="btn btn-primary btn-lg active w-100" aria-pressed="true">
                             {% trans 'Lataa ja merkitse käsitellyiksi' %}
                         </button>
                     </form>
```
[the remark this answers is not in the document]
nikomakela: I would say this would be a good change if the GET would be changed to a POST, but it's not in scope of this PR. I will create a ticket.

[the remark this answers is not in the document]
nikomakela: https://helsinkisolutionoffice.atlassian.net/browse/YJDH-903

nikomakela (approved):
The scope of this PR has already expanded significantly from the original and intended one, so I accept the changes. However, I made a follow-up PR https://github.com/City-of-Helsinki/yjdh/pull/4133 with small development ideas, which I think could be merged here if you think it makes sense.

I would like to get closer to true DRF class-based model views, where one view has 1 serializer. I quickly asked AI to provide an example of the changes and they are here https://github.com/City-of-Helsinki/yjdh/compare/YJDH-782-openapi-docs-improvements...YJDH-782-refactor-drf-views. However, I would not work on that branch further at this time (if ever).

PR #4102, opened 2026-05-27
feat(ks): move to single submit flow and implement friendly errors

nikomakela:
YJDH-847.

Refactors the employer application flow to prevent intermediate draft saves during employee data lookups, improves audit log reliability, and provides specific localized error toasts for validation issues.

Backend:
- Automatically creates an empty EmployerSummerVoucher inside EmployerApplicationSerializer.create, ensuring new applications are initialized with a voucher ID and enabling immediate attachment uploads.
- Updates YouthApplicationViewSet.fetch_employee_data to return specific error_code responses ("youth_application_not_accepted", "summer_voucher_already_used") on 400 Bad Request.
- Fixes test_applications_api.py to filter audit LogEntry queries by ContentType, avoiding assertions conflicts with the new auto-created voucher audit logs.

Frontend:
- Modifies fetchEmployment in useApplicationApi.ts to update the draft application locally instead of triggering a PUT request to the backend.
- Implements a DRY error handler in fetchEmployment that dynamically maps status and error codes to translation keys before calling showErrorToast.
- Simplifies handleGetEmployeeData in EmploymentForm.tsx to perform direct fetch lookups.
- Adds new translation keys for lookups under fi, sv, and en locales.

Tests:
- Deletes the obsolete updateApplication fallback test inside EmploymentForm.test.tsx.
- Adds useApplicationApi.test.ts to test success data-merging and all error toast scenarios in fetchEmployment.

### DIFF
--- a/backend/kesaseteli/applications/api/v1/views.py
+++ b/backend/kesaseteli/applications/api/v1/views.py
@@ -306,6 +306,24 @@ class YouthApplicationViewSet(ModelViewSet):
             },
         }
 
+        if not found_match:
+            # No match found for employee name in YouthApplication
+
+            # Log access because no match was found:
+            AuditAccessLogService.create_access_log_entry_with_no_related_object_instance(
+                actor=request.user,
+                actor_email=request.user.email,
+                content_type=ContentType.objects.get_for_model(YouthApplication),
+                additional_data=additional_data_for_access_audit_log,
+            )
+            youth_application_pk = youth_application.pk if youth_application else "None"
+            LOGGER.warning(
+                f"No match for employee name {employee_name} with voucher"
+                f" {voucher_number} in YouthApplication {youth_application_pk}"
+                f" - Cannot set employee data."
+            )
+            return Response(status=status.HTTP_404_NOT_FOUND)
+
         if (
             youth_application
             and youth_application.status != YouthApplicationStatus.ACCEPTED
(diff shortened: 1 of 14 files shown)


backend/kesaseteli/applications/api/v1/views.py
```diff
@@ -305,15 +307,53 @@ def fetch_employee_data(self, request, *args, **kwargs) -> HttpResponse:
...
+            if used_in_other.exists():
+                LOGGER.warning(
+                    f"Summer voucher {youth_summer_voucher.pk} already used in other"
+                    f" application - Cannot set employee data."
+                )
+                return Response(
+                    data={"error_code": "summer_voucher_already_used"},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+
```
ToivoMattilaHelsinki, line 356: if not found_match section above has audit logging
Should these 2 sections also have audit logging?

ToivoMattilaHelsinki: Alternatively remove the audit logging from the 404 path above

Audit logging should log what data a user has seen and had access to
None of these paths result in a user seeing and having access to PII data

karisal-anders: The audit logging of the 404 case was for keeping track of possible abuse. It was added because of the possibility of bruteforcing sequentially numbered youth summer voucher serial numbers + employees' surnames. There are still some sequentially numbered vouchers this year still, only come next year only randomly generated and very hard to guess voucher numbers are fully in use (Unless someone still tries to apply for compensation for this year's sequentially numbered voucher during early next year—can be possible).

Long story short: audit logging of the 404 case is valid and should IMHO be in place until randomly generated voucher numbers are the only ones actually used in practice, which is not true until perhaps spring 2027 (Not yet in January or February 2027 because of some possible latecoming employers seeking compensation during early 2027).

nikomakela: The purpose of the function that @karisal-anders  brought up is good and the concern is justified, but I agree with @ToivoMattilaHelsinki that logging to the audit log is questionable. Some kind of access log would be better for that purpose. I personally would like to see only information about what someone has actually seen in the audit log. What someone has tried to see or do is more the job of some kind of access log.

So I think Toivo's point is also good, and it's good that he brought it up, but I'm not necessarily going to remove the function at this point, when Kari wants to keep the function in place, and that would actually also be outside the original scope of this PR, because I haven't made any changes to it within this PR and it wasn't expected to be done before this review.

ToivoMattilaHelsinki (approved):

PR #4116, opened 2026-06-02
fix(benefit): apprenticeship decoupling and other fixes

ext-hjasplund:
## Description :sparkles:

* Apprenticeship was coupled with pay subsidy. Setting pay subsidy to null and remove coupling from UI.
* List of applications with second instalment in REQUESTED status was improperly formatted. Fixed formatting.
* Applications partially sent to TALPA were shown in HANDLING state. Fixed.
* Instalment amounts in handler decision texts contained decimals. Fixed to use integers only.
* Fixed ugly error message of company number of employees

## Contextx

[HL-1767](https://helsinkisolutionoffice.atlassian.net/browse/HL-1767)
[HL-1745](https://helsinkisolutionoffice.atlassian.net/browse/HL-1745)
[HL-1816](https://helsinkisolutionoffice.atlassian.net/browse/HL-1816)
[HL-1817](https://helsinkisolutionoffice.atlassian.net/browse/HL-1817)
[HL-1819](https://helsinkisolutionoffice.atlassian.net/browse/HL-1819)

## Testing :alembic:

TestCafe tests were modified for apprenticeship decoupling
Visual checking of integers and second instalment formatting.

[HL-1767]: https://helsinkisolutionoffice.atlassian.net/browse/HL-1767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HL-1745]: https://helsinkisolutionoffice.atlassian.net/browse/HL-1745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HL-1816]: https://helsinkisolutionoffice.atlassian.net/browse/HL-1816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HL-1819]: https://helsinkisolutionoffice.atlassian.net/browse/HL-1819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HL-1817]: https://helsinkisolutionoffice.atlassian.net/browse/HL-1817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/shared/browser-tests/page-models/PageComponent.ts
+++ b/frontend/shared/browser-tests/page-models/PageComponent.ts
@@ -44,13 +44,13 @@ abstract class PageComponent {
   );
 
   protected constructor(datatestId?: string) {
-    this.componentSelector = this.screen.findByTestId(
-      datatestId ?? MAIN_CONTENT_ID
+    this.componentSelector = Selector(
+      `[data-testid="${datatestId ?? MAIN_CONTENT_ID}"]`
     );
     this.component = this.within(this.componentSelector);
   }
 
-  loadingSpinners = this.screen.findAllByTestId('hidden-loading-indicator');
+  loadingSpinners = Selector('[data-testid="hidden-loading-indicator"]');
 
   public async isLoadingSpinnerNoMorePresent(timeout?: number): Promise<void> {
     return (
@@ -67,7 +67,7 @@ abstract class PageComponent {
    */
   public async isLoaded(timeout?: number): Promise<void> {
     await this.isLoadingSpinnerNoMorePresent(timeout);
-    return this.expect(this.componentSelector);
+    return this.expect(this.componentSelector, timeout);
   }
 
   /** expectation utility */
@@ -76,10 +76,12 @@ abstract class PageComponent {
     selector: Selector | SelectorPromise,
     timeout?: number
   ): Promise<void> {
+    const exists = await selector.exists;
+
     return (
       t
         // eslint-disable-next-line security/detect-non-literal-fs-filename
-        .expect(selector.exists)
+        .expect(exists)
         .ok(await getErrorMessage(t), { timeout })
     );
   }
@@ -88,10 +90,12 @@ abstract class PageComponent {
     selector: Selector | SelectorPromise,
     timeout?: number
   ): Promise<void> {
+    const exists = await selector.exists;
+
     return (
       t
         // eslint-disable-next-line security/detect-non-literal-fs-filename
-        .expect(selector.exists)
+        .expect(exists)
         .notOk(await getErrorMessage(t), { timeout })
     );
   }
--- a/backend/benefit/applications/api/v1/serializers/application.py
+++ b/backend/benefit/applications/api/v1/serializers/application.py
@@ -769,6 +769,11 @@ class BaseApplicationSerializer(DynamicFieldsModelSerializer):
         ):
             data["association_immediate_manager_check"] = None
 
+    def _normalize_pay_subsidy_granted(self, pay_subsidy_granted, data):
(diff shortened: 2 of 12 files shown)


frontend/shared/browser-tests/page-models/PageComponent.ts
```diff
@@ -44,13 +44,13 @@ abstract class PageComponent {
   );
 
   protected constructor(datatestId?: string) {
-    this.componentSelector = this.screen.findByTestId(
-      datatestId ?? MAIN_CONTENT_ID
+    this.componentSelector = Selector(
+      `[data-testid="${datatestId ?? MAIN_CONTENT_ID}"]`
     );
```
nikomakela, line 49: I did not really run or test different option for this test, but is this change really necessary? User-event library does not come with Testcafe, but if we have code where we have used it, I guess we at least have had it installed. In that case, is there a real reason why it was removed? If it's installed, it's better to also use it, because it tests the UI in a way that a user would really use it. Sure, user never really selects an input with some data-testid, so this is a bad example of that, but still. I just would like to know whether this change was *really* needed?

frontend/shared/browser-tests/page-models/PageComponent.ts
```diff
@@ -88,10 +90,12 @@ abstract class PageComponent {
...
+    const exists = await selector.exists;
+
     return (
       t
         // eslint-disable-next-line security/detect-non-literal-fs-filename
-        .expect(selector.exists)
+        .expect(exists)
         .notOk(await getErrorMessage(t), { timeout })
     );
   }
```
nikomakela, line 101: Shouldn't this be awaited (and let the testcafe handle the promise)

```suggestion
protected async expectNotPresent(
  selector: Selector | SelectorPromise,
  timeout?: number
): Promise<void> {
  // Fetch the error message beforehand if it's async
  const errorMessage = await getErrorMessage(t); 

  // Do NOT await the selector here. Let TestCafe handle the promise evaluation.
  await t
    .expect(selector.exists)
    .notOk(errorMessage, { timeout });
}
```

frontend/shared/browser-tests/page-models/PageComponent.ts
```diff
@@ -76,10 +76,12 @@ abstract class PageComponent {
...
+    const exists = await selector.exists;
+
     return (
       t
         // eslint-disable-next-line security/detect-non-literal-fs-filename
-        .expect(selector.exists)
+        .expect(exists)
         .ok(await getErrorMessage(t), { timeout })
     );
   }
```
nikomakela, line 87: I would say that this should be like this:
```suggestion
/** expectation utility */
  /* eslint-disable class-methods-use-this */
  protected async expect(
    selector: Selector | SelectorPromise,
    timeout?: number
  ): Promise<void> {
    // Resolve the error message first if getErrorMessage is asynchronous
    const errorMessage = await getErrorMessage(t);

    // Pass the actual property promise to .expect() so TestCafe can retry it
    await t
      .expect(selector.exists)
      .ok(errorMessage, { timeout });
  }
```

PR #4135, opened 2026-06-08
Add Matomo analytics and cookie consent to all Kesäseteli UIs

nikomakela:
YJDH-723.

Integrates Matomo web analytics across the youth, employer and handler frontends via the Helsinki Design System cookie consent mechanism. Users are shown a cookie consent banner on their first visit and can revisit their preferences at any time through a dedicated `/cookie-settings` page — **without needing to log in first**.

---

#### Analytics & cookie consent infrastructure (`kesaseteli-shared`)
- Implemented `useMatomo` hook that initialises Matomo tracking and subscribes to HDS cookie consent events to enable/disable tracking dynamically
- Extended `useCookieConsentParams` to support the Matomo consent cookie group
- Added `cookieConsentSettings` support for the `COOKIE_CONSENT_SITE_NAME` constant used across all apps
- Added unit tests for the `useMatomo` hook
- Introduced `constants/routes.ts` with shared `ROUTES.LOGIN`, `ROUTES.COOKIE_SETTINGS` and `ROUTES.NO_ORGANISATION` constants to replace scattered hardcoded strings

#### Youth UI
- Mounts `useMatomo` and renders the HDS `CookieConsent` banner in the app layout
- Cookie banner is hidden on the `/cookie-settings` page itself

#### Employer UI
- Mounts `useMatomo` and renders the `CookieConsent` banner in the app layout
- Adds `/cookie-settings` page accessible **without authentication** — `AuthProvider`, `Header` and `useUserQuery` all skip auth checks and suppress login redirects when on this route

#### Handler UI
- Mounts `useMatomo` and renders the `CookieConsent` banner in the app layout
- Adds `/cookie-settings` page (new `cookie-settings.tsx`) accessible **without authentication**
- Links to the cookie settings page from the `Footer`
- Adds Finnish translation key for the footer link

---

### Testing

- Unit tests added for `useMatomo` (shared), and for the unauthenticated `/cookie-settings` page in both the employer and handler apps
- Typechecks pass across all three apps (`employer`, `handler`, `youth`)

## Summary by CodeRabbit

* **New Features**
  * Added a Cookie Settings page, localized cookie-settings links in footers, and updated cookie banner to show when analytics is configured or when an env flag is set (hidden on the cookie settings page).

* **Chores**
  * Clarified Matomo analytics configuration and consent handling, improved analytics initialization and SPA page‑view tracking, added shared routing constants, and introduced new cookie consent defaults.

* **Tests**
  * Added/updated tests for cookie banner/settings, auth/user hooks, Matomo, and related utilities.


frontend/shared/src/utils/cookieConsentSettings.ts
```diff
@@ -322,6 +334,13 @@ export const getDefaultKesaseteliOptionalGroups = (
         expiration: ONE_HOUR_EXPIRATION,
         storageType: 1,
       },
+      {
```
ToivoMattilaHelsinki, line 337: Should this section also include mtm_consent?

We add these in useCookieConsentParams:
          ['rememberConsentGiven'],
          ['rememberCookieConsentGiven']

And inform the user about mtm_cookie_consent

Shouldn't we also inform the user about mtm_consent?

frontend/kesaseteli/handler/src/hooks/backend/useUserQuery.ts
```diff
@@ -17,10 +19,20 @@ const useUserQuery = <T = User>({
 }: UseQueryOptions<T> = {}): UseQueryResult<T> => {
   const isRouting = useIsRouting();
   const goToPage = useGoToPage();
+  const router = useRouter();
   return useQuery(BackendEndpoint.USER as QueryKey, {
     enabled: !!enabled && !isRouting,
     onError: useErrorHandler({
-      onServerError: () => goToPage('/login?error=true'),
+      onServerError: () => goToPage(`${ROUTES.LOGIN}?error=true`),
```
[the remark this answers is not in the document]
nikomakela: Isn't this handled now? I'm resolving this.

frontend/shared/src/utils/cookieConsentSettings.ts
nikomakela: Hmm.. yes, and no...

Image

Actually I don't think we should use them both at the same time, which then means that this section could be okay, but we should get rid of one of them.

> Choose ONLY mtm_consent if your legal team says: "We cannot track anything or touch a user's machine until they click Accept." (Standard GDPR implementation).

> Choose ONLY mtm_cookie_consent if your legal team says: "We can track basic page views anonymously under 'Legitimate Interest', but we can't drop tracking cookies without permission."

nikomakela: I self added a recommendation to use `mtm_consent` only in: https://github.com/City-of-Helsinki/yjdh/pull/4135/changes/e5a520b6fb297967f69fcae4a49055c7da263d2d

frontend/kesaseteli/employer/src/hooks/backend/__tests__/useUserQuery.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  describe('onAuthError callback', () => {
+    it.each([
+      [ROUTES.LOGIN, false],
+      [ROUTES.NO_ORGANISATION, false],
+      [ROUTES.COOKIE_SETTINGS, false],
+      ['/', true],
+      ['/dashboard', true],
+    ])(
+      'when route is %s, should call goToPage: %s',
```
karisal-anders, line 69: How about this? Not a big deal either way:
```suggestion
      'when route is %s, should redirect to login page: %s',
```

frontend/kesaseteli/shared/src/hooks/useMatomo.ts
```diff
@@ -16,8 +17,16 @@ export type UseMatomoProps = {
...
 
 /**
- * Use Matomo tracking
- * @returns True if Matomo is enabled and has URL & site ID, false otherwise.
+ * Custom hook to initialize Matomo analytics and track client-side page views.
+ *
+ * It registers two effects:
+ * 1. Initializer Effect: Runs once to load the Matomo analytics script into the DOM
+ *    if Matomo is fully configured (enabled, valid URL, and site ID).
+ * 2. Route Tracker Effect: Listens to Next.js route transitions (`router.asPath`)
```
karisal-anders: Is `router.asPath` accurate here? What about e.g. `router.events routerChangeComplete`?

karisal-anders (approved):
Looks ok to me in the big picture. Left some comments, please take a look at them.

Looked through the code, didn't test in action. Also ran through code review with Claude Opus 4.8 a couple of times at different times of PR development.

frontend/shared/src/hooks/useCookieConsentParams.ts
```diff
@@ -20,7 +20,7 @@ const onChange = (changeEvent: CookieConsentChangeEvent): void => {
     if (hasStatisticsConsent) {
       // Start Matomo only after statistics consent is granted.
       if (window._paq) {
-        window._paq.push(['setConsentGiven'], ['setCookieConsentGiven']);
+        window._paq.push(['setConsentGiven'], ['rememberConsentGiven']);
       }
     } else if (window._paq) {
       // Tell Matomo to forget consent when statistics consent is removed.
```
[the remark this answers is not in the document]
nikomakela: Gemini said it's okay ✅ :

```
The potential issue identified by the AI agent is invalid (not a real issue). The current implementation works correctly and does not lose the initial page view analytics data when a new user grants consent.

Verification of the Behavior
Matomo's Client-Side Queueing System: When Matomo's tracking snippet runs on a new user's first visit:

initMatomo() executes _paq.push(['requireConsent']); before calling _paq.push(['trackPageView']);.
When the Matomo JavaScript tracker (piwik.js) loads, it sees requireConsent and pauses all outgoing HTTP tracking requests.
Any tracking requests pushed afterward (like the initial trackPageView) are not discarded. Instead, they are held in a pending queue (buffer) in client-side memory.
Retroactive Processing on Consent:

When the user selects their cookie preferences and accepts statistics cookies, the onChange event in 

useCookieConsentParams.ts
 fires.
It pushes ['setConsentGiven'] and ['rememberConsentGiven'] to window._paq.
Upon receiving setConsentGiven, Matomo's tracker automatically changes the consent state and retroactively processes and dispatches all queued requests (including the initial trackPageView that was blocked during initialization).
Returning Users:

On subsequent visits, rememberConsentGiven will have set the Matomo consent cookie mtm_consent.
When initMatomo() executes, the tracker detects this cookie and immediately processes all tracking calls without queueing them.
Conclusion
No code changes are necessary. The initial page view event is temporarily queued in compliance with privacy consent requirements and successfully sent once the user grants consent.
```

However, Claude said it's a real bug ❗ 

```
The Claimed Scenario
trackPageView is pushed while requireConsent is active → suppressed → setConsentGiven is called later → but the original trackPageView is never replayed

This is the critical question: does Matomo retroactively process or discard queued trackPageView calls when setConsentGiven arrives?

What the Matomo Source Code and Docs Actually Say
The _paq queue is a simple JavaScript array. When the tracker JS loads, it processes the queue in order. The commands requireConsent → trackPageView are already in the queue as pre-init commands, processed synchronously in sequence.

Here is what Matomo's tracker does with trackPageView when requireConsent is active:

If the tracker has already loaded (consent given asynchronously later via banner interaction): trackPageView ran through the pre-init queue, evaluated against configConsentRequired=true, and aborted without queueing for later replay. The event is simply dropped.
setConsentGiven does not replay already-evaluated trackPageView calls. It only enables future tracking requests.
The Matomo official docs example pattern is explicit about this:

js
// Correct pattern for consent-required setups — call trackPageView AFTER setConsentGiven:
_paq.push(['requireConsent']);
// ... user grants consent ...
_paq.push(['setConsentGiven']);
_paq.push(['trackPageView']); // ← must come AFTER consent
Verdict: This is a real bug
For a new visitor who hasn't previously consented:

initMatomo() pushes requireConsent → trackPageView into _paq
The tracker loads, sees requireConsent, sets the consent flag, then encounters trackPageView — which is immediately discarded because consent is not yet given
The user interacts with the cookie banner and setConsentGiven is called
Matomo starts accepting new tracking calls, but the initial trackPageView is gone — no re-emission happens
The first page visit goes untracked
Note: Returning users who have mtm_consent cookie are fine — rememberConsentGiven sets consent state before trackPageView is evaluated.

Fix
In the onChange handler, after calling setConsentGiven, also emit a trackPageView for the current page:
```
Changes that Claude suggested:
image

After this, Gemini said that fix was valid

```
Upon re-evaluating the behavior of the Matomo tracker and the implemented fix, here is the corrected analysis:

The change at lines 25–28 of 

useCookieConsentParams.ts
 is absolutely correct and solves a real bug.

Why the Change is Correct
No Client-Side Request Buffering: When requireConsent is active, Matomo does not cache or buffer tracking requests (like the initial page view) to replay them later. Instead, the tracker simply discards any incoming tracking requests while consent is missing.
The Lost First Page View: Because the first trackPageView triggered during initialization is discarded, when the user subsequently provides consent via the cookie banner and setConsentGiven is called, Matomo is unlocked for future requests but has no memory of the initial page view.
How the Fix Solves It: By adding an explicit call to 

trackPageView
 immediately after setConsentGiven, we guarantee that the current page is tracked right as the user grants consent. Returning users remain unaffected as their existing consent cookie lets the initial page view pass through immediately upon load.
```

Riippi (approved):

PR #4140, opened 2026-06-09
feat(ks): restrict application cleanup to 5 years and add monthly scheduled job

nikomakela:
## Ticket
YJDH-652.

## Summary
This PR implements a 5-year retention restriction on the application cleanup process, adds a monthly scheduled job to trigger the cleanup, and updates the documentation.

### Management Command
- Modified `cleanup_applications` Django management command:
  - Enforced a default 5-year retention policy: only draft employer applications and expired youth applications older than 5 years are removed.
  - Added a new `--ignore-5-years-restriction` CLI flag to bypass the 5-year limit and purge matching applications of any age immediately (useful because youth applications expire much faster, e.g. within hours/days).

### Scheduled Job
- Created a new monthly django-extensions job `Job(MonthlyJob)` that executes `cleanup_applications --employer-drafts --youth-expired` once a month.

### Unit Tests
- Updated existing tests in `test_cleanup_applications_command.py` to use `--ignore-5-years-restriction` where appropriate.
- Added new retention tests for the cleanup command to verify both default and bypassed behaviors for employer drafts and expired youth applications.
- Added a unit test in `test_cron_job.py` to verify that the monthly job executes successfully and respects the default 5-year retention limit.

### Documentation
- Updated `README.md` to document the updated `cleanup_applications` command, explain the 5-year retention policy and the bypass flag, and provide a monthly cronjob example (`0 3 1 * *`).

## Summary by CodeRabbit

* **New Features**
  * Added `--ignore-5-years-restriction` flag to override the default 5-year retention policy for application cleanup operations.

* **Documentation**
  * Updated documentation with details about the cleanup operation, including CLI arguments and example cronjob configuration.

* **Tests**
  * Added tests for cleanup operations and 5-year retention behavior validation.

### DIFF
--- a/backend/kesaseteli/applications/management/commands/cleanup_applications.py
+++ b/backend/kesaseteli/applications/management/commands/cleanup_applications.py
@@ -1,6 +1,8 @@
 import logging
 
+from dateutil.relativedelta import relativedelta
 from django.core.management.base import BaseCommand
+from django.utils import timezone
 
 from applications.enums import EmployerApplicationStatus
 from applications.models import EmployerApplication, YouthApplication
@@ -28,11 +30,20 @@ class Command(BaseCommand):
             action="store_true",
             help="Dry run mode, don't actually delete anything.",
         )
+        parser.add_argument(
+            "--ignore-5-years-restriction",
+            action="store_true",
+            help="Bypass the 5-year retention restriction. By default, "
+            "all the applications must be stored for 5 years by information "
+            "management plan.",
+        )
 
     def handle(self, *args, **options):
+        """Handle cleanup operations."""
         employer_drafts = options["employer_drafts"]
         youth_expired = options["youth_expired"]
         dry_run = options["dry_run"]
+        ignore_5_years_restriction = options["ignore_5_years_restriction"]
 
         if not employer_drafts and not youth_expired:
             self.stdout.write(
@@ -41,13 +52,17 @@ class Command(BaseCommand):
             return
 
         if employer_drafts:
-            self.cleanup_employer_drafts(dry_run)
+            self.cleanup_employer_drafts(dry_run, ignore_5_years_restriction)
 
         if youth_expired:
-            self.cleanup_youth_expired(dry_run)
+            self.cleanup_youth_expired(dry_run, ignore_5_years_restriction)
 
-    def cleanup_employer_drafts(self, dry_run):
+    def cleanup_employer_drafts(self, dry_run, ignore_5_years_restriction):
+        """Clean up employer applications in draft mode."""
         qs = EmployerApplication.objects.filter(status=EmployerApplicationStatus.DRAFT)
+        if not ignore_5_years_restriction:
+            five_years_ago = timezone.now() - relativedelta(years=5)
+            qs = qs.filter(created_at__lte=five_years_ago)
         count = qs.count()
         if dry_run:
             self.stdout.write(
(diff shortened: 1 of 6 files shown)


backend/kesaseteli/applications/management/commands/cleanup_applications.py
```diff
@@ -59,12 +72,15 @@ def cleanup_employer_drafts(self, dry_run):
...
 
-    def cleanup_youth_expired(self, dry_run):
+    def cleanup_youth_expired(self, dry_run, ignore_5_years_restriction):
         # YouthApplication.expired() returns those where created_at older than
         # expiration according to YouthApplication.expiration_duration()
         qs = YouthApplication.objects.filter(
             receipt_confirmed_at__isnull=True
         ).expired()
+        if not ignore_5_years_restriction:
+            five_years_ago = timezone.now() - timedelta(days=5 * 365)
```
ToivoMattilaHelsinki: Could use relativedelta(years=5) instead of 5 * 365 days

Relativedelta with years is bit more explicit and handles leap years

ToivoMattilaHelsinki (approved):

backend/kesaseteli/applications/management/commands/cleanup_applications.py
nikomakela: Good. I changed to relativedelta.

backend/kesaseteli/applications/management/commands/cleanup_applications.py
```diff
@@ -59,12 +74,16 @@ def cleanup_employer_drafts(self, dry_run):
...
+    def cleanup_youth_expired(self, dry_run, ignore_5_years_restriction):
+        """Clean up youth applications that have expired."""
         # YouthApplication.expired() returns those where created_at older than
         # expiration according to YouthApplication.expiration_duration()
         qs = YouthApplication.objects.filter(
             receipt_confirmed_at__isnull=True
         ).expired()
+        if not ignore_5_years_restriction:
+            five_years_ago = timezone.now() - relativedelta(years=5)
+            qs = qs.filter(created_at__lte=five_years_ago)
```
helfijuvon, line 86: The data retention policies dictate that the relative delta of 5 years should be calculated from the moment when the YouthApplication handling has been finalized.

Logically the status of the YouthApplication should be Handled, meaning either Accepted or Rejected to count as being finalized and relativedelta should be based on the date when the last modification (accept or reject) was made.

We can chat about this in daily or slack hudle.

nikomakela: The problem is that it might never be "finalized". Could finalization mean that it has expired?  If it has only be filled, but never activated.

karisal-anders: There are the following statuses of YouthApplication:
- Submitted (+expired or unexpired depending on how long since the YouthApplication was submitted, but the expiration is not saved as a different status)
- Awaiting manual processing
- Additional information requested
- Additional information provided
- Accepted
- Rejected

So Accepted and Rejected are easy, they have corresponding timestamp YouthApplication.handled_at for when they were accepted or rejected.

All the other statuses are not as easy, as e.g. a youth application can remain indefinitely in "Additional information requested" if the person never provides additional information—such applications are not shown to the youth application handlers.

karisal-anders: > The problem is that it might never be "finalized". Could finalization mean that it has expired?

I think that sounds appropriate for the submitted status youth applications that have expired (YouthApplicationQuerySet.expired → ... → depends on NEXT_PUBLIC_ACTIVATION_LINK_EXPIRATION_SECONDS).

karisal-anders: One possibility for a some kind of an expiration of sorts for e.g. youth applications caught in "Additional information requested" status limbo without any such additional information ever been provided could be the end of the year when the youth application was created, because youth applications are specific to a year. The same person can have a youth application (+associated youth summer voucher) for e.g. 2025 and for 2026, but not two for a single year.

Youth applications don't cross year boundaries, so a youth application created in 2025 applies to year 2025, not to any others (Employer applications and employer summer vouchers can be a bit tricky, as sometimes they have AFAIK referred to a previous year's youth summer voucher, although this is not the normal case. This can happen if employer applications are delayed and only done during January/February instead of during the employment year's summer/autumn/winter).

helfijuvon: Yes expiration counts as being finalized as well, good catch.

helfijuvon: Good points, I think we (city of helsinki) needs to define when these applications which end in a status limbo can be defined as being expired since users can just abandon the application in the "Additional information requested" status.

helfijuvon: Added new ticket regarding the "Additional information requested"-limbo:
https://helsinkisolutionoffice.atlassian.net/browse/YJDH-904

PR #4143, opened 2026-06-10
feat(ks): display related summer vouchers inline in admin

nikomakela:
YJDH-884.

Simplify administration and improve usability of the admin site. Because YouthSummerVoucher heavily relies on YouthApplication and EmployerSummerVoucher, and EmployerSummerVoucher heavily relies on EmployerApplication and YouthSummerVoucher, showing these relations directly inside the admin inline views makes it significantly easier to track and inspect associated entities.

## Summary by CodeRabbit

* **New Features**
  * Read-only inline admin views showing employer↔youth summer voucher relationships, linked application and company info, admin links to related records, and masked employee SSN.
  * Admin list/detail views now pre-load related voucher and application data for smoother inline rendering.

* **Tests**
  * Added tests for inline registration, read-only permissions, admin link generation, and unsaved/empty-object cases.

### DIFF
--- a/backend/kesaseteli/applications/admin.py
+++ b/backend/kesaseteli/applications/admin.py
@@ -517,9 +517,88 @@ class YouthApplicationAdmin(AuditlogAdminViewAccessLogMixin, admin.ModelAdmin):
         return False
 
 
+class YouthEmployerSummerVoucherInline(admin.TabularInline):
+    """
+    Django Admin Inline View for Employer Summer Vouchers
+
+    Display employer summer vouchers in the YouthSummerVoucher admin view.
+    Shows related employer application and voucher details in a tabular format
+    with links to the respective admin pages.
+
+    Features:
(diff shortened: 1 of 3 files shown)


backend/kesaseteli/applications/admin.py
```diff
@@ -517,9 +517,74 @@ def has_delete_permission(self, request, obj=None):
         return False
 
 
+class YouthEmployerSummerVoucherInline(admin.TabularInline):
```
karisal-anders, line 520: Please would you add a docstring that clearly tells what this class is.
I find the naming ambiguous and unclear, but as I have no better names to suggest at this point I'd be ok with a description that clearly tells what this class is. That way IDE can show at least the description for clarity so user can then be clarified as to what this class really is, if they don't get its functionality straight from the name.

karisal-anders (approved):
Left one comment, please take a look at it.

PR #4155, opened 2026-06-15
feat(ks): add csp middleware

ToivoMattilaHelsinki:
Refs: YJDH-778

## Description :sparkles:

Add django-csp to the Kesäseteli and enable a minimal Content Security Policy.
CSP covers backend-only views such ReDoc pages
Add CSP headers to frontend

## Issues :bug:

No known issues.

## Testing :alembic:

Added regression tests for checking CSP headers in backend and frontend

## Screenshots :camera_flash:

N/A.

healthz and readiness are explicitly exempted so probe responses stay header-free.

## Summary by CodeRabbit

* **Chores**
  * Enabled Content Security Policy (CSP) headers across the backend and frontend for improved browser security.
  * Secured API documentation pages (Swagger UI and ReDoc) and applied the correct CSP on the excel download view while keeping the email preview on the global policy.
  * Ensured health and readiness probe endpoints do not send CSP headers.
  * Wrapped Next.js builds to consistently emit CSP headers.
* **Tests**
  * Added coverage verifying CSP headers (and their directives) on key pages and confirming probes omit the header.
* **Documentation**
  * Updated API documentation routing to use CSP-enabled documentation pages.

### DIFF
--- a/backend/kesaseteli/kesaseteli/settings.py
+++ b/backend/kesaseteli/kesaseteli/settings.py
@@ -297,6 +297,7 @@ MIDDLEWARE = [
     "django.contrib.sessions.middleware.SessionMiddleware",
     "corsheaders.middleware.CorsMiddleware",
     "csp.middleware.CSPMiddleware",
+    "common.middleware.PermissionsPolicyMiddleware",
     "django.middleware.locale.LocaleMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
--- a/backend/kesaseteli/kesaseteli/settings.py
+++ b/backend/kesaseteli/kesaseteli/settings.py
@@ -346,6 +346,8 @@ CONTENT_SECURITY_POLICY = {
         "default-src": [SELF],
         "style-src": [SELF, UNSAFE_INLINE],
         "img-src": [SELF, "data:"],
+        "base-uri": [SELF],
+        "object-src": ["'none'"],
     }
 }
 
--- a/backend/kesaseteli/kesaseteli/settings.py
+++ b/backend/kesaseteli/kesaseteli/settings.py
@@ -9,6 +9,7 @@ import environ
 import saml2
 import saml2.saml
 import sentry_sdk
+from csp.constants import SELF, UNSAFE_INLINE
 from django.core.exceptions import ImproperlyConfigured
 from django.urls import reverse_lazy
 from django.utils.translation import gettext_lazy as _
@@ -268,6 +269,7 @@ INSTALLED_APPS = [
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "corsheaders",
+    "csp",
     "rest_framework",
     "mozilla_django_oidc",
     "django_extensions",
@@ -294,6 +296,7 @@ MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "corsheaders.middleware.CorsMiddleware",
+    "csp.middleware.CSPMiddleware",
     "django.middleware.locale.LocaleMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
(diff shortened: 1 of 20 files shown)


backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -337,6 +340,14 @@
 CORS_ALLOW_CREDENTIALS = True
 CORS_ALLOWED_ORIGINS = env.list("CORS_ALLOWED_ORIGINS")
 CORS_ALLOW_ALL_ORIGINS = env.bool("CORS_ALLOW_ALL_ORIGINS")
+
+CONTENT_SECURITY_POLICY = {
+    "DIRECTIVES": {
+        "default-src": [SELF],
+        "style-src": [SELF, UNSAFE_INLINE, "cdn.jsdelivr.net"],
```
nikomakela: Where are UNSAFE_INLINE and "cdn.jsdelivr.net" needed? If only in OpenAPI docs, I second what coderabbit[bot] says: https://github.com/City-of-Helsinki/yjdh/pull/4155/changes#r3426229021.

> Scope cdn.jsdelivr.net to docs-only CSP, not the global policy.
>
> The global style-src currently permits jsdelivr on every backend response. Since only the Swagger and ReDoc documentation views (in openapi_views.py) require it, and those views use @csp_update(API_DOCS_CSP) to append their additional directives, removing jsdelivr from the global policy and relying on the decorator's merge behavior for docs views narrows the unnecessary trust boundary.

And I see that you have already done that too! Which means that this should be needless?

backend/kesaseteli/kesaseteli/settings.py
```diff
@@ -337,6 +340,14 @@
...
+    }
+}
```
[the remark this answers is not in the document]
nikomakela: Yes! Like coderabbit says, if it's only openapi docs related, it should not be globally set here.

nikomakela (commented):
At first glance, the code looks okay (except for the global CSP rule in the API for docs), but I will still manually test all the frontends and the API, should I find something.

nikomakela (changes requested):
1. API has a trivial CSP configuration error for svg image in http://localhost:8000/saml2/metadata/.
```Loading the image 'data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' fill='%23909090' width='10' height='10'><path d='M0 0 L8 0 L4 7 Z'/></svg>' violates the following Content Security Policy directive: "default-src 'self'". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback. The action has been blocked.```
2. API has another trivial CSP configuration issue In Swagger UI OpenAPI docs: http://localhost:8000/api_docs/swagger/.
```Connecting to 'https://cdn.jsdelivr.net/npm/swagger-ui-dist@latest/swagger-ui.css.map' violates the following Content Security Policy directive: "default-src 'self'". The request has been blocked. Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.```
3. Trivial API CSP issue in http://localhost:8000/excel-download/. I wonder why we even have a bootstrap CSS there? Is there a need for that? Could it be replaced with something that is shipped with Django?
```Connecting to 'https://cdn.jsdelivr.net/npm/bootstrap@5.1.0/dist/css/bootstrap.min.css.map' violates the following Content Security Policy directive: "default-src 'self'". The request has been blocked. Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.```
4. API is responding writing with CSP header that includes ```style-src 'self' 'unsafe-inline' cdn.jsdelivr.net; default-src 'self'``` when a frontend requests _anything_.
5. In every frontend app, it looks like **the front page ("root index") does not have any CSP header at all**, but e.g. http://localhost:3100/fi/cookie-settings has.
6. In frontend apps, at least the http://localhost:3100/fi/cookie-settings page seems to not need "unsafe-inline" in  ```script-src 'self' 'unsafe-inline'```.

File: frontend/kesaseteli/shared/src/config/csp.js [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+  const connectOrigins = [
+    "'self'",
+    getOriginFromUrl(process.env.NEXT_PUBLIC_BACKEND_URL),
+    getOriginFromUrl(process.env.NEXT_PUBLIC_SENTRY_DSN),
+    ...matomoOrigins,
+  ].filter(Boolean);
+
+  return [
+    "default-src 'self'",
+    `script-src 'self' 'unsafe-inline'${isDev ? " 'unsafe-eval'" : ''} ${matomoOrigins.join(' ')}`,
```
nikomakela: To me it looks like the unsafe-inline here is not needed

```suggestion
    `script-src 'self' ${isDev ? " 'unsafe-eval'" : ''} ${matomoOrigins.join(' ')}`,
```

nikomakela (approved):
Good job

PR #4165, opened 2026-06-18
feat(ks): add summer voucher attachments to Django admin

nikomakela:
YJDH-899.

Allow staff members with view permissions to list, inspect, and securely download employer summer voucher attachments. A standalone AttachmentAdmin view and an AttachmentInline component are registered, enforcing superuser-only delete permissions and read-only attributes, with query prefetching/select_related optimization.

## Summary by CodeRabbit

* **New Features**
  * Added attachment inline support in the admin, showing attachment details and a download link while keeping inline editing disabled.
  * Introduced an attachment admin download page with secure, permission-based access and file streaming.
  * Added audit logging when attachments are downloaded/viewed.

* **Bug Fixes**
  * Strengthened read-only enforcement for attachment-related admin views.
  * Improved admin link rendering for voucher inlines to use safe HTML formatting.

* **Tests**
  * Added coverage for attachment admin UI rendering, download authorization behavior, delete/add restrictions, and audit-log creation on download.

### DIFF
--- a/backend/kesaseteli/applications/admin.py
+++ b/backend/kesaseteli/applications/admin.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from itertools import chain
 
 from auditlog_extra.mixins import AuditlogAdminViewAccessLogMixin
@@ -7,17 +8,24 @@ from django.apps import apps
 from django.conf import settings
 from django.contrib import admin
 from django.contrib.admin import helpers
-from django.http import HttpResponse, HttpResponseRedirect
+from django.http import (
+    FileResponse,
+    Http404,
+    HttpResponse,
+    HttpResponseForbidden,
+    HttpResponseRedirect,
+)
 from django.shortcuts import render
 from django.urls import path, reverse
 from django.utils import timezone
 from django.utils.formats import date_format
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html
 from django.utils.translation import gettext_lazy as _
 from django.utils.translation import ngettext
 
 from applications.admin_forms import SchoolImportForm
 from applications.models import (
+    Attachment,
     EmailTemplate,
     EmployerApplication,
     EmployerSummerVoucher,
@@ -26,7 +34,11 @@ from applications.models import (
     YouthApplication,
     YouthSummerVoucher,
 )
-from applications.services import EmailTemplateService, SchoolService
+from applications.services import (
+    AuditAccessLogService,
+    EmailTemplateService,
+    SchoolService,
+)
 from applications.target_groups import get_target_group_choices
 from common.utils import mask_social_security_number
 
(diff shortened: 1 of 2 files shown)


backend/kesaseteli/applications/admin.py
```diff
@@ -913,9 +925,197 @@ def get_queryset(self, request):
...
+        "created_at",
+    ]
+    list_filter = ["attachment_type", "created_at", "content_type"]
+    search_fields = [
+        "id",
+        "summer_voucher__id",
+        "summer_voucher__youth_summer_voucher__summer_voucher_serial_number",
+        "summer_voucher__application__company__name",
+        "summer_voucher__application__company__business_id",
+    ]
```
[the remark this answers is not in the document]
nikomakela: e potential issue identified by the AI agent is a false positive.

Here is why the agent is incorrect and why no fix is needed:

In older versions of Django or with certain naive SQL implementations, running a LIKE operation directly on a UUID column in PostgreSQL would indeed cause a DataError.

However, modern versions of Django (this project is running Django 5.2.15) are smart enough to handle this automatically in the ORM. When Django's Admin generates a search query using __icontains on a UUIDField, it automatically tells PostgreSQL to cast the UUID to text before performing the substring match.

backend/kesaseteli/applications/admin.py
```diff
@@ -913,9 +925,218 @@ def get_queryset(self, request):
...
+    download_link.short_description = _("download link")
+
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
+        return request.user.is_superuser
```
ToivoMattilaHelsinki: download_link, has_add_permission, has_change_permission, and has_delete_permission are duplicated between AttachmentInline and AttachmentAdmin

Is this on purpose?

backend/kesaseteli/applications/tests/test_admin_attachments.py
```diff
@@ -0,0 +1,273 @@
...
+    assert response.status_code in [302, 403]
+
+    # 2. Superuser should have delete permission (should load delete confirmation page)
+    superuser = StaffSuperuserFactory()
+    client.force_login(superuser)
+
+    response = client.get(delete_url)
+    assert response.status_code == 200
+    # The title on the confirmation page is "Poista" (localized Django admin delete)
+    assert "Poista" in response.content.decode("utf-8")
```
ToivoMattilaHelsinki: The assertion uses hardcoded Finnish word "Poista" on the assertion

We get a false negative if for whatever reason the page is not in Finnish

We could harden the test by either forcing the translation language to be Finnish or by using _("Delete") as the assertion string

ToivoMattilaHelsinki (approved):

backend/kesaseteli/applications/admin.py
nikomakela: Yes, I would say so. They are used in different classes. AttachmentAdmin is not inheriting AttachmentInline, but AttachmentInline is used as inline component in some other domains. The only relation between the AttachmentAdmin and AttachmentInline is that they are for Attachment model, but they are "never used together". How I undestand this is that I need to add the permission functions to tabularinline-class too, because thse scope is only for that component, not for the class that is using it as inline component.

Or did you mean that we could use a mixin (class), because that would actually really make some sense!

nikomakela: I can refactor it so that it uses a mixin. Probably the same invention can be used else where too.

backend/kesaseteli/applications/tests/test_admin_attachments.py
nikomakela: You are right. I can do that. I'm not 100% sure how the lazy translations are used in pytests, but I can try. If I remember right, this isn't the only place using Finnish like that, but I can at least try to fix this new code.

nikomakela: I will define the accepted language in headers:

```
    response = client.get(delete_url, HTTP_ACCEPT_LANGUAGE="en")
    assert response.status_code == 200
    assert "Delete" in response.content.decode("utf-8")
```

backend/kesaseteli/applications/admin.py
nikomakela: I added ReadOnlyAdminMixin and SuperuserDeleteAdminMixin to top of the file nad used them where appropriate.

PR #4180, opened 2026-06-24
feat(ks): gate 2026 columns and status field under EXCLUDE_2026_EXCEL_FIELDS flag

nikomakela:
YJDH-909.
Gate the target group calculation status field (`Erikoistapauksen laskentatila`) and the 2026 end-fields under the shared feature flag `EXCLUDE_2026_EXCEL_FIELDS` (defaulting to True) so they are only included in Talpa exports when the flag is disabled.
### New Features / Changes
- **Feature Flag Gating**: Gated the `"Erikoistapauksen laskentatila"` status column behind the `EXCLUDE_2026_EXCEL_FIELDS` feature flag. The column is now excluded by default (rollback safety) and only visible when the toggle is set to `False`.
- **Excel Exporter Column Ordering**: Included `_("Erikoistapauksen laskentatila")` directly inside `TALPA_END_FIELD_TITLES` so that both the status column and the 2026 fields are positioned at the very end of the sheet to avoid disrupting Talpa robot processing when enabled.
- **Simplified get_talpa_columns**: Cleaned up the filtering logic in `get_talpa_columns()` to cleanly choose between the old baseline column list (excluding all 2026 and status fields) and the new columns list.
### Tests
- **Unified & Parametrized Gating Test**: Combined and parameterized the gating checks into a single, clean test (`test_talpa_excel_gated_2026_and_status_fields`) to validate the presence/absence of all 7 gated columns based on `EXCLUDE_2026_EXCEL_FIELDS` using a single parameter.
- **Import cleanup**: Moved `TALPA_END_FIELD_TITLES` to top-level imports in `test_excel_export.py`.
### Documentation/Configuration
- Updated the backend `.env.kesaseteli-backend.example` to default to `EXCLUDE_2026_EXCEL_FIELDS=True`.


ToivoMattilaHelsinki (approved):
Didn't find any blocking changes

Minor style nitpick: the test_excel_export and excel_export files are quite long and could be broken into separate, smaller files at some point

PR #4186, opened 2026-06-25
fix(ks): restrict handler permissions to authenticated users

nikomakela:
YJDH-910.

Ensure that HandlerPermission is not granted to AnonymousUser even when NEXT_PUBLIC_MOCK_FLAG is True. This prevents unauthenticated users from accessing sensitive views like Excel downloads and makes audit logging robust by ensuring an authenticated actor is always present.

Add detailed comments in HandlerPermission explaining why the mock flag bypass check is kept for authenticated users (to support api_client in parametrized tests and local development).

## Summary by CodeRabbit

* **Bug Fixes**
  * In mock mode, unauthenticated requests across the youth application flow now consistently redirect to the login page instead of returning mixed success/error responses.
  * Permission checks now only grant access in mock mode to authenticated users, improving consistency and preventing unintended handling.
  * Updated youth application action/acceptance behavior so unauthorized clients are redirected prior to viewing or processing requests.

### DIFF
--- a/backend/kesaseteli/applications/tests/test_excel_export.py
+++ b/backend/kesaseteli/applications/tests/test_excel_export.py
@@ -730,3 +730,15 @@ def test_youth_excel_download_with_missing_birthdate(staff_client):
     # Test that birth_year and birthdate use their respective fallback values:
     assert row_by_field["birth_year"] == -1
     assert row_by_field["birthdate"] == "N/A"
+
+
+@pytest.mark.django_db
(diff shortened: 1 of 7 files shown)


ToivoMattilaHelsinki (approved):
Looks good

Optionally an additional Excel regression test might be useful

Right now all Excel tests run with NEXT_PUBLIC_MOCK_FLAG=False, and the “unauthenticated” cases use user_client (authenticated non-staff → 403). A small test with mock flag on + plain unauthenticated client GET on youth-excel-download → ADFS redirect might be good.

PR #4206, opened 2026-07-08
feat(ks,handler): persist active tab state in application lists

nikomakela:
YJDH-916.

Create a custom React hook useSessionStorageState that manages state and synchronizes it with sessionStorage to preserve the user's active tab index (Pending vs. Processed) in both the Employer and Youth application list tables.

This ensures that when a handler navigates to a detail view and back, they are returned to their active tab queue instead of the default tab.


helfijuvon (approved):
Reviewed with @nikomakela and approved to merge.

PR #4241, opened 2026-08-11
HL-1791 | upgrade to HDS 6 and React 19 (DO NOT MERGE!)

jorilindell:
DO NOT MERGE. The Comodori team has a feature they want to review and test before this is merged.

## Description :sparkles:

Upgrades the frontend to **HDS v6** and **React 19** across all five Next.js apps (benefit applicant, benefit handler, kesäseteli employer, kesäseteli handler, kesäseteli youth) and the shared packages.

Everything else in this PR exists only because those two upgrades required it — none of it is optional or unrelated cleanup.

### The actual upgrade

**HDS v5 → v6** (`hds-react`, `hds-design-tokens` `^6.0.5`, `hds-react: 5.0.0` pnpm override dropped)
- `ErrorSummary` was removed in v6 → replaced with `Notification type="error"` in employer, ks-handler and youth flows.
- `Link` wrapper prop typings adjusted to avoid ref-spread issues with the new HDS types.

**React 18 → 19** (`react`, `react-dom` `^19.2.7`, `@types/react` / `@types/react-dom` 19)
- Legacy `JSX.*` namespace annotations migrated to `React.ReactElement` / `React.ReactNode` / `React.ElementType`.

### Forced by the upgrade

| Change | Why it was unavoidable |
| --- | --- |
| **react-query v3 → @tanstack/react-query v5** (`5.101.0`) | `react-query` v3 does not support React 19. ~90 query/mutation hooks migrated to the object API; `QueryClient` defaults updated, unsupported options (e.g. `notifyOnChangeProps`) removed, `queryFn` signatures and error handling adjusted. |
| **styled-components v5 → v6.1.19** | v5 is not React 19 compatible. `@types/styled-components` dropped (v6 ships its own), `cssprop.d.ts` and the `.sc.ts` files adjusted. |
| **react-input-mask → Maskito** | `react-input-mask` relies on `findDOMNode`, removed in React 19. Bank account and IBAN inputs migrated to `useMaskito`, with a shared helper converting the legacy mask format strings and `TextInput` forwarding `inputRef` alongside the RHF ref. |
| **react-markdown 8 → 10** (applicant) | Old typings referenced the removed `JSX` global namespace and failed typecheck. |
| **Jest 27 → 30**, `@testing-library/react` 14 → 16, `jest-dom` 5 → 7, `jest-environment-jsdom` 30 | RTL 14 cannot render React 19; RTL 16 requires a newer Jest/jsdom. Legacy `ts-jest` config and dependency removed as part of the move. |
| **hds-core cookie consent shim** | HDS 6.0.3 ships `cookieConsent.ts` but not the `cookieConsent.js` that `hds-react` resolves, breaking Next.js page-data collection. Added fix-hds-core-cookie-consent.js, run as `fix-hds-shim` before `dev`/`build` in every app and wired into both Dockerfiles so the production image keeps the patched package. |
| **Sentry edge runtime config** | Surfaced by the Next/React upgrade. Instrumentation now loads runtime-specific config, edge config files added for benefit applicant and handler, shared benefit options extracted into `sentry-common.ts`. |

## Issues

Refs: HL-1791


ext-hjasplund (approved):

frontend/shared/src/hooks/useQuerySideEffect.ts
```diff
@@ -0,0 +1,47 @@
...
+  const handleError = useEffectEvent((error: E) => {
+    onError?.(error);
+  });
+  const shouldHandleError = useEffectEvent(
+    (error: E) => onErrorPredicate === undefined || onErrorPredicate(error)
+  );
+
+  // Run side effects only when query state genuinely changes
+  useEffect(() => {
+    if (query.data) {
```
Riippi: **`onError` is unreachable whenever the query holds any data.**

react-query keeps the last successful `data` when a refetch fails — on error only `error`/`status` change, `data` stays. So once a query has succeeded even once, `query.data` is truthy forever and the `else if` branch can never run. This hook is now the *only* error path for ~15 queries that previously used `onError`, so several real behaviours are lost:

- `kesaseteli/employer/src/hooks/backend/useUserQuery.ts` and `kesaseteli/handler/src/hooks/backend/useUserQuery.ts` poll with a 5-minute `refetchInterval`. When the session expires, the poll 401s, but `query.data` still holds the cached user — so `onAuthError` never fires and the user is never redirected to `/login?sessionExpired=true`. They keep sitting on a stale page whose every subsequent write will fail.
- Same for `benefit/applicant/src/hooks/useUserQuery.ts` and `benefit/handler/src/hooks/useUserQuery.ts` (401/403 → push to login; the handler one also does the `is_staff` check).
- Every list query using `placeholderData: keepPreviousData` (`useEmployerApplicationsListQuery`, `useYouthApplicationsListQuery`, employer `useApplicationsQuery`) *always* has data, so their error toasts never appear at all.

Worse, on those same renders `errorUpdatedAt` changes → the effect re-runs → `query.data` is truthy → `handleSuccess` is invoked **on failure**. For applicant `useUserQuery` that means re-writing the CSRF header from stale data after a 401.

Suggested fix — branch on status, not on data truthiness:

```ts
useEffect(() => {
  if (query.isError && query.error) {
    if (shouldHandleError(query.error)) handleError(query.error);
  } else if (query.isSuccess) {
    handleSuccess(query.data);
  }
}, [...]);
```

Note the truthiness check is also wrong for legitimately-falsy data: employer `AuthProvider` uses `useUserQuery<boolean>({ select: (user) => Boolean(user) })`, so `data === false` is a valid success that this skips.

Riippi (commented):
### Code review: correctness pass over the react-query v5 / React 19 migration

Read every non-test source hunk. The migration is mostly mechanical and correct — I verified that `isLoading → isPending` is only applied to **mutations** (so the v5 "disabled query is permanently `pending`" trap is avoided), that the `'idle'|'loading' → 'pending'` status mappings are right, and that the string→array query-key conversions preserve v3's matching semantics (v3 also compared string keys element-wise, so e.g. `['/v1/employerapplications/']` never matched `['/v1/employerapplications/123/']` before either). `useIsFetching({fetching:false}) → useIsFetching()` is also behaviour-preserving, since v3 force-overrode `fetching: true`.

6 inline comments below. One extra finding that lands on a deleted line, so it can't be inline:

**`frontend/benefit/applicant/package.json` — `i18next` removed but still imported directly.**
`"i18next": "^25.3.2"` was dropped from benefit/applicant, but `frontend/benefit/applicant/test/i18n/i18n-test.ts:1` still does `import i18n from 'i18next'`, and that file is wired into `setupFilesAfterEnv` in `jest.config.js` — so it loads on *every* applicant Jest run. Under pnpm's non-hoisted layout (`shamefullyHoist: false`) that's an undeclared dependency that only resolves by luck. Either keep `i18next` as a declared devDependency or import it through `next-i18next`.

(The `@react-pdf/renderer` and `fuse.js` removals are clean — no remaining usages anywhere.)

frontend/benefit/applicant/src/hooks/useMessagesQuery.ts
```diff
@@ -15,28 +16,29 @@ const useMessagesQuery = (
...
+
+  useEffect(() => {
+    if (query.isError) {
+      showErrorToast(
+        t('common:messenger.list.errors.fetch.label'),
+        t('common:messenger.list.errors.fetch.text')
+      );
     }
-  );
+  }, [query, t]);
```
Riippi: **Depending on the whole `query` object makes this fire on every render.**

v5's `useQuery` returns a *new* tracked result object on each render (`observer.trackResult(...)` builds a fresh object unless `notifyOnChangeProps` is set), so `[query, t]` never compares equal and the effect re-runs on every render.

Scenario: the messages sidebar is open (`enabled: isOpen`, `refetchInterval: 30_000`) and the fetch is failing. `query.isError` stays `true`, so every re-render — each keystroke in the message box, every parent re-render, every 30s poll — pushes another "failed to fetch messages" toast. `showErrorToast` mints a fresh uuid `toastId` each time, so they stack rather than dedupe.

```suggestion
  }, [query.isError, t]);
```

frontend/benefit/handler/src/hooks/useApplicationsWithMessagesQuery.ts
```diff
@@ -34,12 +28,20 @@ const useApplicationMessagesQuery = (): UseQueryResult<
...
+
+  useEffect(() => {
+    if (query.isError) {
+      showErrorToast(
+        t('common:applications.list.errors.fetch.label'),
+        t('common:applications.list.errors.fetch.text', { status: 'error' })
+      );
     }
-  );
+  }, [query, t]);
```
Riippi: **Same issue as `benefit/applicant/src/hooks/useMessagesQuery.ts`:** `query` is a new object on every render, so this effect re-runs every render and stacks a fresh error toast on each one while the unread-messages poll is failing.

```suggestion
  }, [query.isError, t]);
```

frontend/shared/src/hooks/useErrorHandler.ts
```diff
@@ -33,8 +34,11 @@ const useErrorHandler = (props: Props = {}): ErrorHandlerFunction => {
   const { pathname } = useRouter();
   const goToPage = useGoToPage();
 
-  return (error: Error | unknown) =>
-    handleError({ error, t, pathname, goToPage, ...props });
+  return useCallback(
+    (error: Error | unknown) =>
+      handleError({ error, t, pathname, goToPage, ...props }),
+    [t, pathname, goToPage, props]
```
Riippi: **This `useCallback` never actually memoizes.**

`props` defaults to a fresh `{}` on every call, and the callers that do pass options pass inline object literals (`useCompanyQuery`, both employer/handler `useUserQuery`). So the dependency changes identity every render and the returned handler is new every render — the memoization is a no-op.

That propagates: `useCreateApplication` in `kesaseteli/employer/src/hooks/application/useLoadDraftOrCreateNewApplication.ts` lists `errorHandler` in its `useCallback` deps, and the resulting `createApplication` is itself an effect dependency — so that effect re-runs on every render too (currently saved only by the `createApplicationQuery.isIdle` guard).

Depend on the individual handlers instead of the object, e.g.:

```ts
const { onServerError, onAuthError, ...rest } = props;
return useCallback(
  (error: Error | unknown) => handleError({ error, t, pathname, goToPage, onServerError, onAuthError, ...rest }),
  [t, pathname, goToPage, onServerError, onAuthError]
);
```

frontend/kesaseteli/employer/src/components/application/steps/step1/EmployerErrorSummary.tsx
```diff
@@ -27,8 +27,9 @@ const EmployerErrorSummary: React.FC = () => {
 
   return (
     <$ErrorSummary
+      type="error"
       label={t(`common:application.form.notification.title`)}
-      autofocus={isJustSubmitted}
+      autoFocus={isJustSubmitted}
```
Riippi: **`autofocus` → `autoFocus` silently drops the focus behaviour.**

HDS `ErrorSummary` had a real `autofocus` prop that moved focus to the summary. HDS `Notification` has no `autoFocus` prop and renders a non-focusable container, so this is either dropped by the component or forwarded to a `div` — where `autofocus` is inert.

Result: after a failed submit, focus no longer moves to the error summary. Keyboard and screen-reader users get no indication the submit failed and have to hunt for it.

Same in `kesaseteli/youth/src/components/youth-form/SubmitErrorSummary.tsx:43` and `kesaseteli/handler/src/components/form/SubmitErrorSummary.tsx:49`.

Needs to be done explicitly, e.g. a `ref` + `tabIndex={-1}` on the notification and `useEffect(() => { if (isJustSubmitted) ref.current?.focus(); }, [isJustSubmitted])`.

frontend/shared/src/components/app/BaseApp.tsx
```diff
@@ -1,10 +1,8 @@
-import * as Sentry from '@sentry/browser';
+import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
```
Riippi, line 1: **Sentry reporting for query/mutation failures is dropped with no replacement.**

This commit deletes the `setLogger({ log, warn, error })` block that forwarded react-query errors to `Sentry.captureException` / `captureMessage`. `setLogger` is indeed gone in v5, but the v5 equivalent isn't wired up anywhere — I grepped the diff and there's no `QueryCache({ onError })` / `MutationCache({ onError })` in any of the five `_app.tsx` files or `create-query-client.ts`.

So every backend failure across all five apps that previously surfaced in Sentry is now silent in production. That's an observability regression rather than pure cleanup, and it's easy to miss because nothing fails visibly.

The replacement lives on the `QueryClient`:

```ts
new QueryClient({
  queryCache: new QueryCache({
    onError: (error) => { Sentry.captureException(error); },
  }),
  mutationCache: new MutationCache({
    onError: (error) => { Sentry.captureException(error); },
  }),
});
```

Riippi (approved):

